PR #45, opened 2025-10-28
remove channels from `vardiff` once channel is dropped

### LINKED ISSUE #43: pool not removing closed channels from vardiff
While debugging some problem with my testing deck I noticed that:

* when the pool receives a `CloseChannel` message it does not remove the channel from the vardiff hashmap:

https://github.com/stratum-mining/stratum/blob/0b7f8257ffe6f05bdc025e8619045dab77afbd08/roles/pool/src/lib/channel_manager/mining_message_handler.rs#L48-L71

* In the same context, if a downstream drops with an open channel the vardiff still runs for the channels of that disconnected downstream:

https://github.com/stratum-mining/stratum/blob/0b7f8257ffe6f05bdc025e8619045dab77afbd08/roles/pool/src/lib/channel_manager/mod.rs#L337-L342

lucasbalieiro:
closes #43 

This makes the pool and JDC drop channels from the vardiff hashmap:

- when receive CloseChannel message.
- if the downstream drops without closing channels

### DIFF
--- a/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
+++ b/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
@@ -66,6 +66,9 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         downstream_data.standard_channels.remove(&msg.channel_id);
                         downstream_data.extended_channels.remove(&msg.channel_id);
                     });
+                channel_manager_data
+                    .vardiff
+                    .remove(&(downstream_id, msg.channel_id).into());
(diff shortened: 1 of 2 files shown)


lucasbalieiro:
> no changes on tProxy?

As far as I could understand the code there, the tproxy is already handling these cases in [difficulty_manager.rs](https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs)

plebhash (approved):

PR #48, opened 2025-10-28
Add contextual error to Pool

Shourya742:
This PR adds contextual error handling for issues originating in the channel manager. With these changes, errors caused by a single downstream no longer bring down the entire Pool only the affected downstream connection is gracefully disconnected.


Shourya742:
Closing this

PR #59, opened 2025-11-03
Bitcoin Core IPC

### LINKED ISSUE #39: new lib crate: Sv2 ↔ Bitcoin Core IPC
up until now, we had two options for deploying a Sv2 Template Provider:
- build a modified `bitcoind` from [`sv2` branch of @Sjors fork](https://github.com/Sjors/bitcoin/tree/sv2) that listens on a dedicated Sv2 port
- build a `bitcoin-mine` sidecar from [`sv2-ipc` branch of @Sjors fork](https://github.com/Sjors/bitcoin/tree/sv2-ipc), which connects via [`capnproto`](https://capnproto.org/) protocol over IPC to `bitcoin-node` that's built from early development iterations of Bitcoin Core

with https://github.com/bitcoin/bitcoin/pull/31802 merged, Bitcoin Core v30.0 is about to add `bitcoin-node` to the official releases, meaning that essentially any software can leverage `capnproto` over IPC and benefit from Bitcoin Core mining interface.

some early work has already been done for this, namely:
- [`github.com/Sjors/sv2-tp`](https://github.com/Sjors/sv2-tp) is a dedicated repo for the C++ implementation of `bitcoin-mine` sidecar mentioned above (now renamed as `sv2-tp`)
- @rustaceanrob has done some [experimental rust integration](https://github.com/rustaceanrob/bitcoin-ipc/tree/mining), described in detail in https://github.com/bitcoin-core/libmultiprocess/issues/174
- @Shourya742 has done some [experimental integration into SRI](https://github.com/Shourya742/stratum/tree/2025-05-05-template-provider-role), where a new crate provides an executable that binds to Bitcoin Core over IPC and listens on a dedicated Sv2 port to perform the Sv2 Template Distribution Protocol

this issue is meant to track the work where, inspired by the references above, we build a crate that can be used as a library to be integrated into Rust apps.

this library shall act as a "translation" layer between Bitcoin Core's `capnproto` Mining Interface over IPC and Sv2 Template Distribution Protocol.

SRI apps using this library (e.g.: JDC, Pool) shall be able to connect to `bitcoin-node` (from official v30+ Bitcoin Core releases) over a UNIX socket and leverage APIs providing Sv2 Template Distribution Protocol messages over an `async` runtime.

plebhash:
close #39 

- [x] add `bitcoin_core_sv2` crate
- [x] adapt Pool to optionally use Bitcoin Core IPC for Template Provider
- [x] adapt JDC to optionally use Bitcoin Core IPC for Template Provider
- [x] adapt CI

slideshow explaining design: https://docs.google.com/presentation/d/1uIh9RPz2DSUe_6WODoOTOAKrit54TjbDClZQf-4FZZM/edit?usp=sharing

### DIFF
--- a/miner-apps/jd-client/README.md
+++ b/miner-apps/jd-client/README.md
@@ -48,6 +48,21 @@ It can receive templates from two potential sources:
 - Bitcoin Core v30+: an officially released Bitcoin Core node running locally, on the same machine, for which a UNIX socket connection will be established
 
 ```
+
+## Requirements
+
+In order to build this, crate you need `capnproto` on your system.
+
+For example, on Ubuntu/Debian:
+```
+apt-get install capnproto libcapnp-dev
+```
+
+Or macOS:
+```
+brew install capnproto
+```
+
 ## Setup
 
 ### Configuration File
--- a/pool-apps/pool/README.md
+++ b/pool-apps/pool/README.md
@@ -23,6 +23,20 @@ It can receive templates from two potential sources:
 - Sv2 Template Provider: a separate Sv2 application running either locally or on a different machine, for which a (optionally encrypted) TCP connection will be established
 - Bitcoin Core v30+: an officially released Bitcoin Core node running locally, on the same machine, for which a UNIX socket connection will be established
 
+## Requirements
+
+In order to build this, crate you need `capnproto` on your system.
+
+For example, on Ubuntu/Debian:
+```
+apt-get install capnproto libcapnp-dev
+```
+
+Or macOS:
+```
+brew install capnproto
+```
+
 ## Setup
 
 ### Configuration File


plebhash:
based on the CI failures, it seems having `capnproto` installed on the system will become a pre-requisite for building Pool and JDC

https://capnproto.org/install.html

I did a quick test on a Ubuntu box and simply doing `apt-get install capnproto` is not sufficient, because the include path is not being resolved out-of-the-box

requires more research, mainly because this will not only impact CI, but it will also impact UX of Pool and JDC

turning the PR back to draft

plebhash:
> But it should also remain possible to run the Template Provider as a standalone app. 

yeah that is a potential feature for `bitcoin_core_sv2` crate

but in this journey, it's important to make each step strategically in the correct order... if we gave this kind of functionality top 1 priority and implemented it first, we would end up with something that's not too flexible, and then later on trying to "adapt" it for more generic use would very likely end up in a mess

so the first step we're taking is to implement a general purpose library, and validate it by integrating it into Pool and JDC

this will greatly simplify UX by allowing Pool and JDC to connect directly to Bitcoin Core without any intermediary processes, which has been extensively discussed as a high priority requirement

this library can also be eventually used to create a sidecar-style application that's equivalent to `sv2-tp`, and that can be done within the same `bitcoin_core_sv2` crate, just not yet on this PR

plebhash:
additionally, it should be noted that Pool and JDC can still connect to a Sv2 TP in the same way they always did

the user chooses between:
- Sv2 TP: `[template_provider_type.Sv2Tp]`
- Bitcoin Core IPC: `[template_provider_type.BitcoinCoreIpc]`

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+        template_ipc_client_request_options.set_use_mempool(true);
+
+        tracing::debug!("Sending createNewBlock request to Bitcoin Core");
+        let template_ipc_client_response = match template_ipc_client_request.send().promise.await {
+            Ok(response) => response,
+            Err(e) => {
+                tracing::error!("Failed to send template IPC client request: {}", e);
+                return Err(BitcoinCoreSv2Error::CapnpError(e));
+            }
+        };
```
Shourya742: We can implement Display for our error types so that this message is automatically included when the error is formatted.
```suggestion
        let template_ipc_client_response = template_ipc_client_request.send().promise.await?; 
```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+
+        let mut template_ipc_client_request = self.mining_ipc_client.create_new_block_request();
+        let mut template_ipc_client_request_options =
+            match template_ipc_client_request.get().get_options() {
+                Ok(options) => options,
+                Err(e) => {
+                    tracing::error!("Failed to get template IPC client request options: {}", e);
+                    return Err(BitcoinCoreSv2Error::CapnpError(e));
+                }
+            };
```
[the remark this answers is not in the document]
Shourya742: We can implement Display for our error types so that this message is automatically included when the error is formatted.

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+                                    match self_clone.outgoing_messages.send(TemplateDistribution::NewTemplate(non_future_template.clone())).await {
+                                        Ok(_) => (),
+                                        Err(e) => {
+                                            tracing::error!("Failed to send future NewTemplate message: {:?}", e);
+                                            tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                                            self_clone.global_cancellation_token.cancel();
+                                            break;
+                                        }
+                                    }
+                                }
```
Shourya742:
```suggestion
                                   if let Err(e) =  self_clone.outgoing_messages.send(TemplateDistribution::NewTemplate(non_future_template.clone())).await {
                                            tracing::error!("Failed to send future NewTemplate message: {:?}", e);
                                            tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
                                            self_clone.global_cancellation_token.cancel();
                                            break;
                                        }
                                    }
                                
```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+                let template_ipc_client =
+                    match self_clone.current_template_ipc_client.borrow().clone() {
+                        Some(template_ipc_client) => template_ipc_client,
+                        None => {
+                            tracing::error!("Template IPC client not found");
+                            tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                            self_clone.global_cancellation_token.cancel();
+                            return;
+                        }
+                    };
```
Shourya742:
```suggestion
                    let Some(template_ipc_client) =  self_clone.current_template_ipc_client.borrow().clone() else {
                            tracing::error!("Template IPC client not found");
                            tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
                            self_clone.global_cancellation_token.cancel();
                            return;
                    };
```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+                Ok(_) => {
+                    tracing::debug!("Successfully sent initial SetNewPrevHash message");
+                }
+                Err(e) => {
+                    tracing::error!("Failed to send set new prev hash message: {:?}", e);
+                    tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                    self.global_cancellation_token.cancel();
+                    return;
+                }
+            }
```
Shourya742:
```suggestion
            if let Err(e) = self
                                .outgoing_messages
                                .send(TemplateDistribution::SetNewPrevHash(set_new_prev_hash.clone()))
                                .await
                                {
                                   tracing::error!("Failed to send SetNewPrevHash message: {e:?}");
                                   tracing::warn!("Terminating Sv2 Bitcoin Core IPC connection");
                                   self.global_cancellation_token.cancel();
                                   return;
                               }

                          tracing::debug!("Successfully sent initial SetNewPrevHash message");

```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+                Ok(_) => {
+                    tracing::debug!("Successfully sent initial NewTemplate message");
+                }
+                Err(e) => {
+                    tracing::error!("Failed to send future template message: {:?}", e);
+                    tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                    self.global_cancellation_token.cancel();
+                    return;
+                }
+            }
```
Shourya742:
```suggestion
            if let Err(e) = self
                                 .outgoing_messages
                                 .send(TemplateDistribution::NewTemplate(future_template.clone()))
                                 .await
                                {
                                   tracing::error!("Failed to send NewTemplate message: {e:?}");
                                    tracing::warn!("Terminating Sv2 Bitcoin Core IPC connection");
                                    self.global_cancellation_token.cancel();
                                   return;
                                }

                            tracing::debug!("Successfully sent initial NewTemplate message");
```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+                                    coinbase_output_constraints.coinbase_output_max_additional_size,
+                                    coinbase_output_constraints.coinbase_output_max_additional_sigops);
+                                match self_clone.handle_coinbase_output_constraints(coinbase_output_constraints).await {
+                                    Ok(_) => (),
+                                    Err(e) => {
+                                        tracing::error!("Failed to handle coinbase output constraints: {:?}", e);
+                                        tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                                        self_clone.global_cancellation_token.cancel();
+                                        break;
+                                    }
```
Shourya742:
```suggestion
                                if let Err(e) =  self_clone.handle_coinbase_output_constraints(coinbase_output_constraints).await 
                                {
                                        tracing::error!("Failed to handle coinbase output constraints: {:?}", e);
                                        tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
                                        self_clone.global_cancellation_token.cancel();
                                        break;
                                    }
```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,1138 @@
...
+        // todo: remove this once https://github.com/bitcoin/bitcoin/issues/33575 is implemented
+        loop {
+            if self.pending_fetch.load(Ordering::SeqCst) {
+                tracing::debug!("Pending fetch detected, waiting for 100ms");
+                tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+            } else {
+                self.pending_fetch.store(true, Ordering::SeqCst);
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
plebhash: I see how a lock could be a more elegant solution for this

but this is merely a temporary workaround (the referenced issue has already been closed), and I don't see any meaningful performance hits because of this approach

so I'd prefer not to spend too much time aiming for optimizations around it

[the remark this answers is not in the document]
plebhash: https://github.com/stratum-mining/sv2-apps/issues/81

bitcoin-core-sv2/src/template_data.rs
```diff
@@ -0,0 +1,333 @@
...
+                let hash_bytes: [u8; 32] = *current_level[target_index].as_byte_array();
+                let u256_hash = U256::try_from(hash_bytes.to_vec())
+                    .expect("32-byte hash should always be valid for U256");
+                merkle_path.push(u256_hash);
+            }
+
+            // Calculate the next level of the merkle tree
+            for i in (0..current_level.len()).step_by(2) {
+                let left = current_level[i];
+                let right = if i + 1 < current_level.len() {
```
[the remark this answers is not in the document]
Shourya742: The snippet I shared above is just one way to define the tree, we could also go with a boxed version, something like this:

```rust
#[derive(Clone, Debug)]
pub enum MerkleTree {
    Leaf(sha256d::Hash),
    Node {
        left: Box<MerkleTree>,
        right: Box<MerkleTree>,
        hash: sha256d::Hash,
    },
}
```

Traits aren’t only for expressing shared behavior; they’re also a great way to leverage Rust’s type system. That said, I’m not too opinionated on how to implement it, but we do need some abstraction here.

pool-apps/pool/src/lib/template_receiver/bitcoin_core.rs
```diff
@@ -0,0 +1,170 @@
...
+                "Failed to send CoinbaseOutputConstraints message to channel manager: {:?}",
+                e
+            );
+            handle_error(
+                &status_sender,
+                PoolError::FailedToSendCoinbaseOutputConstraints,
+            )
+            .await;
+        }
+    };
```
Shourya742:
```suggestion
    if let Err(e) =  incoming_tdp_sender
        .send(TemplateDistribution::CoinbaseOutputConstraints(
            coinbase_output_constraints,
        ))
        .await
    {
            tracing::error!(
                "Failed to send CoinbaseOutputConstraints message to channel manager: {:?}",
                e
            );
            handle_error(
                &status_sender,
                PoolError::FailedToSendCoinbaseOutputConstraints,
            )
            .await;
    };
```

bitcoin-core-sv2/src/template_data.rs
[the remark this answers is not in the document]
plebhash: added a new `merkle_path.rs` module

xyephy (commented):
Concept ACK

This is a major architectural enhancement that adds significant flexibility to the SV2 ecosystem, allowing Pool and JDC applications to connect directly to Bitcoin Core v30+ via UNIX socket instead of requiring a separate Stratum V2 Template Provider. I have tested locally and everything compiles fine without errors. However, I have a few minor suggestions to consider before merge:

#### 1. Replace `.expect()` calls with proper error handling

***Locations:*** `bitcoin-core-sv2/src/template_data.rs` lines 2531, 2539, 2548, 2573, 2584

Multiple `.expect()` calls will cause panics if assumptions are violated. While Bitcoin Core should provide valid data, defensive programming is essential for production code.

***Current code (line 2531-2533):***
````rust
fn get_version(&self) -> u32 {
    self.block.header.version.to_consensus()
        .try_into()
        .expect("block version conversion to u32 should never fail")
}
````

***Suggested fix:***
````rust
fn get_version(&self) -> Result<u32, TemplateDataError> {
    self.block.header.version.to_consensus()
        .try_into()
        .map_err(|_| TemplateDataError::InvalidBlockVersion)
}
````

***Functions needing updates:***
- `get_version()`
- `get_coinbase_tx_version()`
- `get_coinbase_script_sig()`
- `get_serialized_empty_coinbase_outputs()`
- `get_coinbase_tx_value_remaining()`

Then propagate `Result` types through `get_new_template_message()` and its callers.


#### 2. Add unit tests for merkle path computation

***File:*** `bitcoin-core-sv2/src/merkle_path.rs`

Suggested tests:
````rust
#[cfg(test)]
mod tests {
    use super::*;
    use stratum_core::bitcoin::hashes::sha256d;

    #[test]
    fn test_merkle_path_single_coinbase_tx() {
        // Should return empty path for coinbase-only block
        let coinbase_hash = sha256d::Hash::hash(&[0u8; 32]);
        let tree = MerkleTree::build(vec![coinbase_hash]);
        let path = tree.extract_coinbase_path();
        assert_eq!(path.len(), 0);
    }

    #[test]
    fn test_merkle_path_two_txs() {
        // Basic two-transaction merkle tree
        let tx1 = sha256d::Hash::hash(&[1u8; 32]);
        let tx2 = sha256d::Hash::hash(&[2u8; 32]);
        let tree = MerkleTree::build(vec![tx1, tx2]);
        let path = tree.extract_coinbase_path();
        assert_eq!(path.len(), 1);
        assert_eq!(path[0], tx2);
    }

    #[test]
    fn test_merkle_path_multiple_txs() {
        // Test with several transactions to verify path correctness
        let txs: Vec<_> = (0..8)
            .map(|i| sha256d::Hash::hash(&[i; 32]))
            .collect();
        let tree = MerkleTree::build(txs);
        let path = tree.extract_coinbase_path();
        assert_eq!(path.len(), 3); // log2(8) = 3
    }
}
````


#### 3. Document Bitcoin Core issue #33575 workaround

***File:*** `bitcoin-core-sv2/README.md`

The codebase has extensive workarounds for the inability to cancel `waitNext` requests. This needs documentation so users understand the 50-second shutdown delay.

***Suggested addition to README:***
````markdown
## Known Limitations

### Shutdown Delay

Due to a limitation in Bitcoin Core's IPC implementation (see [bitcoin/bitcoin#33575](https://github.com/bitcoin/bitcoin/issues/33575)), `waitNext` requests cannot be cancelled. This means:

- On shutdown, the library must wait for pending `waitNext` requests to complete
- Maximum shutdown time is capped at 50 seconds
- Counters and locking are used internally to track request completion

This workaround will be removed once Bitcoin Core implements proper cancellation support.
````


#### 4. Security note for UNIX socket permissions

Consider adding to the README:
````markdown
## Security Considerations

Ensure the Bitcoin Core UNIX socket has restrictive permissions:
```bash
chmod 600 /path/to/bitcoin/node.sock
```

Only the Pool/JDC process user should have access to prevent unauthorized template requests.
````

#### 5. Integration test for stale templates

Consider adding a test that verifies `RequestTransactionDataError` is properly sent when requesting data for stale template IDs (after chain tip change).

Otherwise this PR should be ready to merge

File: bitcoin-core-sv2/README.md [hunk 11]
```diff
@@ -0,0 +1,26 @@
...
+
+A Rust library that integrates [Bitcoin Core](https://bitcoin.org/en/bitcoin-core/) with the [Stratum V2 Template Distribution Protocol](https://github.com/stratum-mining/sv2-spec/blob/main/07-Template-Distribution-Protocol.md) via IPC over a UNIX socket.
+
+## Overview
+
+`bitcoin_core_sv2` allows for the official Bitcoin Core distribution to be leveraged for the following use-cases:
+- building Sv2 applications that act as a Client under the Template Distribution Protocol (e.g.: Pool or JDC) while connecting directly to the Bitcoin Core node.
+- building a Sv2 Template Provider application that acts as a Template Distribution Protocol Server while creating templates from a Bitcoin Core node.
+
+## Important Notes
```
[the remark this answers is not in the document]
plebhash: a9d1b673888ad3ea692f78ee8d7aaba8c99bb71d

plebhash:
@xyephy thanks

I addressed points 1 and 2.

for 3, the shutdown delay only really happens on the `logger` examples because we're blocking on `BitcoinCoreSv2::run`

overall, I don't really think this deserves a section on `README.md`... if someone wants to dive into the source code, they will see the comments with references to https://github.com/stratum-mining/sv2-apps/issues/81 and https://github.com/bitcoin/bitcoin/issues/33575

for 4, I don't really see the value in changing permissions of `node.sock`... this file is created ephemerally at runtime by `bitcoin-node` and if there's security concerns around its permissions, that should be handled by `bitcoin-node` itself, and not manually by the user

for 5, we still don't have proper infrastructure to integrate this crate on Integration Tests

that's probably something to be covered on #71, I took note here https://github.com/stratum-mining/sv2-apps/issues/71#issuecomment-3523464104

bitcoin-core-sv2/src/lib.rs
plebhash: choosing to leave this as-is, because we're still using the `Ok()` arm

bitcoin-core-sv2/src/handlers.rs
```diff
@@ -0,0 +1,185 @@
...
+use tokio_util::sync::CancellationToken;
+
+impl BitcoinCoreSv2 {
+    pub async fn handle_coinbase_output_constraints(
+        &mut self,
+        coinbase_output_constraints: CoinbaseOutputConstraints,
+    ) -> Result<(), BitcoinCoreSv2Error> {
+        tracing::debug!("handle_coinbase_output_constraints() called");
+        tracing::debug!("Cancelling template_ipc_client_cancellation_token");
+        self.template_ipc_client_cancellation_token.cancel();
```
GitGab19: Can you add a comment to explain why we're calling `.cancel()` here?

plebhash: to further clarify, the current strategy for killing the loop inside `monitor_ipc_templates` task is to rely on `self.coinbase_output_constraints_counter`

on every iteration of the loop, we check `self.coinbase_output_constraints_counter` against what was the original value when this instance of `monitor_ipc_templates` task was spawned

if the value changed, we kill the loop, because there's new Coinbase constraints and the templates coming from this task are no longer relevant

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+        global_cancellation_token: CancellationToken,
+    ) -> Result<Self, BitcoinCoreSv2Error> {
+        info!(
+            "Creating new Sv2 Bitcoin Core Connection via IPC over UNIX socket: {}",
+            bitcoin_core_unix_socket_path.display()
+        );
+
+        let stream = UnixStream::connect(bitcoin_core_unix_socket_path)
+            .await
+            .map_err(|_| {
```
xyephy: noticed some error context loss here which might bring a real ux issue because   `.map_err(|_| {`  throws away actual capnp error. When connection fails (wrong path, permissions, bitcoind not running), users get generic
  message instead of actual capnp error details. This makes debugging harder.
  **suggested fix:**
```rust
.map_err(|e| {
    tracing::error!("Socket connect failed: {}", e);
    BitcoinCoreSv2Error::CannotConnectToUnixSocket(
        bitcoin_core_unix_socket_path.into(),
        e.to_string()  // preserve it
    )
})?;
```

Update error enum to include the string:
```rust
// in error.rs
CannotConnectToUnixSocket(Box<Path>, String),
```

And Display impl:
```rust
BitcoinCoreSv2Error::CannotConnectToUnixSocket(path, err) => {
    write!(f, "Cannot connect to Unix socket at {}: {}", path.display(), err)
}
```

bitcoin-core-sv2/src/monitors.rs
```diff
@@ -0,0 +1,425 @@
...
+                                let mut template_data_guard = match self_clone.template_data.write() {
+                                    Ok(guard) => guard,
+                                    Err(e) => {
+                                        tracing::error!("Failed to acquire write lock on template_data: {:?}", e);
+                                        tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                                        self_clone.global_cancellation_token.cancel();
+                                        break;
+                                    }
+                                };
+                                template_data_guard.insert(new_template_data.get_template_id(), new_template_data.clone());
```
xyephy: noticed templates keep accumulating on mempool fee changes:
```rust
template_data_guard.insert(new_template_data.get_template_id(), new_template_data.clone());
```
and only cleared on chain tip change (line 262), not fee changes. During volatile mempool periods this grows unbounded.

**suggested fix:** Add `MAX_TEMPLATES` constant (say 100) and evict oldest when hit or periodic cleanup for old templates.

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+///
+/// Incoming [`SubmitSolution`] messages are used to submit solutions to a specific template.
+#[derive(Clone)]
+pub struct BitcoinCoreSv2 {
+    fee_threshold: u64,
+    thread_map: ThreadMapIpcClient,
+    thread_ipc_client: ThreadIpcClient,
+    mining_ipc_client: MiningIpcClient,
+    current_template_ipc_client: Rc<RefCell<Option<BlockTemplateIpcClient>>>,
+    current_prev_hash: Rc<RefCell<Option<U256<'static>>>>,
```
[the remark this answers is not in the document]
xyephy: thank you for clarifying, now I see why its  single-threaded.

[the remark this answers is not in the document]
plebhash: as @Shourya742 pointed out, this limitation stems from `capnp` crate itself, where multiple types are limited by the absence of `Send` implementations

this has been pointed out by multiple users of the crate:
- https://github.com/capnproto/capnproto-rust/issues/558
- https://github.com/capnproto/capnproto-rust/issues/180
- https://github.com/capnproto/capnproto-rust/issues/256

while this seems to be a popular feature request, one thing to note on those issues is that some are around ~5 years old and still haven't been addressed... so I don't think it's realistic to expect this to change any time soon, and therefore bounding `BitcoinCoreSv2::run` to a single-threaded environment becomes a hard-prerequisite on the design of this crate

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+            })?;
+
+        let coinbase_weight = (coinbase_output_max_additional_size * WEIGHT_FACTOR) as u64;
+        let block_reserved_weight = coinbase_weight.max(MIN_BLOCK_RESERVED_WEIGHT); // 2000 is the minimum block reserved weight
+        tracing::debug!("Setting block_reserved_weight: {block_reserved_weight}");
+        template_ipc_client_request_options.set_block_reserved_weight(block_reserved_weight);
+        template_ipc_client_request_options.set_coinbase_output_max_additional_sigops(
+            coinbase_output_max_additional_sigops as u64,
+        );
+        template_ipc_client_request_options.set_use_mempool(true);
```
GitGab19: What does this do?

bitcoin-core-sv2/src/merkle_path.rs
```diff
@@ -0,0 +1,157 @@
...
+        assert_eq!(path[0], tx2);
+    }
+
+    #[test]
+    fn test_merkle_path_multiple_txs() {
+        // Test with several transactions to verify path correctness
+        let txs: Vec<_> = (0..8).map(|i| sha256d::Hash::hash(&[i; 32])).collect();
+        let tree = MerkleTree::build(txs);
+        let path = tree.extract_coinbase_path();
+        assert_eq!(path.len(), 3); // log2(8) = 3
```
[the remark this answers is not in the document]
plebhash: this module is about to be removed entirely

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+        let mut template_ipc_client_request_options = template_ipc_client_request
+            .get()
+            .get_options()
+            .map_err(|e| {
+                tracing::error!("Failed to get template IPC client request options: {e}");
+                e
+            })?;
+
+        let coinbase_weight = (coinbase_output_max_additional_size * WEIGHT_FACTOR) as u64;
+        let block_reserved_weight = coinbase_weight.max(MIN_BLOCK_RESERVED_WEIGHT); // 2000 is the minimum block reserved weight
```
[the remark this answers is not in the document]
plebhash: 2000 is what Bitcoin Core used to do before IPC, I'm not sure what was the rationale there

we've been having discussions on potentially smaller values, but it's been hard to reach consensus there:

https://github.com/stratum-mining/sv2-tp/pull/51

https://github.com/stratum-mining/sv2-spec/pull/163#discussion_r2476690382

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+                "wait_next_request_counter: {}",
+                self.wait_next_request_counter.load(Ordering::SeqCst)
+            );
+            tracing::debug!(
+                "Shutdown: wait_next_request_counter={}, elapsed={}s",
+                counter_value,
+                elapsed
+            );
+
+            if counter_value == 0 || elapsed > 50 {
```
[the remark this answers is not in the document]
plebhash: on v30 (before https://github.com/bitcoin/bitcoin/issues/33554 and https://github.com/bitcoin/bitcoin/issues/33575 were solved), Bitcoin Core could potentially crash if we exited `BitcoinCoreSv2::run` (or most importantly: dropped `BitcoinCoreSv2` and all it's `capnp` objects) before all `waitNext` requests had been resolved

we're using (arbitrary) 10s for `waitNext` timeout, which means that if no new templates are returned within this interval, the request is automatically cancelled on Bitcoin Core side (and we don't need to worry about it on shutdown)

once the `global_cancellation_token` is activated, we know for sure no new `waitNext` requests will be created, but we still need to wait until all of them are either returned or timed out (at least before we can start using `interruptWait` in v30.x)

due to the fact that (even if extremely rare) we could receive multiple `CoinbaseOutputConstraints`, we can have more than one `waitNext` request happening at the same time

worst case would be a `waitNext` request that was sent immediately before the `global_cancellation_token` is activated, for which we would need to wait 10s plus some extra

in order to be extra conservative, I wanted this extra to be one order of magnitude bigger than 10s, so I arrived at 50s, which is definitely overkill, and `counter_value` should always kick in before it's needed

but also, hopefully this nasty hack will go away as soon as v30.1 is out, when we'll be able to leverage `interruptWait`, which will:
- allow us to avoid having multiple `waitNext` running at the same time
- allow us to kill the (single running) `waitNext` for shutdowbn

plebhash:
I refactored `bitcoin_core_sv2` crate to mitigate the issues @Sjors pointed out [here](url) and [here](https://github.com/bitcoin/bitcoin/pull/33819#issuecomment-3541021155)

when `waitNext` returns, we want to send a `NewTemplate` as soon as possible

the previous design was using `getBlock()` to get:
- header
- coinbase tx
- txdata (then used for merkle_path)

this is arguably suboptiomal, given that we already have `getHeader`, `getCoinbaseTx` and `getCoinbaseMerklePath`

even though communication over a local UNIX socket should be very fast, we're still fetching up-to 4MB of data while a few bytes (or kb at most) are sufficient for crafting the `NewTemplate` ASAP

so it's better to use each granular method, and leave `getBlock` for when we receive a `RequestTransactionData`

the only app who could send this message is a JDC, so the extra delays introduced by `getBlock` would only delay the subsequent `DeclareMiningJob` it wants to send to JDS, while it's likely already hashing on `NewTemplate` and caching shares

the changes from this refactor are on 8a232afd211c5e78d230b35c69a46579a4f94cd6, which I'll probably squash into 56b6f1b077cd00683cf112550608527192986d6a soon

---

also, this means that we'll likely need to replace `getCoinbaseTx` with `getCoinbase` soon

ideally, as pointed out by @Sjors [here](https://github.com/bitcoin/bitcoin/pull/33819#issuecomment-3504056653), we should try to use `getCoinbase` (to be introduced in v30.x) and fall back to `getCoinbaseTx` if Bitcoin Core is v30

but we still don't have `getCoinbase` on [`bitcoin-capnp-types`](https://github.com/2140-dev/bitcoin-capnp-types/blob/master/capnp/mining.capnp), so we'll have to wait before this is implemented

bitcoin-core-sv2/src/monitors.rs
```diff
@@ -0,0 +1,421 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+            tracing::debug!("monitor_ipc_templates() task exiting");
+        });
+    }
+
+    pub fn monitor_incoming_messages(&self) {
```
GitGab19: Can you add some docs here?

bitcoin-core-sv2/src/template_data.rs
```diff
@@ -0,0 +1,319 @@
...
+        set_new_prev_hash.into_static()
+    }
+
+    pub async fn get_request_transaction_data_success_message(
+        &self,
+        thread_ipc_client: ThreadIpcClient,
+    ) -> Result<RequestTransactionDataSuccess<'static>, TemplateDataError> {
+        let request_transaction_data_success = RequestTransactionDataSuccess {
+            template_id: self.template_id,
+            transaction_list: self.get_tx_data(thread_ipc_client).await?,
```
GitGab19: Oh I guess this is related to the discussion here: https://github.com/stratum-mining/sv2-apps/pull/59/files#r2535012571

So you're basically asking full template data to Bitcoin Core only when you receive `RequestTransactionData`, and you removed that from the template data, right?

miner-apps/jd-client/src/lib/template_receiver/bitcoin_core.rs
```diff
@@ -0,0 +1,166 @@
...
+                    break;
+                }
+            }
+        }
+    });
+
+    let status_sender_clone = status_sender.clone();
+
+    // spawn a dedicated thread to run the BitcoinCoreSv2 instance
+    // because we're limited to tokio::task::LocalSet
```
GitGab19: Can you detail that the limitation comes from `capnp-rpc`?

bitcoin-core-sv2/src/monitors.rs
```diff
@@ -0,0 +1,421 @@
+use crate::BitcoinCoreSv2;
+
+use std::{collections::HashSet, sync::atomic::Ordering};
+use stratum_core::parsers_sv2::TemplateDistribution;
+use tracing::info;
+
+impl BitcoinCoreSv2 {
+    pub fn monitor_ipc_templates(&self) {
```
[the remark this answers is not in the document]
plebhash: done

pool-apps/pool/config-examples/pool-config-bitcoin-core-ipc-example.toml
```diff
@@ -0,0 +1,33 @@
...
+
+# Pool signature (string to be included in coinbase tx)
+pool_signature = "Stratum V2 SRI Pool"
+
+# Enable this option to set a predefined log file path.
+# When enabled, logs will always be written to this file.
+# The CLI option --log-file (or -f) will override this setting if provided.
+# log_file = "./pool.log"
+
+# How many shares we expect to receive in a minute (determintes difficulty targets)
```
GitGab19:
```suggestion
# How many shares we expect to receive in a minute (determines difficulty targets)
```

pool-apps/pool/src/lib/template_receiver/sv2_tp/mod.rs
GitGab19: Couldn't we do a git mv here?

pool-apps/pool/src/lib/template_receiver/bitcoin_core.rs
```diff
@@ -0,0 +1,169 @@
...
+                    break;
+                }
+            }
+        }
+    });
+
+    let status_sender_clone = status_sender.clone();
+
+    // spawn a dedicated thread to run the BitcoinCoreSv2 instance
+    // because we're limited to tokio::task::LocalSet
```
GitGab19: Can you detail that the limitation comes from `capnp-rpc` in `bitcoin_core_sv2`?

pool-apps/pool/src/lib/template_receiver/sv2_tp/mod.rs
plebhash: I did, commit 6b982c9059c393b47a375ceb3013d0700f8a7ae6

bitcoin-core-sv2/examples/logger.rs
```diff
@@ -0,0 +1,164 @@
...
+    template_distribution_sv2::{CoinbaseOutputConstraints, RequestTransactionData},
+};
+use tokio_util::sync::CancellationToken;
+use tracing::info;
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt::init();
+
+    // the user must provide the path to the Bitcoin Core UNIX socket
```
Sjors: Would be a nice followup to find the default socket path based on just the network. Though a custom path should always be possible.

https://en.bitcoin.it/wiki/Data_directory

One caveat, especially for CI: different operating systems have a fairly short maximum length for unix socket paths. See e.g. https://github.com/bitcoin/bitcoin/blob/115d298a9fa31ffacd980381e2f459c9440d8833/test/functional/test_framework/test_node.py#L139-L148

bitcoin-core-sv2/src/handlers.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        tracing::debug!("Submitting solution to Bitcoin Core");
+        let Err(e) = template_data
+            .submit_solution(submit_solution, self.thread_ipc_client.clone())
+            .await
+        else {
+            tracing::debug!("Solution submitted successfully");
+            return Ok(());
+        };
+        tracing::error!("Failed to submit solution: {:?}", e);
```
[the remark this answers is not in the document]
plebhash: that feels more like an advanced feature to be brainstormed and discussed to be potentially added in the future?

I'm not sure we really need this in the initial iterations

plebhash:
@GitGab19 reported today of crashing Bitcoin Core after killing Pool/JDC via ctrl+c

here's an analysis of why that was happening, and how it was fixed:

## context: OS thread spawning

we can only do `BitcoinCoreSv2::run` inside a `tokio::task::LocalSet`, which means this **has** to be blocking.

on integrations with Pool and JDC, accommodating for a blocking task would be too disruptive and it would require deep architectural refactors

so the compromise is to spawn a dedicated OS thread so we can block on `BitcoinCoreSv2::run` without affecting the rest of Pool/JDC runtime

## context: v30 bug

on https://github.com/bitcoin/bitcoin/issues/33554 I reported that Bitcoin Core would crash (similarly to what we saw here)... after some investigation we realized that our Rust code cannot terminate the IPC connection before all pending `waitNext` calls are fully completed

that led to https://github.com/bitcoin/bitcoin/issues/33575, which is set to land on v30.1, and will allow us to explicitly kill pending `waitNext` calls 

but while we're stuck with v30, we need to **wait** until either `waitNext` returns with a new template, or it completes its timeout

## root cause

after `global_cancellation_token` is activated, `BitcoinCoreSv2::run` enters a loop where we observe some counters of pending `waitNext` calls so that we only exit the function after all of them are completed

the problem is that this was happening inside a dedicated OS thread that was spawned in a "fire-and-forget" fashion, without any `JoinHandle` that would allow us to wait for its completion... so when we did ctrl+c on the main Pool/JDC app, the entire app process would halt and the dedicated OS thread would be dropped before `BitcoinCoreSv2::run` had the chance to wait on all pending `waitNext` calls

## solution

37fbf1bd165906b2de25136f1a643f33eeb33c86 fixes the issue by introducing a `JoinHandle` on Pool and JDC, which we `.join` during the app graceful shutdown routines to make sure we don't halt before `BitcoinCoreSv2::run` has finished its internal cleanup routines.

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,614 @@
...
+//! ### `LocalSet` Requirement
+//!
+//! Due to limitations in the `capnp-rpc` dependency, [`BitcoinCoreSv2`] must be run within a
+//! [`tokio::task::LocalSet`]. The crate examples demonstrate the proper setup pattern.
+//!
+//! ### Fee Threshold
+//!
+//! The `fee_threshold` parameter (in satoshis) determines when a new template is distributed due
+//! to mempool changes. When the mempool fee delta exceeds this threshold, a new `NewTemplate`
+//! message is sent.
```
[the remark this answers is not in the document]
plebhash: introduced via 59df241f44ebfda12e469cf4bad04d4a29aaf451

bitcoin-core-sv2/src/monitors.rs
GitGab19: > I know we agreed on our call a few hours ago on introducing a constant to start dropping new templates, but now I'm starting to lean towards not doing that again...

Well, given the correct numbers you shared, I don't think we need any constant anymore.

Also, in your computation you're considering one template every second (which is not doable in real scenarios), so if we consider a more concrete example of one template every 10s, the computation becomes: (96+149+512) * 6 * 144 = ~0,65MB.

We can definitely proceed without any `MAX_TEMPLATES` constant and rest easy.

plebhash:
adding 84d48c846d6dc8420e7501f4e2053ffb7fbb09a9 to follow https://github.com/stratum-mining/sv2-tp/pull/61

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+                                    self.global_cancellation_token.cancel();
+                                    return;
+                                }
+                            };
+
+                            let mut current_template_ipc_client_guard = self.current_template_ipc_client.borrow_mut();
+                            *current_template_ipc_client_guard = Some(template_ipc_client);
+                            tracing::debug!("Set current_template_ipc_client to initial template");
+
+                            self.coinbase_output_constraints_counter.fetch_add(1, Ordering::SeqCst);
```
Shourya742: We can live with Relaxed ordering considering we are in single threaded environment.

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+                    }
+                }
+            }
+        }
+
+        // bootstrap the first template
+        {
+            tracing::debug!("Bootstrapping first template...");
+            let template_data = match self
+                .fetch_template_data(self.thread_ipc_client.clone())
```
Shourya742: Do we need to pass `thread_ipc_client` to `fetch_template_data`?

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+                        tracing::error!("Failed to acquire read lock on template_data: {:?}", e);
+                        tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                        self_clone.global_cancellation_token.cancel();
+                        return;
+                    }
+                };
+
+                let mut removed_template_data: Vec<TemplateData> = Vec::new();
+
+                for stale_template_id in stale_template_ids.clone() {
```
Shourya742:
```suggestion
                for stale_template_id in &stale_template_ids {
```

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+            .get_context()?
+            .set_thread(thread_ipc_client.clone());
+
+        let coinbase_tx_bytes = coinbase_tx_request
+            .send()
+            .promise
+            .await?
+            .get()?
+            .get_result()?
+            .to_vec();
```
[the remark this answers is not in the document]
Shourya742: Hmm, considering the reason why futures were introduced to capnp RPC system, was to mitigate this exact issue of waiting on multiple calls. I will verify this locally and update here.

bitcoin-core-sv2/src/template_data.rs
```diff
@@ -0,0 +1,428 @@
...
+            return Err(TemplateDataError::InvalidSolution);
+        }
+        if solution_coinbase_tx.input[0].witness != self.coinbase_tx.input[0].witness {
+            return Err(TemplateDataError::InvalidSolution);
+        }
+        if solution_coinbase_tx.input[0].previous_output
+            != self.coinbase_tx.input[0].previous_output
+        {
+            return Err(TemplateDataError::InvalidSolution);
+        }
```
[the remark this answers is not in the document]
plebhash: I checked `sv2-tp` and it's not fully validating the solution before sending it to Bitcoin Core

I guess that makes sense, in a way we're validating it 2x (once here and once in Bitcoin Core), which feels superfluous and introduced unnecessary delays to a hot execution path

I'm still transfering those validations to `dump_solution_to_disk`, because it wouldn't make sense to write a bad solution to disk... but there's no problem in introducing delays on this execution path

253390e866954738596a98202516ff1acad333a9

bitcoin-core-sv2/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+                if let Err(e) = template_data
+                    .destroy_ipc_client(self_clone.thread_map.clone())
+                    .await
+                {
+                    tracing::error!("Failed to destroy template IPC client: {:?}", e);
+                    tracing::warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                    self_clone.global_cancellation_token.cancel();
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
plebhash: tbh I don't think executing this sequentially is so bad

first, this is merely freeing up memory and it's not a critical execution path

second, each `destroy_ipc_client` will send a `make_thread_request`, which will inevitably have to be executed sequentially on Bitcoin Core side, leaving only the execution of each `destroy_request` to be done concurrently

so executing this concurrently is only a marginal optimization, at best

but sure, 47a222191100a89b2e7f6e38035517c6987bfcba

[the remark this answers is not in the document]
Shourya742: Oh, correct this is different. Let me check the latency benefits if any, we can squeeze without breaking things up.

bitcoin-core-sv2/src/lib.rs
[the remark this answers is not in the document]
plebhash: tbh I worry that trying to introduce `capnp_futures` would require major refactors into a completely different paradigm

is this rabbit hole really worth it?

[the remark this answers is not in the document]
plebhash: apparently `sv2-tp` does these 3 calls sequentially as well: https://github.com/Sjors/sv2-tp/blob/fe42cf39f4c518b6e7953e4429c24dbdfdc508eb/src/sv2/template_provider.cpp#L565

@Sjors is this something we want to optimize for? or would it be overkill?

---

I just pushed e310d12a5ffbfa6d3aa29eaee481e93c8af33f58 which makes `BitcoinCoreSv2` statically hold one dedicated `ThreadIpcClient` for each request category:
- `submitSolution`
- `getHeader`
- `getCoinbaseTx`
- `getCoinbaseMerklePath`
- `waitNext`

this way, we're guaranteed to always get isolated (and potentially concurrent) execution on Bitcoin Core side (without the delay of instantiating a new `ThreadIpcClient` every time we need to send one of these requests, which would be a silly tradeoff)

but I'm still a bit on the fence about this, because now, for each `BitcoinCoreSv2` instance, we have 5 dedicated threads on Bitcoin Core side

prior to this, we could get away with only 2:
- one for `submitSolution` (which is initially used for bootstrapping and then becomes dedicated to this request)
- one for `waitNext`+`getHeader`+`getCoinbaseTx`+`getCoinbaseMerklePath` (where everything happens sequentially)

(this is all ignoring the `ThreadIpcClient` instances we spawn ad-hoc for `getBlock` and `destroy`, which are all dropped immediately after they're used... this happens on both approaches but is not too relevant for the point I'm trying to articulate here)

the reason why this worries me is that spawning OS threads are more expensive than what we're used to in tokio-land

and once we move into the sidecar implementation, we will instantiate one new `BitcoinCoreSv2` for each client that connects to it (with 5 `ThreadIpcClient` instead of only 2)... so I wonder if this could scale badly on the Bitcoin Core side, leading to resource starvation 2.5x faster than necessary?

lucasbalieiro:
The apps survived:

* **24 hours on signet**, mined blocks successfully
* **12 hours on mainnet**
* **12 hours on testnet4**

During this time, I hit **three sudden crashes**, one on each chain.

After digging into it, it looks like the Bitcoin node just *dropped dead* with no obvious cause and naturally, when the node goes down, the whole stack follows.

Checking `debug.log` on all three chains, there’s **zero indication** of what happened: no error, no shutdown message, nothing odd. The node is doing business as usual… and then it flatlines. Last lines look like:

```bash
2025-11-22T12:45:09Z CreateNewBlock(): block weight: 3998523 txs: 4945 fees: 815256 sigops 775
2025-11-22T12:45:11Z New outbound-full-relay v2 peer connected: version: 70016, blocks=924707, peer=6213
2025-11-22T12:45:13Z CreateNewBlock(): block weight: 3998669 txs: 4946 fees: 822247 sigops 815
2025-11-22T12:45:14Z CreateNewBlock(): block weight: 3998698 txs: 4948 fees: 826372 sigops 841
2025-11-22T12:45:18Z CreateNewBlock(): block weight: 3998352 txs: 4941 fees: 829924 sigops 871
2025-11-22T12:45:19Z CreateNewBlock(): block weight: 3998550 txs: 4940 fees: 830855 sigops 889
2025-11-22T12:45:23Z CreateNewBlock(): block weight: 3998479 txs: 4895 fees: 882915 sigops 974
2025-11-22T12:45:24Z CreateNewBlock(): block weight: 3998756 txs: 4895 fees: 883703 sigops 983
2025-11-22T12:45:31Z CreateNewBlock(): block weight: 3998538 txs: 4893 fees: 884727 sigops 993
<EOF>
```

The app logs also don’t show anything interesting, just the fallout after the node dies. Before that, everything looks perfectly normal.

plebhash:
@lucasbalieiro thanks for reporting this

let's try to reproduce them again, now with stack traces:

on macOS I didn't have the option of using GDB, so I leveraged LLDB, for which the steps are:

- launch it via `lldb -- ./build/bin/bitcoin-node <usual args>`
- you'll drop into a console, for which you can just do `run`
- once it crashes, you do `thread backtrace all` inside the console

---

if you're on Ubuntu, you could use GDB, but then the command might be `thread apply all bt` instead (which I haven't validated)

you could probably use LLDB as well, and use the commands above that are already validated

---

for further reference: https://github.com/bitcoin/bitcoin/issues/33463#issuecomment-3347520481

plebhash:
> Do you suggest running the setup on a more beefy machine.

for now, this is the only solution I can imagine for running this with v30, although that's far from ideal.

ideally, Bitcoin Core should be able to run on low-end machines

---

I reported these out-of-memory crashes here: https://github.com/bitcoin/bitcoin/issues/33899#issuecomment-3568788623

I'm curious to see what people will say there when I check it again tomorrow, but for now my analysis is that this is not something for us to fix on our side

and if we proceed with the release, we should probably put out warnings of known bugs, since whatever bug this is, it's already baked into v30

Shourya742:
> > Do you suggest running the setup on a more beefy machine.
> 
> for now, this is the only solution I can imagine for running this with v30, although that's far from ideal.
> 
> ideally, Bitcoin Core should be able to run on low-end machines
> 
> I reported these out-of-memory crashes here: [bitcoin/bitcoin#33899 (comment)](https://github.com/bitcoin/bitcoin/issues/33899#issuecomment-3568788623)
> 
> I'm curious to see what people will say there when I check it again tomorrow, but for now my analysis is that this is not something for us to fix on our side
> 
> and if we proceed with the release, we should probably put out warnings of known bugs, since whatever bug this is, it's already baked into v30

@plebhash Should we reintroduce the destroy method and test it again? I don’t really expect it to change anything, and we may even see the crash occur earlier because of those ephemeral thread spawns.

pool-apps/pool/config-examples/pool-config-local-sv2-tp-example.toml
lucasbalieiro: running with this config file I got:

```bash
cargo run -- --config config-examples/pool-config-local-sv2-tp-example.toml
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.07s
     Running `/Users/lucasbalieiro/Projects/sv2-apps/pool-apps/target/debug/pool_sv2 --config config-examples/pool-config-local-sv2-tp-example.toml`

thread 'main' panicked at pool/src/args.rs:37:10:
Failed to load or deserialize config: missing field `supported_extensions`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

we are missing a few new fields

lucasbalieiro: I think this parameters got lost during the rebase:
```toml
# Protocol Extensions Configuration
# Extensions that the pool supports (will accept if requested by clients)
# Comment/uncomment to enable/disable specific extensions:
supported_extensions = [
    # 0x0002,  # Worker-Specific Hashrate Tracking
]

# Extensions that the pool requires (clients must support these to connect)
# Use with caution - requiring extensions may prevent some clients from connecting
required_extensions = [
    # Example: require Worker-Specific Hashrate Tracking
    # 0x0002,
]
```

plebhash: indeed I had some issues rebasing these, thanks for reporting

PR #67, opened 2025-11-05
JDC and JDS listen on `0.0.0.0` instead of `127.0.0.1`

### LINKED ISSUE #66: `jds-config-local-example.toml` has footgun on `listen_jd_address`
I guess this is somewhat subjective, but my interpretation of `*-local-*` is that I'm making the choice to make JDS point to a local TP

so this config file is my go-to when setting up the hosted JDS on the VPS

however, the following line is problematic:
https://github.com/stratum-mining/sv2-apps/blob/dcfeea1d793167f78de517ae3fd02a6f87cfd70b/pool-apps/jd-server/config-examples/jds-config-local-example.toml#L21

if we listen on `127.0.0.1`, it cannot be reached externally

while `jds-config-local-example.toml` does have `listen_jd_address = "0.0.0.0:34264"`, it also has `core_rpc_url =  "http://75.119.150.111"`, 

this also aligns with my interpretation: `*-hosted-*` means it connects to the hosted TP, without any other assumptions about where JDC connections are going to come from

TDLR: `jds-config-local-example.toml` should have `listen_jd_address = "0.0.0.0:34264"`

plebhash:
close #66

### DIFF
--- a/miner-apps/jd-client/config-examples/jdc-config-hosted-example.toml
+++ b/miner-apps/jd-client/config-examples/jdc-config-hosted-example.toml
@@ -1,5 +1,5 @@
 # SRI JDC config
-listening_address = "127.0.0.1:34265"
+listening_address = "0.0.0.0:34265"
 
 # Version support
 max_supported_version = 2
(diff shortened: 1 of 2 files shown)


lucasbalieiro (approved):
tACK

Tested locally with two separated computers in the same LAN and I was able to establish a connection

Shourya742 (commented):
Can we do a similar change for JDC.

Shourya742 (approved):
ACK

PR #76, opened 2025-11-06
Add request_id to downstream_id mapping in translator

Shourya742:
In the translator, when opening a channel, we were previously assigning the downstream_id to be the same as the message’s request_id. This was a bit hacky and also limited the number of downstreams to a u32 range (which is large, but still an unnecessary constraint). To address this, we’ve introduced a mapping between request_id and downstream_id, which exists only for the duration of the request’s lifetime on the upstream.

### DIFF
--- a/miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
@@ -39,7 +39,7 @@ use stratum_apps::{
         sv1_api::IsServer,
     },
     task_manager::TaskManager,
-    utils::types::{DownstreamId, Hashrate, RequestId, SharesPerMinute},
+    utils::types::{DownstreamId, Hashrate, SharesPerMinute},
 };
 use tokio::{
     net::TcpListener,
--- a/miner-apps/translator/src/lib/sv1/sv1_server/data.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/data.rs
@@ -1,14 +1,17 @@
 use crate::sv1::downstream::downstream::Downstream;
 use std::{
     collections::HashMap,
-    sync::{atomic::AtomicUsize, Arc, RwLock},
+    sync::{
+        atomic::{AtomicU32, AtomicUsize},
+        Arc, RwLock,
+    },
 };
 use stratum_apps::{
     stratum_core::{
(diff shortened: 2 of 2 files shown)


miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
```diff
@@ -462,10 +466,13 @@ impl Sv1Server {
...
                 );
-                let downstream_id = m.request_id;
-                let downstreams = self
-                    .sv1_server_data
-                    .super_safe_lock(|v| v.downstreams.clone());
+                let (downstream_id, downstreams) = self.sv1_server_data.super_safe_lock(|v| {
+                    let downstream_id = v
+                        .request_id_to_downstream_id
+                        .remove(&m.request_id)
+                        .expect("Downstream should exist for each request");
```
plebhash: what if a faulty server sends `OpenExtendedMiningChannelSuccess` with a non-existing `request_id`?

Shourya742: Added here: a1cb581

Shourya742: Ah, I really shouldn’t be writing comments late at night 😅. Yes, for errors like these, we’ll fall back but right now that also disconnects all downstreams connected to the translator (that’s how it currently works). The status management in the translator isn’t as refined as in JDC yet (though we do have the tracker to handle that). Once that’s fully integrated and the error orchestration gains better contextual awareness, we’ll be able to implement a much more precise fallback or disconnection logic across all role implementations.

GitGab19 (approved):

miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
plebhash: ok sorry for the delay here

so I followed the execution path and I can see how this edge case would trigger a `Sv1ServerShutdown`, which then leads to a `ShutdownMessage::ShutdownAll`

doesn't this mean tProxy simply shuts down, without trying to reconnect to a fallback server?

Shourya742: Yes, we need granular error orchestration for tproxy. I will open a PR for this soonish.

Issue #32, opened 2025-09-15
tProxy: Fallback after `SetupConnection.Error` and `OpenMiningChannel.Error` 


GitGab19:
When tProxy receives a `SetupConnection.Error`, and a `OpenMiningChannel.Error` from any upstream, it should fallback to a secondary one.

More context: https://github.com/stratum-mining/stratum/pull/1791#discussion_r2346875580, https://github.com/stratum-mining/stratum/pull/1791#discussion_r2346795935

PR #84, opened 2025-11-17
Tproxy fallback on SetupConnection.Error and OpenMiningMessage.Error

Shourya742:
closes: #32

### DIFF
--- a/integration-tests/tests/translator_integration.rs
+++ b/integration-tests/tests/translator_integration.rs
@@ -99,7 +99,7 @@ async fn test_translator_fallback_on_setup_connection_error() {
     let (pool_translator_sniffer_2, pool_translator_sniffer_addr_2) =
         start_sniffer("B", pool_addr_2, false, vec![], None);
 
-    let (_, _) = start_sv2_translator(
+    let (_, tproxy_addr) = start_sv2_translator(
         &[
             pool_translator_sniffer_addr_1,
             pool_translator_sniffer_addr_2,
@@ -107,6 +107,8 @@ async fn test_translator_fallback_on_setup_connection_error() {
         false,
     )
     .await;
+
+    let (_minerd_process, _minerd_addr) = start_minerd(tproxy_addr, None, None, false).await;
 
     pool_translator_sniffer_1
         .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
@@ -126,6 +128,19 @@ async fn test_translator_fallback_on_setup_connection_error() {
         .wait_for_message_type(
             MessageDirection::ToDownstream,
             MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+        )
+        .await;
+
+    pool_translator_sniffer_2
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_OPEN_EXTENDED_MINING_CHANNEL,
+        )
+        .await;
+    pool_translator_sniffer_2
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_OPEN_EXTENDED_MINING_CHANNEL_SUCCESS,
         )
         .await;
 }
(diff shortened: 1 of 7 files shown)


Shourya742:
> Can you add an integration test for this?

I am adding it, thanks for suggestion.

miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
```diff
@@ -598,19 +598,6 @@ impl Sv1Server {
...
-
-            Mining::OpenMiningChannelError(_) => {
-                todo!("Handle OpenMiningChannelError message from upstream");
-            }
-
-            Mining::UpdateChannelError(_) => {
-                todo!("Handle UpdateChannelError message from upstream");
-            }
-
             _ => unreachable!("Unexpected message type received from upstream"),
```
plebhash: maybe out of scope for this PR, but do we really want to panic here?

Shourya742: That’s provable, since we intercept the message in the channel manager before relaying it to the SV1 server. If needed, we can introduce an additional abstraction here to make that guarantee explicit.

plebhash: I guess that would making things extra safe, but on the other hand it feels like a lot of complexity overhead, while there's already plenty

I'd just add a new comment above the `unreachable` making this idea clear, and also change the message inside of it

the fact that it says `"Unexpected message type received from upstream"` suggests that this is a possible situation, which contradicts the reason why we're labeling this as `unreachable` in the first place

miner-apps/translator/src/lib/mod.rs
```diff
@@ -248,4 +219,107 @@ impl TranslatorSv2 {
...
+                i + 1,
+                upstream_len,
+                upstream_addr
+            );
+
+            tokio::time::sleep(Duration::from_secs(1)).await;
+
+            if upstream_addr.2 {
+                info!(
+                    "Upstream previously marked as malicious, skipping initial attempt warnings."
```
GitGab19: Let's rephrase this in a clear way: it seems to me that this flag means we already connected to this upstream.

miner-apps/translator/src/lib/mod.rs
```diff
@@ -248,4 +219,107 @@ impl TranslatorSv2 {
...
+        }
+
+        tracing::error!("All upstreams failed after {} retries each", MAX_RETRIES);
+        Err(TproxyError::Shutdown)
+    }
+}
+
+// Attempts to initialize a single upstream.
+#[allow(clippy::too_many_arguments)]
+async fn try_initialize_single(
```
GitGab19:
```suggestion
async fn try_initialize_upstream(
```

integration-tests/tests/translator_integration.rs
```diff
@@ -49,3 +65,139 @@ async fn translate_sv1_to_sv2_successfully() {
         )
         .await;
 }
+
+// Demonstrates the scenario where TProxy falls back to the secondary pool
+// after the primary pool returns a setup-connection error.
```
GitGab19:
```suggestion
// after the primary pool returns a `SetupConnection.Error`.
```

integration-tests/tests/translator_integration.rs
```diff
@@ -49,3 +65,139 @@ async fn translate_sv1_to_sv2_successfully() {
...
+
+    pool_translator_sniffer_2
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+        )
+        .await;
+}
+
+// Demonstrates the scenario where the primary pool returns an OpenMiningChannelError,
```
GitGab19:
```suggestion
// Demonstrates the scenario where the primary pool returns an `OpenMiningChannel.Error`,
```

miner-apps/translator/src/lib/sv2/upstream/upstream.rs
```diff
@@ -57,7 +56,10 @@ impl Upstream {
     /// to connect to each server multiple times before giving up.
     ///
     /// # Arguments
-    /// * `upstreams` - List of (address, public_key) pairs for upstream servers
+    /// * `upstreams` - A `(address, public_key, is_flagged)` tuples for the upstream server.
+    ///   `is_flagged` indicates whether the upstream has been marked as flagged. This value is
+    ///   always reliable at this point because the method is only invoked during
+    ///   `try_initialize_upstream`.
```
GitGab19:
```suggestion
    /// * `upstreams` - A single `(address, public_key, is_flagged)` tuple representing the upstream
    ///   candidate currently being attempted. The `is_flagged` is set once the upstream has
    ///   either been connected to successfully or marked as malicious. Because `new` is only called
    ///   from `try_initialize_upstream`, we can treat this flag as the definitive state for that
    ///   upstream.
```

plebhash (changes requested):
I'll set the Request changes here so we don't merge this by accident

PR #85, opened 2025-11-17
add workflow to build docker images and bootstrap docker setup

### LINKED ISSUE #50: provide release binaries and docker setups
as discussed during the dev call, we should provide binary releases for the sv2-apps alongside Docker images for each role.

The SRI team will need to create a Docker Hub account to publish the images.

lucasbalieiro:
to be merged after #59 ;

closes #50 ;

This PR adds:
- Github Action to build [multi-platform images](https://docs.docker.com/build/building/multi-platform/) for `linux/amd64`, `linux/arm64`.
- A `docker-compose` file with instructions on how to setup the environment


docker/README.md
```diff
@@ -0,0 +1,184 @@
...
+| mainnet  | `~/.bitcoin/node.sock`                     |
+| testnet4 | `~/.bitcoin/testnet4/node.sock`            |
+| signet   | `~/.bitcoin/signet/node.sock`              |
+| macOS    | Inside `~/Library/Application Support/...` |
+
+Two of the services (`pool_sv2` and `jd_client_sv2`) need access to your local Bitcoin Core `node.sock`.
+Because this path differs across operating systems, it is **not hardcoded**.
+Instead, you must provide it via an environment variable:
+
+### 1. Create a `.env` file (recommended)
```
GitGab19: Should we create a `.env.example`?

docker/docker-compose.yml
```diff
@@ -0,0 +1,77 @@
...
+      - ./config/proxy-config.toml:/app/proxy-config.toml:ro
+    networks:
+      sv2_apps:
+        ipv4_address: 172.28.0.14
+    depends_on:
+      - jd_client_sv2
+    ports:
+      - "34255:34255"
+    restart: unless-stopped
+    profiles:
```
GitGab19: Can we have a profile which can be used to run only the translator?

docker/docker-compose.yml
```diff
@@ -0,0 +1,77 @@
...
+    volumes:
+      - ./config/pool-config.toml:/app/pool-config.toml:ro
+      - "${BITCOIN_SOCKET_PATH?}:/app/node.sock"
+    ports:
+      - "34254:34254"
+    networks:
+      sv2_apps:
+        ipv4_address: 172.28.0.11
+    restart: unless-stopped
+    profiles:
```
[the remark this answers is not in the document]
GitGab19: Or maybe even better, we could have a profile called `no_jd` which runs Pool and tProxy together.

docker/docker-compose.yml
GitGab19: Or maybe even better, we could have a profile called `no_jd` which runs Pool and tProxy together.

docker/.dockerignore
```diff
@@ -0,0 +1 @@
+config/
```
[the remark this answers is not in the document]
lucasbalieiro: Removed. It was just leftovers from experiments

docker/config/proxy-config.toml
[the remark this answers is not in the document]
plebhash: what are the UX patterns we're aiming for?

the patterns I see on `docker-compose.yaml` is to simply have isolated images for each app

I was under the impression that we were aiming for things that would improve UX, mainly JDC+tProxy bundled together

having tProxy+Pool bundled together goes in a yet new direction, where we have apps from both pool and miner side bundled together

docker/docker-compose.yml
lucasbalieiro: added a profile `no_jd`

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
+###
+# ---- Builder Stage ----
+###
+FROM rust:1.85-slim-bookworm AS builder
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    capnproto libcapnp-dev && \
+    rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
Shourya742: TIL, thanks.

docker/config/jdc-config.toml.template
```diff
@@ -0,0 +1,79 @@
...
+# [[upstreams]]
+# authority_pubkey = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"
+# pool_address = "127.0.0.1:34254"
+# pool_port = "34254"
+# jds_address = "127.0.0.1:34264"
+# jds_port = "34264"
+
+# Bitcoin Core IPC config
+[template_provider_type.BitcoinCoreIpc]
+unix_socket_path = "/app/node.sock"
```
GitGab19: why don't we set this in the `docker_env` file, instead of mounting it to every container?

docker/config/translator-proxy-config.toml.template
```diff
@@ -0,0 +1,53 @@
...
+]
+
+# Difficulty params
+[downstream_difficulty_config]
+# hashes/s of the weakest miner that will be connecting (e.g.: 10 Th/s = 10_000_000_000_000.0)
+min_individual_miner_hashrate= ${TPROXY_MIN_INDIVIDUAL_MINER_HASHRATE}
+# target number of shares per minute the miner should be sending
+shares_per_minute = ${TPROXY_SHARES_PER_MINUTE}
+# disable variable difficulty adjustment when using with JDC (JDC handles vardiff)
+enable_vardiff = false
```
GitGab19: Let's put this configurable, with the "default" in the `docker_env.example` set to true.

The tProxy could be directly connected to the Pool, and in that case we set this to true.

docker/README.md
```diff
@@ -0,0 +1,177 @@
+# SV2 Docker Compose Setup
+
+This repository provides a ready-to-run Docker Compose setup for the full SV2 stack, including:
+
+* **Pool SV2 Service**
```
GitGab19:
```suggestion
* **Pool Service**
```

docker/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+Example:
+
+```yaml
+image: pool_sv2:v0.1.0   # pinned version
+# or
+image: pool_sv2:latest   # latest release
+```
+
+This applies to all images: `pool_sv2`, `jd_server`, `jd_client_sv2`, and `translator_sv2`.
```
GitGab19:
```suggestion
This applies to all images: `pool_sv2`, `jd_server_sv2`, `jd_client_sv2`, and `translator_sv2`.
```

docker/config/jdc-config.toml.template
```diff
@@ -0,0 +1,79 @@
...
+pool_port = "34254"
+jds_address = "172.28.0.12"
+jds_port = "34264"
+ 
+# [[upstreams]]
+# authority_pubkey = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"
+# pool_address = "127.0.0.1:34254"
+# pool_port = "34254"
+# jds_address = "127.0.0.1:34264"
+# jds_port = "34264"
```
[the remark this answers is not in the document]
lucasbalieiro: yess! thanks. added and updated the `docker_env.example`:
```diff
[[upstreams]]
authority_pubkey = "${JDC_UPSTREAM_AUTHORITY_PUBKEY}"
pool_address = "${JDC_POOL_ADDRESS}"
pool_port = "${JDC_POOL_PORT}"
jds_address = "${JDC_UPSTREAM_JDS_ADDRESS}"
jds_port = "${JDC_UPSTREAM_JDS_PORT}"
```

docker/README.md
lucasbalieiro: updated

docker/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+COPY pool-apps/ pool-apps/
+COPY miner-apps/ miner-apps/
+COPY bitcoin-core-sv2/ bitcoin-core-sv2/
+COPY stratum-apps/ stratum-apps/
+
+# build target app
+RUN echo "Building ${APP}" && \
+    case "$APP" in \
+        pool_sv2) cargo build --release --manifest-path pool-apps/pool/Cargo.toml --target-dir ./ ;; \
+        jd_server) cargo build --release --manifest-path pool-apps/jd-server/Cargo.toml --target-dir ./ ;; \
```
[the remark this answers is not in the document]
lucasbalieiro: I’d rather not change this right now because the Bash script I wrote depends on the current binary name.

`jd-server`’s Cargo.toml produces a binary called `jd_server`. I could rename it in Cargo.toml, but that risks breaking other workflows -- like the one that build the binaries -- and I haven’t tested the fallout yet.

If we change the name now, I’ll need to make the script more complex to handle both naming conventions (extra conditionals, branching logic, etc.). It’s doable, just messier.

If this isn’t a blocker, I’d prefer to leave the naming as-is and revisit it during the JDS refactor, where it’ll be easier to clean everything up consistently.

docker/README.md
lucasbalieiro: same as https://github.com/stratum-mining/sv2-apps/pull/85#discussion_r2566329282

The image name is being based on the binary name

docker/Dockerfile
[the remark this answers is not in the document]
GitGab19: It's definitely not a blocker, let's proceed with the name as it is 👍

Issue #29, opened 2025-09-01
JDC: Improve error orchestration system with prioritized error handling


Shourya742:
the entire application is currently brought down whenever an error occurs. This is not ideal, since not all errors are fatal.

We need a more robust error orchestration system that:

* Assigns priorities or severity levels to different error types.
* Lets the orchestration logic decide the appropriate action:

  * Shutdown the entire application (for critical errors).
  * Gracefully stop or restart only the affected subsystem (for localized issues).
  * Continue running with warnings (for non-lethal errors).

Goal:
Ensure the application can tolerate and recover from non-fatal errors without unnecessary downtime.

PR #86, opened 2025-11-18
New contextual error

Shourya742:
This PR introduces an initial design for contextual errors, enabling the status loop to react appropriately based on an error’s source and severity. The design is still evolving. For now, the changes apply only to pool; updates for JDC and tproxy will be added once the design is finalized.

closes: #29 #35

### DIFF
--- a/miner-apps/jd-client/config-examples/signet/jdc-config-bitcoin-core-ipc-local-infra-example.toml
+++ b/miner-apps/jd-client/config-examples/signet/jdc-config-bitcoin-core-ipc-local-infra-example.toml
@@ -1,5 +1,5 @@
 # SRI JDC config
-listening_address = "0.0.0.0:34265"
+listening_address = "127.0.0.1:34265"
 
 # Version support
 max_supported_version = 2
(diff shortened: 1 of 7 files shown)


pool-apps/pool/src/lib/error.rs
```diff
@@ -20,11 +20,75 @@ use stratum_apps::{
...
+    pub fn with_ctx(self, source: SourceCtx) -> Self {
+        Self {
+            source: Some(source),
+            ..self
+        }
+    }
+
+    pub fn with_severity(self, severity: Severity) -> Self {
+        Self { severity, ..self }
+    }
```
plebhash: why do we have `with_ctx` and `with_severity` to both `PoolError` and `PoolErrorKind`?

pool-apps/pool/src/lib/error.rs
```diff
@@ -94,9 +158,27 @@ pub enum PoolError {
...
+            error: self,
+            source: Some(source),
+            severity: Severity::Fatal,
+        }
+    }
+
+    pub fn with_severity(self, severity: Severity) -> PoolError {
+        PoolError {
+            error: self,
+            source: None,
```
plebhash: why are we hardcoding `None` here?

pool-apps/pool/src/lib/error.rs
```diff
@@ -20,11 +20,75 @@ use stratum_apps::{
...
         parsers_sv2::{Mining, ParserError},
     },
-    utils::types::ChannelId,
+    utils::types::{ChannelId, DownstreamId},
 };
 
 pub type PoolResult<T> = Result<T, PoolError>;
 
+#[derive(Debug)]
+pub enum SourceCtx {
```
plebhash: I don't see why we should abbreviate words here, we can write the entire word.

```suggestion
pub enum SourceContext {
```

also, since the corresponding method is called `with_ctx` (which I'd also rename to `with_context`), we should consider simply calling this `Context`

```suggestion
pub enum Context {
```

PR #87, opened 2025-11-19
Persistence Abstraction

average-gary:
Created abstract persistence interface for storing data, like hashrate, to a database.

Added a feature flag and Persistence Trait to stratum-apps and implemented FileBackend.  with the Pool role. A basic file writer implementation was added with the intent to expand to other persistence options. 

Additionally, `entities` are defined that allow for this framework to be more generally useful for any data that can be derived from the mining hot path of share and block validation. Initial implementation is `shares` but metrics or data of other sort could be added and handled appropriately.

Using the TaskManager that existed already, a Persistence implementation uses its own dedicated thread for operations. This allows for data to passed to the thread without blocking the mining hot path and offers flexibility in handling Persistence failures gracefully without interrupting mining operations.

### DIFF
--- a/stratum-apps/src/persistence/file.rs
+++ b/stratum-apps/src/persistence/file.rs
@@ -20,11 +20,13 @@ use tokio::io::AsyncWriteExt;
 /// # Example
 ///
 /// ```rust,no_run
-/// use std::path::PathBuf;
+/// use std::{path::PathBuf, sync::Arc};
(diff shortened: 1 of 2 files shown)


average-gary:
There's some bug here.
Running this now with a bit of hash and the TProxy falls over after about 90min of hashing. Have seen this a few times now. Running stratum-mining/sv2-app main branch today to see if I can isolate to our changes made here. Will dig into logs to suss out possible issues.

PR #96, opened 2025-11-25
get `stratum-core` from crates.io

### LINKED ISSUE #3: make apps use crates.io
Currently, our apps import the low-level crates via local paths.
As discussed during the dev call, we need to switch them to pull these crates from crates.io instead.

GitGab19:
Closes #3

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,11 +46,17 @@ jobs:
       - name: Test stratum-apps workspace
         run: cargo test --manifest-path=stratum-apps/Cargo.toml --all-features
 
-      - name: Test pool applications workspace
-        run: cargo test --manifest-path=pool-apps/Cargo.toml
+      - name: Test pool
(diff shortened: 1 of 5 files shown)


Shourya742 (commented):
Why did we remove jds and mining_device from workspace?

Shourya742:
> > Why did we remove jds and mining_device from workspace?
> 
> after the first commit of this PR (which is the main goal), the fact that their SRI dependencies point to crates.io (because they're not keeping up with the latest changes) was causing dependency hell to build:
> 
> * their respective workspaces
> * integration tests
> 
> JDS is about to be refactored anyways, where we can re-introduce it into the workspace
> 
> as for mining-device, perhaps we can come up with a better solution on the long term... but for now, this is the best we could find to unblock this PR (and the release as a whole)

Ah!!! tried building the first commit, saw the issue.

Shourya742 (approved):
ACK

PR #129, opened 2025-12-10
Monitoring module to expose channels' data through HTTP APIs

### LINKED ISSUE #105: Monitoring APIs to expose mining data
To further improve UX, we should provide a basic UI for our applications.

To fill this UI with some mining data and stats, we need some new APIs in all our apps to expose the necessary data.

GitGab19:
This PR introduces a new module called `monitoring` to the `stratum-apps` crate.

This new module defines the necessary structures which are used to represent every channels' data already retrievable through the `channels_sv2` getter methods.

It also defines  traits with methods to serve these data through the HTTP APIs, implemented by our applications.

It also adds support to Prometheus, by exporting metrics at `/metrics` endpoint.

Closes #105

### DIFF
--- a/miner-apps/translator/src/lib/sv1_monitoring.rs
+++ b/miner-apps/translator/src/lib/sv1_monitoring.rs
@@ -34,17 +34,16 @@ fn downstream_to_sv1_client_info(downstream: &Arc<Downstream>) -> Option<Sv1Clie
 
 impl Sv1ClientsMonitoring for Sv1Server {
     fn get_sv1_clients(&self) -> Vec<Sv1ClientInfo> {
-        let mut clients = Vec::new();
+        // Clone Arc references and release lock immediately to avoid contention
+        let downstream_refs: Vec<Arc<Downstream>> = self
+            .sv1_server_data
+            .safe_lock(|data| data.downstreams.values().cloned().collect())
+            .unwrap_or_default();
 
-        let _ = self.sv1_server_data.safe_lock(|data| {
-            for (_downstream_id, downstream) in data.downstreams.iter() {
-                if let Some(client_info) = downstream_to_sv1_client_info(downstream) {
-                    clients.push(client_info);
-                }
-            }
-        });
-
-        clients
+        downstream_refs
+            .iter()
+            .filter_map(downstream_to_sv1_client_info)
+            .collect()
     }
 
     fn get_sv1_client_by_id(&self, client_id: usize) -> Option<Sv1ClientInfo> {
--- a/pool-apps/pool/src/lib/monitoring.rs
+++ b/pool-apps/pool/src/lib/monitoring.rs
@@ -9,14 +9,15 @@ use stratum_apps::monitoring::client::{
 
 use crate::{channel_manager::ChannelManager, downstream::Downstream};
 
-/// Helper to convert a Downstream to ClientInfo
-fn downstream_to_client_info(client: &Downstream) -> ClientInfo {
-    let mut extended_channels = Vec::new();
-    let mut standard_channels = Vec::new();
-
+/// Helper to convert a Downstream to ClientInfo.
+/// Returns None if the lock cannot be acquired (graceful degradation for monitoring).
+fn downstream_to_client_info(client: &Downstream) -> Option<ClientInfo> {
     client
         .downstream_data
         .safe_lock(|dd| {
+            let mut extended_channels = Vec::new();
+            let mut standard_channels = Vec::new();
+
             for (_channel_id, extended_channel) in dd.extended_channels.iter() {
                 let channel_id = extended_channel.get_channel_id();
                 let target = extended_channel.get_target();
@@ -68,34 +69,38 @@ fn downstream_to_client_info(client: &Downstream) -> ClientInfo {
                     share_batch_size: share_accounting.get_share_batch_size(),
                 });
             }
-        })
-        .unwrap();
 
-    ClientInfo {
-        client_id: client.downstream_id,
-        extended_channels,
-        standard_channels,
-    }
+            ClientInfo {
+                client_id: client.downstream_id,
+                extended_channels,
+                standard_channels,
+            }
+        })
+        .ok()
 }
 
 impl ClientsMonitoring for ChannelManager {
     fn get_clients(&self) -> Vec<ClientInfo> {
-        let mut clients = Vec::new();
+        // Clone Downstream references and release lock immediately to avoid contention
+        // with template distribution and message handling
+        let downstream_refs: Vec<Downstream> = self
+            .channel_manager_data
+            .safe_lock(|data| data.downstream.values().cloned().collect())
+            .unwrap_or_default();
 
-        self.channel_manager_data
-            .safe_lock(|d| {
-                for (_client_id, client) in d.downstream.iter() {
-                    clients.push(downstream_to_client_info(client));
-                }
-            })
-            .unwrap();
-
-        clients
+        downstream_refs
+            .iter()
+            .filter_map(downstream_to_client_info)
+            .collect()
     }
 
     fn get_client_by_id(&self, client_id: usize) -> Option<ClientInfo> {
         self.channel_manager_data
-            .safe_lock(|d| d.downstream.get(&client_id).map(downstream_to_client_info))
-            .unwrap()
+            .safe_lock(|d| {
+                d.downstream
+                    .get(&client_id)
+                    .and_then(downstream_to_client_info)
+            })
+            .unwrap_or(None)
     }
 }
--- a/miner-apps/jd-client/src/lib/monitoring.rs
+++ b/miner-apps/jd-client/src/lib/monitoring.rs
@@ -15,11 +15,11 @@ use crate::{channel_manager::ChannelManager, downstream::Downstream};
 
 impl ServerMonitoring for ChannelManager {
     fn get_server(&self) -> ServerInfo {
-        let mut extended_channels = Vec::new();
-        let standard_channels = Vec::new(); // JDC only uses extended channels
-
         self.channel_manager_data
             .safe_lock(|d| {
+                let mut extended_channels = Vec::new();
+                let standard_channels = Vec::new(); // JDC only uses extended channels
+
                 if let Some(upstream_channel) = &d.upstream_channel {
                     let channel_id = upstream_channel.get_channel_id();
                     let target = upstream_channel.get_target();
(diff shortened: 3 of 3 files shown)


pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
```diff
@@ -40,8 +40,11 @@ required_extensions = [
     # 0x0002,
 ]
 
+# Monitoring API address (optional)
+monitoring_address = "0.0.0.0:9090"
+
 # Bitcoin Core IPC config
 [template_provider_type.BitcoinCoreIpc]
-unix_socket_path = "/path/to/node.sock"
+unix_socket_path = "/home/gitgab19/.bitcoin/node.sock"
```
lucasbalieiro: I think this was added by mistake

miner-apps/jd-client/config-examples/mainnet/jdc-config-bitcoin-core-ipc-hosted-infra-example.toml
```diff
@@ -56,6 +56,9 @@ required_extensions = [
     # 0x0002,
 ]
 
+# Monitoring HTTP server address for exposing channel data
+monitoring_address = "0.0.0.0:9091"
```
lucasbalieiro: we are missing this config in the signet also

GitGab19:
@lucasbalieiro @gimballock this PR is ready for review again.

I refactored the http server to use Axum, and I added support to Prometheus, by exporting metrics at the `/metrics/` endpoint.

pool-apps/pool/src/lib/monitoring.rs
```diff
@@ -0,0 +1,127 @@
...
+                }
+            })
+            .unwrap();
+
+        channels
+    }
+
+    fn get_client_summary(&self) -> ChannelsWithClientSummary {
+        let extended = self.get_client_extended_channels();
+        let standard = self.get_client_standard_channels();
```
gimballock: Both of these functions loop over downstream clients to count channels and hashrate. 
You could combine them to collect these stats in a single pass. 
Additionally, if you collect all the clients in one pass and then collect the stats in a second pass then you can avoid having nested locks as well.

GitGab19: I think that adding other helper functions called internally by these two can become something too convoluted for this purpose.

If you have a concrete suggestion to improve this, I'm open to it!

GitGab19: If the problem is the nested lock, we should probably remove it also from the other two method implementations (?)

GitGab19: I refactored the trait, and now it has only one method to implement, which is called `get_clients()`.

And then there's another method already implemented called `get_clients_summary()` which internally calls the `get_clients()`, to remove redundancy of the operations.

Let me know your thoughts about this new structure.

stratum-apps/src/monitoring/http_server.rs
```diff
@@ -0,0 +1,759 @@
...
+    ///
+    /// Automatically exposes:
+    /// - Swagger UI at `/swagger-ui`
+    /// - OpenAPI spec at `/api-docs/openapi.json`
+    /// - Prometheus metrics at `/metrics`
+    pub async fn run(
+        self,
+        shutdown_signal: impl Future<Output = ()> + Send + 'static,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+        info!("Starting monitoring server on {}", self.bind_address);
```
lucasbalieiro:
```suggestion
        info!("Starting monitoring server on http://{}", self.bind_address);
```

stratum-apps/src/monitoring/http_server.rs
```diff
@@ -0,0 +1,759 @@
...
+                    Json(serde_json::json!({
+                        "client_id": client_id,
+                        "offset": params.offset,
+                        "limit": params.effective_limit(),
+                        "total_extended": total_extended,
+                        "total_standard": total_standard,
+                        "extended_channels": extended,
+                        "standard_channels": standard,
+                    }))
+                    .into_response()
```
lucasbalieiro: I noticed that we have the definitions for the `ClientResponse` and `ServerResponse`, but we are only using it for the OpenAPI metadata, we should use it in the actual response as well. Leaving like this might make refactoring, if needed, risky. Because we will have to update the same response in multiple places.

Also, if we forgot to update one, we will have a mismatch between the OpenAPI metadata and the actual response

lucasbalieiro: Same apply for the other types like: `ErrorResponse`, `HealthResponse` etc

GitGab19: Done

stratum-apps/src/monitoring/http_server.rs
```diff
@@ -0,0 +1,759 @@
...
+        Pagination
+    ),
+    responses(
+        (status = 200, description = "Client details", body = ClientResponse),
+        (status = 404, description = "Client not found", body = ErrorResponse)
+    )
+)]
+async fn handle_client_by_id(
+    Path(client_id): Path<usize>,
+    Query(params): Query<Pagination>,
```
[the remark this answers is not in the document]
lucasbalieiro: > Do you mean the pagination?

yes, like you did for the path = "/api/v1/sv1/clients/{client_id}",

>Every client has unbounded Vec<ExtendedChannelInfo> and Vec<StandardChannelInfo>.

But I wasn't with this in mind when I made the comment. We should keep the pagination.

stratum-apps/src/monitoring/http_server.rs
lucasbalieiro: ACK

miner-apps/jd-client/config-examples/mainnet/jdc-config-bitcoin-core-ipc-local-infra-example.toml
```diff
@@ -56,6 +56,9 @@ required_extensions = [
     # 0x0002,
 ]
 
+# Monitoring HTTP server address for exposing channel data
+monitoring_address = "0.0.0.0:9091"
```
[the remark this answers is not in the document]
GitGab19: I made the Pool's config examples use `127.0.0.1` in https://github.com/stratum-mining/sv2-apps/pull/129/commits/3a030ff7b593f7edb386cfe9f66a57b2c8f569c6

GitGab19:
> would it make sense to add some Integration Tests to assert the correctness of these new monitoring APIs?
> 
> deploy some apps, do HTTP requests against their endpoints, then parse and assert against the data
> 
> I'd say this suggestion is not necessarily a blocker for this PR, since I'm doing some manual `curl` testing and the APIs seem sane (although I can't say I comprehensively tested everything)
> 
> but it might be a good idea for a follow-up, since we might unintentionally break these new APIs in the future, so having some CI feels like a good long-term strategy?

Opened an issue for it: https://github.com/stratum-mining/sv2-apps/issues/161

lucasbalieiro:
> @lucasbalieiro I adapted the docker configs in [c26abb5](https://github.com/stratum-mining/sv2-apps/commit/c26abb585f45ffd51c3043a7a53f47c215d99e99).
> 
> Please check everything is correct there.

From a config standpoint, this looks fine 👍

However, this PR will break the build image. `utoipa-swagger-ui` requires `curl`, and since we’re using `rust:1.85-slim-bookworm` as the base image, the build will fail without it.

Do you want to handle this change here, or should I take care of it in a follow-up PR?

We’ll need to update `docker/Dockerfile` to install `curl`:

```diff
diff --git a/docker/Dockerfile b/docker/Dockerfile
index 2afbc81b..5c180a79 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,7 @@
 FROM rust:1.85-slim-bookworm AS builder
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    capnproto libcapnp-dev && \
+    capnproto libcapnp-dev curl && \
     rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
```

Also slightly out of scope for this PR, but since we’re already touching Docker: the translator proxy entrypoint is outdated relative to the changes in [https://github.com/stratum-mining/sv2-apps/pull/123](https://github.com/stratum-mining/sv2-apps/pull/123) and should be updated to:

```yaml
   entrypoint: >
      sh -c '
        envsubst < /app/proxy-config.toml.template > /app/translator-config.toml &&
        exec /app/translator_sv2
      '
```

miner-apps/translator/src/lib/sv1_monitoring.rs
```diff
@@ -0,0 +1,60 @@
...
+
+        let _ = self.sv1_server_data.safe_lock(|data| {
+            for (_downstream_id, downstream) in data.downstreams.iter() {
+                if let Some(client_info) = downstream_to_sv1_client_info(downstream) {
+                    clients.push(client_info);
+                }
+            }
+        });
+
+        clients
```
gimballock: Here `sv1_server_data` is held for a time proportional to the number of miners. Several other tasks use that lock occasionally but the vardiff loop, in particular, uses that lock to process all of the `downstreams` we we do here, but it runs every 10s. So there is an increasing chance of contention as a function of the number of connected miners. 

To avoid/prevent lock contention here consider cloning the downstreams and releasing the lock, then process the clones knowing you aren't blocking vardiff.

```suggestion
    let downstream_refs: Vec<Arc<Downstream>> = self.sv1_server_data
        .safe_lock(|data| data.downstreams.values().cloned().collect())
        .unwrap_or_default();
    
    downstream_refs.iter()
        .filter_map(downstream_to_sv1_client_info)
        .collect()
```

pool-apps/pool/src/lib/monitoring.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        self.channel_manager_data
+            .safe_lock(|d| {
+                for (_client_id, client) in d.downstream.iter() {
+                    clients.push(downstream_to_client_info(client));
+                }
+            })
+            .unwrap();
+
+        clients
```
gimballock: The pool doesn't have a 10s vardiff loop like tproxy since this service is more event driven but `channel_manager_data` is still a used for template distribution and message handling. So reducing time spent in the critical section here should prevent lock contention. The change is the same as my comment from the tproxy: snapshot the inner data then process it after releasing the outer lock.
 
```suggestion
        let refs: Vec<Arc<Downstream>> = self.channel_manager_data
            .safe_lock(|data| data.downstream.values().cloned().collect())
            .unwrap_or_default();

        refs.iter()
            .map(|ds| downstream_to_client_info(ds))
            .collect()
```

pool-apps/pool/src/lib/monitoring.rs
```diff
@@ -0,0 +1,101 @@
...
+                    share_work_sum: share_accounting.get_share_work_sum(),
+                    last_share_sequence_number: share_accounting.get_last_share_sequence_number(),
+                    best_diff: share_accounting.get_best_diff(),
+                    last_batch_accepted: share_accounting.get_last_batch_accepted(),
+                    last_batch_work_sum: share_accounting.get_last_batch_work_sum(),
+                    share_batch_size: share_accounting.get_share_batch_size(),
+                });
+            }
+        })
+        .unwrap();
```
gimballock: Monitoring data can survive an occasional bad read. Here `safe_lock()` is used instead of `super_safe_lock()` because the latter panics on error while the former returns the error to be handled by the caller. If we just call `unwrap()` then we are replicating the behavior of super_safe_lock by allowing the process to panic. 
Instead delete this `unwrap()` or handle the error with a log message.

PR #131, opened 2025-12-11
Fix Pool sending `NewExtendedMiningJob` messages  to JDC

### LINKED ISSUE #55: Pool wrongly sends a `NewExtendedMiningJob` to JDC
I was just running all our apps, and I noticed JDC is receiving `NewExtendedMiningJob` messages from the Pool.

This should not happen and I remember we already fixed it in the past, but probably it got back during Pool refactoring.

Here a screenshot of JDC:

Image

GitGab19:
Closes #55

### DIFF
--- a/integration-tests/tests/pool_integration.rs
+++ b/integration-tests/tests/pool_integration.rs
@@ -3,7 +3,7 @@
 // `PoolSv2` is a module that implements the Pool role in the Stratum V2 protocol.
 use integration_tests_sv2::{interceptor::MessageDirection, template_provider::DifficultyLevel, *};
 use stratum_apps::stratum_core::{
-    common_messages_sv2::{Protocol, SetupConnection, *},
+    common_messages_sv2::{has_work_selection, Protocol, SetupConnection, *},
     mining_sv2::*,
     parsers_sv2::{AnyMessage, CommonMessages, Mining, TemplateDistribution},
     template_distribution_sv2::*,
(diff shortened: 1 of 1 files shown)


pool-apps/pool/src/lib/channel_manager/template_distribution_message_handler.rs
```diff
@@ -143,19 +152,27 @@ impl HandleTemplateDistributionMessagesFromServerAsync for ChannelManager {
...
                             for (channel_id, extended_channel) in data.extended_channels.iter_mut() {
                                 if let Err(e) = extended_channel.on_new_template(msg.clone().into_static(), coinbase_output.clone()) {
                                     tracing::error!("Error while adding template to standard channel: {channel_id:?} {e:?}");
                                     continue;
                                 }
+                                // don't send any NewExtendedMiningJob messages to Extended Channels
+                                // if the downstream requires custom work
+                                if downstream.requires_custom_work.load(Ordering::SeqCst) {
+                                    continue;
+                                }
```
Shourya742: Is there a reason we can't short-circuit template handling at the top of the function?
Something like:
```rust

if downstream.requires_custom_work.load(Ordering::SeqCst) {
    return messages;
}

match msg.future_template {
    ...
}

```
If the downstream wants custom work, do we still need to update channel state with the new template?

GitGab19: I think @Shourya742's suggestion is related to the `handle_new_template` and not the `handle_set_new_prev_hash`.

About the above proposal, I also thought about it, but then I looked at how the `handle_set_new_prev_hash` was dealing with that, and I did the same.

I guess your proposal is safe to do (we would need to have a `continue` instead of the `return messages;` you proposed), as long as we're sure that a JDC (or any generic client/proxy) cannot open standard channels with the Pool if it sent a `SetupConnection` with the `REQUIRES_WORK_SELECTION` flag set to true.

plebhash: I should correct my comment above, it's not accurate. Apologies for that.

the snippet @Shourya742 pasted above does allow room for the possibility of no matching future jobs, even with the changes introduced via https://github.com/stratum-mining/stratum/pull/2025/

pool-apps/pool/src/lib/channel_manager/template_distribution_message_handler.rs
```diff
@@ -151,17 +156,16 @@ impl HandleTemplateDistributionMessagesFromServerAsync for ChannelManager {
                                 let extended_job = extended_channel
                                     .get_active_job()
                                     .expect("extended job must exist");
-
                                 let extended_job_message = extended_job.get_job_message();
 
-                                messages.push((*downstream_id,Mining::NewExtendedMiningJob(extended_job_message.clone())).into());
+                                messages.push((*downstream_id, Mining::NewExtendedMiningJob(extended_job_message.clone())).into());
```
Shourya742: Very weird, we now have these formatting changes?

PR #132, opened 2025-12-12
Add contextual error

Shourya742:
I closed earlier PR by mistake and removed all the commits: https://github.com/stratum-mining/sv2-apps/pull/86.

### DIFF
--- a/stratum-apps/src/utils/types.rs
+++ b/stratum-apps/src/utils/types.rs
@@ -29,3 +29,9 @@ impl From<(DownstreamId, ChannelId)> for VardiffKey {
         }
     }
 }
+
+/// Marker traits used to define set of action
+/// the implementor can take
+pub trait CanDisconnect {}
+pub trait CanFallback {}
+pub trait CanShutdown {}


Shourya742:
@plebhash I’ve adapted the pool according to what we discussed. Let me know what you think. I’ll handle the JDC and Tproxy updates next in this PR.

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -297,7 +313,10 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                                     Ok(channel) => channel,
                                     Err(e) => {
                                         error!(?e, "Failed to create group channel");
-                                        return Err(JDCError::FailedToCreateGroupChannel(e));
+                                        return Err(JDCError::fatal(
+                                            e,
+                                            Source::Downstream(downstream_id),
```
GitGab19: Isn't this `Source::ChannelManager` ?

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -306,18 +325,14 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                                 coinbase_outputs.clone(),
                             ) {
                                 error!(?e, "Failed to apply template to group channel");
-                                return Err(JDCError::ChannelSv2(
-                                    ChannelSv2Error::GroupChannelServerSide(e),
-                                ));
+                                return Err(JDCError::fatal(e, Source::Downstream(downstream_id)));
```
GitGab19: Isn't this `Source::ChannelManager` ?

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -924,11 +934,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
         let messages = self.channel_manager_data.super_safe_lock(|channel_manager_data| {
             let Some(downstream) = channel_manager_data.downstream.get_mut(&downstream_id) else {
                 warn!("No downstream found for downstream_id={downstream_id}");
-                return Err(JDCError::DownstreamNotFound(downstream_id));
+                return Err(JDCError::fatal(JDCErrorKind::DownstreamNotFound(downstream_id), Source::Downstream(downstream_id)));
             };
             let Some(prev_hash) = channel_manager_data.last_new_prev_hash.as_ref() else {
                 warn!("No prev_hash available yet, ignoring share");
-                return Err(JDCError::LastNewPrevhashNotFound);
+                return Err(JDCError::fatal(JDCErrorKind::LastNewPrevhashNotFound, Source::Downstream(downstream_id)));
```
GitGab19: Isn't this `Source::ChannelManager` ?

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -940,7 +950,7 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                 };
 
                 let Some(vardiff) = channel_manager_data.vardiff.get_mut(&(downstream_id, channel_id).into()) else {
-                    return Err(JDCError::VardiffNotFound((downstream_id, channel_id).into()));
+                    return Err(JDCError::not_fatal(JDCErrorKind::VardiffNotFound((downstream_id, channel_id).into()), Source::ChannelManager));
```
GitGab19: Does this mean we're not gonna be able to apply vardiff on this `downstream_id`?

Shouldn't it be fatal?

How can we get this condition?

lucasbalieiro:
Started a testing session with the changes of this PR.

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -256,19 +264,28 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         channel_manager_data.last_future_template.clone()
                     else {
                         error!("Missing last_future_template, cannot open channel");
-                        return Err(JDCError::FutureTemplateNotPresent);
+                        return Err(JDCError::fatal(
+                            JDCErrorKind::FutureTemplateNotPresent,
+                            Impact::Downstream(downstream_id),
```
GitGab19: Isn't this `Impact::ChannelManager`?

If we don't get any future template, it means there's something wrong in the ChannelManager.

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -256,19 +264,28 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
+                        ));
                     };
 
                     let Some(last_new_prev_hash) = channel_manager_data.last_new_prev_hash.clone()
                     else {
                         error!("Missing last_new_prev_hash, cannot open channel");
-                        return Err(JDCError::LastNewPrevhashNotFound);
+                        return Err(JDCError::fatal(
+                            JDCErrorKind::LastNewPrevhashNotFound,
+                            Impact::Downstream(downstream_id),
```
GitGab19: Isn't this `Impact::ChannelManager`?

If we don't get any last_new_prev_hash, it means there's something wrong in the ChannelManager.

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -1309,9 +1320,9 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
         _tlv_fields: Option<&[Tlv]>,
     ) -> Result<(), Self::Error> {
         warn!("Received: {}", msg);
-        Err(Self::Error::UnexpectedMessage(
-            0,
-            MESSAGE_TYPE_SET_CUSTOM_MINING_JOB,
+        Err(JDCError::not_fatal(
+            JDCErrorKind::UnexpectedMessage(0, MESSAGE_TYPE_SET_CUSTOM_MINING_JOB),
+            Impact::ChannelManager,
```
GitGab19: Isn't this `Impact::Downstream`?

What do we do in the non_fatal case where the impact is the ChannelManager?

Shourya742: If we received an unexpected frame, we should just ignore it. Like how we did when pool used to send `ExtendedJobs` to JDC. These are not hard errors, just something we log and forget. In non-fatal, we do nothing just log the error.

miner-apps/jd-client/src/lib/channel_manager/jd_message_handler.rs
```diff
@@ -102,7 +101,9 @@ impl HandleJobDeclarationMessagesFromServerAsync for ChannelManager {
                 .tp_sender
                 .send(coinbase_output_constraints_message)
                 .await
-                .map_err(|_e| JDCError::ChannelErrorSender)?;
+                .map_err(|_e| {
+                    JDCError::fatal(JDCErrorKind::ChannelErrorSender, Impact::TemplateReceiver)
```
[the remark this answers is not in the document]
Shourya742: This error only gonna trigger, if the receiver side disconnects.

miner-apps/jd-client/src/lib/channel_manager/template_message_handler.rs
```diff
@@ -265,7 +271,10 @@ impl HandleTemplateDistributionMessagesFromServerAsync for ChannelManager {
         ) {
             return Ok(());
         }
-        Err(JDCError::TxDataError)
+        Err(JDCError::not_fatal(
```
GitGab19: Why is this not_fatal?

GitGab19 (commented):
I just finished to look at JDC changes, and I left many comments.

My hot take is that this PR is introducing more granularity in the error handling for sure, but I still think this entire error management is quite convoluted and hard to reason about.

For sure I would add docs to explain how this system entirely works for the sake of maintainability.

miner-apps/translator/src/lib/sv1/downstream/downstream.rs
```diff
@@ -335,7 +340,10 @@ impl Downstream {
                         .await
                         .map_err(|e| {
                             error!("Down: Failed to send queued message to downstream: {:?}", e);
-                            TproxyError::ChannelErrorSender
+                            TproxyError::fatal(
+                                TproxyErrorKind::ChannelErrorSender,
+                                Impact::Downstream(downstream_id.unwrap_or(0)),
```
GitGab19: Why this `unwrap_or(0)`?

miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
```diff
@@ -444,13 +450,13 @@ impl Sv1Server {
...
         first_target: Target,
-    ) -> Result<(), TproxyError> {
+    ) -> TproxyResult<()> {
         let (message, _tlv_fields) = self
             .sv1_server_channel_state
             .channel_manager_receiver
             .recv()
             .await
-            .map_err(TproxyError::ChannelErrorReceiver)?;
+            .map_err(|error| TproxyError::fatal(error, Impact::Sv1Server))?;
```
GitGab19: Isn't this Impact::ChannelManager?

miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
```diff
@@ -463,18 +469,24 @@ impl Sv1Server {
                     (downstream_id, v.downstreams.clone())
                 });
                 let Some(downstream_id) = downstream_id else {
-                    return Err(TproxyError::DownstreamNotFound(m.request_id));
+                    return Err(TproxyError::not_fatal(
+                        TproxyErrorKind::DownstreamNotFound(m.request_id),
+                        Impact::Sv1Server,
+                    ));
```
GitGab19: Shouldn't we send a `CloseChannel` to the upstream, and close the downstream connection in this case?

Or put it as fatal with Impact::Sv1Server?

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -197,17 +199,17 @@ impl ChannelManager {
...
-    pub async fn handle_upstream_frame(self: Arc<Self>) -> Result<(), TproxyError> {
+    pub async fn handle_upstream_frame(self: Arc<Self>) -> TproxyResult<()> {
         let mut channel_manager = self.get_channel_manager();
         let mut sv2_frame = self
             .channel_state
             .upstream_receiver
             .recv()
             .await
-            .map_err(TproxyError::ChannelErrorReceiver)?;
+            .map_err(|error| TproxyError::fatal(error, Impact::ChannelManager))?;
```
GitGab19: Isn't this Impact:Upstream?

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -197,17 +199,17 @@ impl ChannelManager {
...
         let header = sv2_frame.get_header().ok_or_else(|| {
             error!("SV2 frame missing header");
-            TproxyError::FramingSv2(framing_sv2::Error::MissingHeader)
+            TproxyError::fatal(framing_sv2::Error::MissingHeader, Impact::ChannelManager)
```
GitGab19: Isn't this Impact:Upstream?

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -252,13 +254,13 @@ impl ChannelManager {
...
     /// * `Err(TproxyError)` - Error processing the message
-    pub async fn handle_downstream_message(self: Arc<Self>) -> Result<(), TproxyError> {
+    pub async fn handle_downstream_message(self: Arc<Self>) -> TproxyResult<()> {
         let (message, tlv_fields) = self
             .channel_state
             .sv1_server_receiver
             .recv()
             .await
-            .map_err(TproxyError::ChannelErrorReceiver)?;
+            .map_err(|error| TproxyError::fatal(error, Impact::ChannelManager))?;
```
GitGab19: Isn't this Impact::Sv1Server?

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -270,7 +276,7 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
             .channel_state
             .status_sender
             .send(Status {
-                state: State::UpstreamShutdown(TproxyError::Shutdown),
+                state: State::UpstreamShutdown(TproxyErrorKind::Shutdown),
```
GitGab19: Why aren't we using this new contextual error system here?

miner-apps/jd-client/src/lib/error.rs
```diff
@@ -36,64 +39,108 @@ use stratum_apps::{
...
+    Fallback,
+    Shutdown,
 }
 
-impl JDCError {
-    pub fn fatal<E: Into<JDCErrorKind>>(error: E, source: Impact) -> Self {
+pub trait CanLog {}
+pub trait CanDisconnect {}
+pub trait CanFallback {}
+pub trait CanShutdown {}
```
plebhash: a few questions about these traits:
- why are declaring them inside `jd-client` crate and not `stratum-apps`? are we going to re-declare them over and over again them on every app?
- why are the traits empty? shouldn't they have some methods?
- what's the rationale for enforcing these traits at compile time? why not just review the code and make sure we don't write anything bad?

Shourya742: * Yeah, we can declare these inside `stratum-apps`. This commit was mainly about explaining the approach. I’ll move these marker traits into `stratum-apps`.

* These are marker traits, very similar to `Send` or `Sync`. They don’t define any methods, they just express a capability or property that the implementor follow.

* We already have different shutdown semantics across subsystems, some can fallback, some can disconnect, and some can trigger a full application shutdown. From the last dev sync, it was clear there’s still some confusion around subsystem lifecycles. With these traits, we can make those rules explicit in the type system. That way, even if someone writes incorrect code, the compiler will stop them. Relying purely on reviews makes it easy for subtle, unintended behavior to slip in, and this helps us avoid that class of errors altogether.

miner-apps/translator/src/lib/status.rs
```diff
@@ -79,40 +79,70 @@ pub struct Status {
...
+                reason: error.kind,
+            };
+
+            if let Err(e) = sender.send(Status { state }).await {
+                tracing::error!(
+                    "Failed to send downstream shutdown status from {:?}: {:?}",
+                    sender,
+                    e
+                );
+            }
```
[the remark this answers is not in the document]
Shourya742: Thats true, I have added the process abort.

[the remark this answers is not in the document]
plebhash: it's still missing in many places

I didn't mean only on this specific call, I meant for every case in which `sender.send()` fails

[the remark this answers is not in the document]
plebhash: please don't resolve this review point unless the issue has been properly addressed.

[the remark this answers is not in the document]
Shourya742: I have different opinion on this. Abort should only happen on shutdown. These channels are internal unbounded channels, used only during lifecycle fluctuations, so there’s no realistic risk of lagged items or backpressure. They are never overloaded, and even if an error does occur, retry should be the correct behavior. The only real reason they can error is if the receiver is dropped, but that shouldn’t happen either, since the lifetime of the main status loop is guaranteed to be >= the lifetime of any subsystem.

miner-apps/jd-client/src/lib/channel_manager/template_message_handler.rs
```diff
@@ -307,12 +307,14 @@ impl HandleTemplateDistributionMessagesFromServerAsync for ChannelManager {
         _ = self.allocate_tokens(1).await;
         let Some(token) = token else {
             error!("Token not found, template id: {}", msg.template_id);
-            return Err(JDCError::TokenNotFound);
+            return Err(JDCError::log(JDCErrorKind::TokenNotFound));
```
GitGab19: Without the token we cannot declare any job to JDS or set any job to Pool, shouldn't we try to fallback here?

miner-apps/jd-client/src/lib/channel_manager/template_message_handler.rs
```diff
@@ -307,12 +307,14 @@ impl HandleTemplateDistributionMessagesFromServerAsync for ChannelManager {
...
         };
 
         let Some(template_message) = template_message else {
             error!("Template not found, template id: {}", msg.template_id);
-            return Err(JDCError::TemplateNotFound(msg.template_id));
+            return Err(JDCError::log(JDCErrorKind::TemplateNotFound(
```
GitGab19: Why not shutdown here?

miner-apps/jd-client/src/lib/downstream/extensions_message_handler.rs
```diff
@@ -105,11 +118,17 @@ impl HandleExtensionsFromClientAsync for Downstream {
 
             let success = RequestExtensionsSuccess {
                 request_id: msg.request_id,
-                supported_extensions: Seq064K::new(supported.clone())
-                    .map_err(|_| JDCError::InvalidSupportedExtensionsSequence)?,
+                supported_extensions: Seq064K::new(supported.clone()).map_err(|_| {
+                    JDCError::disconnect(
+                        JDCErrorKind::InvalidSupportedExtensionsSequence,
+                        self.downstream_id,
+                    )
```
GitGab19: Same as https://github.com/stratum-mining/sv2-apps/pull/132/changes#r2665262875

miner-apps/jd-client/src/lib/upstream/mod.rs
```diff
@@ -360,9 +378,7 @@ impl Upstream {
                     .await
                     .map_err(|e| {
                         error!(error=?e, "Failed to send sv2 frame to upstream.");
-                        JDCError::CodecNoise(
-                            stratum_apps::stratum_core::noise_sv2::Error::ExpectedIncomingHandshakeMessage,
-                        )
+                        JDCError::fallback(noise_sv2::Error::ExpectedIncomingHandshakeMessage)
```
GitGab19: The error kind seems wrong

miner-apps/jd-client/src/lib/channel_manager/template_message_handler.rs
plebhash: in order to answer this, we need to evaluate some basic assumptions:

---

first, there's the fact that `data.allocate_tokens` is a `Option<AllocateMiningJobTokenSuccess<'static>>`, which is intialized as `None` during construction of `ChannelManagerData`, and then always populated as `Some` whenever new `AllocateMiningJobTokenSuccess`

however, this leaves room for a fundamental spec-compliance design-flaw, because it allows for token-reusage for multiple `DeclareMiningJob`

this design flaw is not really being introduced on this PR, but it's very relevant for this analysis

---

now let's look at the context here: JDC (under Full-Template mode) sends a `RequestTxData` to TP after receiving a `NewTemplate` message, and this is a `handle_request_tx_data_success`

then, there's the question:

assuming we re-set `data.allocate_tokens` to `None` after the token is used (which IMO should have been the correct implementation prior to this PR)

what should we do if we have a `NewTemplate`+`RequestTxData.Success` available for creation of a `DeclareMiningJob`, BUT no token available? that's the fundamental edge case at hand here

we have two options:
- skip this template, hoping that a `AllocateMiningJobToken.Success` will arrive before the next `NewTemplate`
- yield and poll `data.allocate_tokens` until it's `Some`, allowing us to finally use a fresh and unused token for the `DeclareMiningJob` we want to build on this context

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -258,19 +262,28 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
                     };
 
                     let Some(last_new_prev_hash) = channel_manager_data.last_new_prev_hash.clone()
                     else {
                         error!("Missing last_new_prev_hash, cannot open channel");
-                        return Err(JDCError::LastNewPrevhashNotFound);
+                        return Err(JDCError::disconnect(
+                            JDCErrorKind::LastNewPrevhashNotFound,
+                            downstream_id,
+                        ));
```
GitGab19: Shouldn't this be shutdown?

It seems an internal error.

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -310,18 +323,14 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
                             }
 
                             if let Err(e) =
                                 group_channel.on_set_new_prev_hash(last_new_prev_hash.clone())
                             {
                                 error!(?e, "Failed to apply prevhash to group channel");
-                                return Err(JDCError::ChannelSv2(
-                                    ChannelSv2Error::GroupChannelServerSide(e),
-                                ));
+                                return Err(JDCError::disconnect(e, downstream_id));
```
GitGab19: Shouldn't this be shutdown?

It seems an internal error.

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -928,11 +929,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
         let messages = self.channel_manager_data.super_safe_lock(|channel_manager_data| {
             let Some(downstream) = channel_manager_data.downstream.get_mut(&downstream_id) else {
                 warn!("No downstream found for downstream_id={downstream_id}");
-                return Err(JDCError::DownstreamNotFound(downstream_id));
+                return Err(JDCError::disconnect(JDCErrorKind::DownstreamNotFound(downstream_id), downstream_id));
             };
             let Some(prev_hash) = channel_manager_data.last_new_prev_hash.as_ref() else {
                 warn!("No prev_hash available yet, ignoring share");
-                return Err(JDCError::LastNewPrevhashNotFound);
+                return Err(JDCError::disconnect(JDCErrorKind::LastNewPrevhashNotFound, downstream_id));
```
GitGab19: Shouldn't this be shutdown?

It seems an internal error.

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -425,15 +407,15 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
                                                     .into()));
                                             }
                                             Err(e) => {
-                                                messages_results.push(Err(JDCError::ChannelSv2(
-                                                    ChannelSv2Error::StandardChannelServerSide(e),
+                                                messages_results.push(Err(JDCError::disconnect(
+                                                    e,
+                                                    *downstream_id,
```
GitGab19: Shouldn't this be shutdown?

It seems an internal error

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -462,15 +444,15 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
                                                     .into()));
                                             }
                                             Err(e) => {
-                                                messages_results.push(Err(JDCError::ChannelSv2(
-                                                    ChannelSv2Error::ExtendedChannelServerSide(e),
+                                                messages_results.push(Err(JDCError::disconnect(
+                                                    e,
+                                                    *downstream_id,
```
GitGab19: Shouldn't this be shutdown?

It seems an internal error

miner-apps/jd-client/src/lib/downstream/mod.rs
```diff
@@ -219,30 +221,38 @@ impl Downstream {
...
+    ) -> JDCResult<(), error::Downstream> {
         let (downstream_id, message, _tlv_fields) = match receiver.recv().await {
             Ok(msg) => msg,
             Err(e) => {
                 warn!(?e, "Broadcast receive failed");
-                return Ok(());
+                return Err(JDCError::disconnect(
+                    JDCErrorKind::BroadcastChannelErrorReceiver(e),
+                    self.downstream_id,
+                ));
```
GitGab19: Shouldn't this be shutdown?

It seems an internal error

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -131,11 +135,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
             }
 
             let Some(last_future_template) = channel_manager_data.last_future_template.clone() else {
-                return Err(PoolError::FutureTemplateNotPresent);
+                return Err(PoolError::disconnect(PoolErrorKind::FutureTemplateNotPresent, downstream_id));
```
GitGab19: Shouldn't this be shutdown?

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -131,11 +135,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
             };
 
             let Some(last_set_new_prev_hash_tdp) = channel_manager_data.last_new_prev_hash.clone() else {
-                return Err(PoolError::LastNewPrevhashNotFound);
+                return Err(PoolError::disconnect(PoolErrorKind::LastNewPrevhashNotFound, downstream_id));
```
GitGab19: Shouldn't this be shutdown?

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -153,17 +157,17 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         Ok(channel) => channel,
                         Err(e) => {
                             error!(?e, "Failed to create group channel");
-                            return Err(PoolError::FailedToCreateGroupChannel(e));
+                            return Err(PoolError::disconnect(e, downstream_id));
                         }
                     };
-                    group_channel.on_new_template(last_future_template.clone(), vec![pool_coinbase_output.clone()])?;
+                    group_channel.on_new_template(last_future_template.clone(), vec![pool_coinbase_output.clone()]).map_err(|error| PoolError::disconnect(error, downstream_id))?;
```
GitGab19: Shouldn't this be shutdown?

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -217,7 +221,7 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                 let template_id = last_future_template.template_id;
 
                 // create a future standard job based on the last future template
-                standard_channel.on_new_template(last_future_template, vec![pool_coinbase_output.clone()])?;
+                standard_channel.on_new_template(last_future_template, vec![pool_coinbase_output.clone()]).map_err(|error| PoolError::disconnect(error, downstream_id))?;
```
GitGab19: Shouldn't this be shutdown?

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
Shourya742: Same reasoning here. we’re fine with temporarily not accepting new connections until this is resolved, but there’s no need to shut down the whole application.

miner-apps/jd-client/src/lib/channel_manager/jd_message_handler.rs
```diff
@@ -175,17 +166,23 @@ impl HandleJobDeclarationMessagesFromServerAsync for ChannelManager {
...
-            return Err(JDCError::LastDeclareJobNotFound(msg.request_id));
+            return Err(JDCError::log(JDCErrorKind::LastDeclareJobNotFound(
+                msg.request_id,
+            )));
         };
 
         let Some(prevhash) = last_declare_job.prev_hash else {
             error!("Prevhash not found for request_id = {}", msg.request_id);
-            return Err(JDCError::LastNewPrevhashNotFound);
+            return Err(JDCError::log(JDCErrorKind::LastNewPrevhashNotFound));
```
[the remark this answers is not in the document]
Shourya742: Same reason: https://github.com/stratum-mining/sv2-apps/pull/132#discussion_r2676793445

miner-apps/jd-client/src/lib/channel_manager/jd_message_handler.rs
```diff
@@ -206,10 +203,11 @@ impl HandleJobDeclarationMessagesFromServerAsync for ChannelManager {
                 Some(custom_job)
             })
         else {
-            return Err(JDCError::FailedToCreateCustomJob);
+            return Err(JDCError::log(JDCErrorKind::FailedToCreateCustomJob));
         };
 
-        let custom_job = custom_job.map_err(|_e| JDCError::FailedToCreateCustomJob)?;
+        let custom_job =
+            custom_job.map_err(|_e| JDCError::log(JDCErrorKind::FailedToCreateCustomJob))?;
```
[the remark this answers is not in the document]
Shourya742: Same: https://github.com/stratum-mining/sv2-apps/pull/132#discussion_r2676812222

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -423,20 +427,20 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         let Some(last_set_new_prev_hash_tdp) =
                             channel_manager_data.last_new_prev_hash.clone()
                         else {
-                            return Err(PoolError::LastNewPrevhashNotFound);
+                            return Err(PoolError::disconnect(PoolErrorKind::LastNewPrevhashNotFound, downstream_id));
```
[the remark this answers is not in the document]
Shourya742: Same reason: https://github.com/stratum-mining/sv2-apps/pull/132#discussion_r2676788588

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
Shourya742: If the upstream sends a new extranonce_prefix and that violates the current channel invariants, shouldn’t we just close the channel or disconnect the downstream? At that point, the channel is no longer valid since it no longer satisfies the downstream’s extranonce prefix constraints. Longer term, it might be more appropriate to introduce explicit close channel semantics into the contextual error system this could be very useful.

PR #159, opened 2025-12-30
Role Profiler

Shourya742:
This PR adds basic profiling support to our long-running runtime servers so we can actually see what they’re doing while they’re running. We’re using `hotpath` to profile both latency and memory usage of roles. It gives a pretty nice UII where you can inspect things like flamegraphs, memory consumption, thread activity, async task polling, and even channel state (note: `async_channel` and broadcast channels aren’t supported yet). This is useful for roles, it helps spot which functions are taking most of the time, where performance regresses over time, and how memory behaves under load. 

Profiling is behind feature flags, so there’s no overhead unless you turn it on:

* `hotpath` → time / CPU profiling
* `hotpath-alloc` → memory profiling (This only works in single threaded context)

To run:

```bash
cargo install hotpath --features='tui'
hotpath console
```

Once the role is running with the feature enabled, `hotpath console` will show a live view of what’s going on.

UI looks something like this:

Screenshot from 2025-12-30 15-07-51

### DIFF
--- a/PROFILING.md
+++ b/PROFILING.md
@@ -1,0 +1,71 @@
+# Performance Profiling
+
+> **⚠️ For Development Only**: Profiling is intended for development and debugging. Do not enable profiling features in production deployments.
+
+This guide explains how to profile Sv2 applications using [hotpath-rs](https://hotpath.rs/) to identify bottlenecks and optimize performance.
+
+## Overview
+
+Profiling is **zero-cost when disabled** - all instrumentation is gated behind feature flags and has no overhead unless explicitly enabled. For production deployments, always build without profiling features.
+
+## Setup
+
+### Building with Profiling
+
+```bash
+# Basic profiling (multi-threaded runtime)
+cargo build --release --features hotpath
+
+# Profiling with allocation tracking (single-threaded runtime)
+cargo build --release --features hotpath-alloc
+```
+
+**Note**: The `hotpath-alloc` feature includes base profiling plus memory allocation tracking. It uses a single-threaded tokio runtime, which may affect performance characteristics.
+
+## Profiling Modes
+
+### Option 1: Static Report (Simple)
+
+Prints a profiling summary when the application shuts down.
+
+```bash
+# Example with Pool
+cd pool-apps/pool
+cargo run --release --features hotpath -- -c config-examples/testnet4/pool-config-hosted-sv2-tp-example.toml
+# ... run workload ...
+# Press Ctrl+C to stop and view report
+```
+
+### Option 2: Live TUI Dashboard (Advanced)
+
+Real-time monitoring with an interactive terminal dashboard.
+
+**1. Install the hotpath CLI (once)**
+```bash
+cargo install hotpath --features='tui' --locked
+```
+
+**2. Start the dashboard**
+```bash
+hotpath console
+```
+
+**3. In another terminal, run your application**
+```bash
+# Example with Pool
+cd pool-apps/pool
+cargo run --release --features hotpath -- -c config-examples/testnet4/pool-config-hosted-sv2-tp-example.toml
+```
+
+The TUI will show real-time performance metrics as your application runs.
+
+## Interpreting Results
+
+Profiling data includes:
+- **Call counts**: How many times each function was called
+- **Total time**: Cumulative execution time
+- **Average time**: Mean execution time per call
+- **Percentage**: Time spent relative to total runtime
+- **Percentiles**: p95, p99 timing statistics
+- **Memory allocations**: Bytes allocated and allocation counts (with `--features hotpath-alloc`)
+


miner-apps/translator/src/main.rs
```diff
@@ -4,12 +4,24 @@ pub use translator_sv2::{config, error, status, sv1, sv2, TranslatorSv2};
...
+async fn main() {
+    _ = inner_main().await;
+}
+
+#[cfg(not(feature = "hotpath-alloc"))]
+#[tokio::main]
+async fn main() {
+    _ = inner_main().await;
+}
+
```
plebhash:
```suggestion
#[cfg(not(feature = "hotpath-alloc"))]
#[tokio::main]
async fn main() {
    inner_main().await;
}

```

miner-apps/jd-client/src/main.rs
```diff
@@ -5,8 +5,20 @@ use crate::args::process_cli_args;
 
 mod args;
 
+#[cfg(feature = "hotpath-alloc")]
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    _ = inner_main().await;
```
plebhash:
```suggestion
    inner_main().await;
```

pool-apps/pool/src/lib/mod.rs
```diff
@@ -60,17 +60,15 @@ impl PoolSv2 {
 
         let task_manager = Arc::new(TaskManager::new());
 
-        let (status_sender, status_receiver) = async_channel::unbounded::<Status>();
+        let (status_sender, status_receiver) = unbounded();
```
plebhash: why are we changing this?

pool-apps/pool/src/lib/mod.rs
```diff
@@ -60,17 +60,15 @@ impl PoolSv2 {
...
             broadcast::channel(10);
         let (downstream_to_channel_manager_sender, downstream_to_channel_manager_receiver) =
             unbounded();
 
-        let (channel_manager_to_tp_sender, channel_manager_to_tp_receiver) =
-            unbounded::<TemplateDistribution<'static>>();
-        let (tp_to_channel_manager_sender, tp_to_channel_manager_receiver) =
-            unbounded::<TemplateDistribution<'static>>();
+        let (channel_manager_to_tp_sender, channel_manager_to_tp_receiver) = unbounded();
+        let (tp_to_channel_manager_sender, tp_to_channel_manager_receiver) = unbounded();
```
plebhash: why are we changing this?

miner-apps/jd-client/Cargo.toml
```diff
@@ -24,3 +24,9 @@ ext-config = { version = "0.14.0", features = ["toml"], package = "config" }
 tracing = { version = "0.1" }
 clap = { version = "4.5.39", features = ["derive"] }
 bitcoin_core_sv2 = { path = "../../bitcoin-core-sv2" }
+hotpath = "0.9"
+
+[features]
+hotpath = ["hotpath/hotpath"]
+hotpath-alloc = ["hotpath/hotpath-alloc"]
```
plebhash: I tried running with `--features hotpath-alloc` and nothing happened. After fixing this, it worked.

```suggestion
hotpath-alloc = ["hotpath", "hotpath/hotpath-alloc"]
```

plebhash:
if we're going to have this `hotpath` instrumentation on our code for mid/long term, we should have some kind of docs.

I asked Claude Sonnet 4.5 to help me write a `PROFILING.md` to be added to the root of the repo, documenting this.

I would suggest cherry-picking https://github.com/plebhash/sv2-apps/commit/43dfc78e1cdc79bccc3b2b2acfff722d5eb95472 here

pool-apps/pool/src/lib/mod.rs
Shourya742: was adding the hotpath channel wrapper to profile the channels, and noticed that we were being explicit about the channel type even though the compiler infer it. Since it’s not required, I removed the explicit type.

pool-apps/pool/src/lib/mod.rs
Shourya742: Same reason: https://github.com/stratum-mining/sv2-apps/pull/159#discussion_r2667808011

PROFILING.md
```diff
@@ -0,0 +1,71 @@
...
+# Press Ctrl+C to stop and view report
+```
+
+### Option 2: Live TUI Dashboard (Advanced)
+
+Real-time monitoring with an interactive terminal dashboard.
+
+**1. Install the hotpath CLI (once)**
+```bash
+cargo install hotpath --features='tui'
```
GitGab19: This is what I got after running it:
```
cargo install hotpath --features='tui'
    Updating crates.io index
  Installing hotpath v0.9.2
    Updating crates.io index
     Locking 199 packages to latest compatible versions
      Adding darling v0.23.0 (requires Rust 1.88.0)
      Adding darling_core v0.23.0 (requires Rust 1.88.0)
      Adding darling_macro v0.23.0 (requires Rust 1.88.0)
      Adding instability v0.3.11 (requires Rust 1.88)
      Adding ratatui v0.29.0 (available: v0.30.0, requires Rust 1.86.0)
      Adding unicode-width v0.2.0 (available: v0.2.2)
error: failed to compile `hotpath v0.9.2`, intermediate artifacts can be found at `/tmp/cargo-installbVqLQL`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  rustc 1.85.0 is not supported by the following packages:
    darling@0.23.0 requires rustc 1.88.0
    darling_core@0.23.0 requires rustc 1.88.0
    darling_macro@0.23.0 requires rustc 1.88.0
    instability@0.3.11 requires rustc 1.88
    instability@0.3.11 requires rustc 1.88
    instability@0.3.11 requires rustc 1.88
  Try re-running `cargo install` with `--locked`
```

After running it with `--locked` it worked.


```suggestion
cargo install hotpath --features='tui' --locked
```

PR #160, opened 2026-01-01
Adapt to new sv1 handler

Shourya742:
Part of #37 

This PR adapt translator with new context changes to sv1 traits. 

companion to https://github.com/stratum-mining/stratum/pull/2047

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -179,7 +179,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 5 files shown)


plebhash:
from the discussions on Discord over the past few days, I was under the impression that this would help mitigate the deadlock issues that were recently identified

but on this PR, I see that we're only adapting tProxy to the API changes introduced by https://github.com/stratum-mining/stratum/pull/2047

are we planning to fix the deadlock on a follow up?

plebhash:
@Shourya742 I think you need to update `miner-apps/Cargo.lock` to unblock CI here

plebhash:
> @Shourya742 I think you need to update `miner-apps/Cargo.lock` to unblock CI here

force-pushed this so I can merge and rebase #156

PR #171, opened 2026-01-08
Add hotpath patch

Shourya742:


GitGab19 (commented):
Can you modify the `msrv.yaml` to build the apps with `--all-features`?

In that way we won't experience similar issues anymore.

Shourya742:
> Can you modify the `msrv.yaml` to build the apps with `--all-features`?
> 
> In that way we won't experience similar issues anymore.

Added here:  53227e7

Issue #92, opened 2025-11-21
Pool blindly accepts `SetupConnection`


plebhash:
most fields of `SetupConnection` exist with a purpose, and they must be checked against

this routine blindly replies with `SetupConnection.Success`, regardless of whether `SetupConnection` has nonsense values (e.g.: `protocol` is not Mining)

https://github.com/stratum-mining/sv2-apps/blob/1eb501aac68aa238fff8a2541dd09bf9e408a2cd/pool-apps/pool/src/lib/downstream/common_message_handler.rs#L18-L44

PR #173, opened 2026-01-09
Reject non mining protocol SetupConnection message in Pool

Shourya742:
This PR enforces rejection of all downstream connections that do not use the mining protocol. It removes hardcoded address information from the SetupConnection message throughout the codebase and adds integration tests to cover the new behavior.

Closes: #92

### DIFF
--- a/pool-apps/pool/src/lib/downstream/common_message_handler.rs
+++ b/pool-apps/pool/src/lib/downstream/common_message_handler.rs
@@ -31,7 +31,7 @@ impl HandleCommonMessagesFromClientAsync for Downstream {
 
     async fn handle_setup_connection(
         &mut self,
-        _client_id: Option<usize>,
+        client_id: Option<usize>,
         msg: SetupConnection<'_>,
         _tlv_fields: Option<&[Tlv]>,
     ) -> Result<(), Self::Error> {
@@ -40,8 +40,10 @@ impl HandleCommonMessagesFromClientAsync for Downstream {
             msg.min_version, msg.flags
         );
 
+        let downstream_id = client_id.expect("downstream id should be present");
+
         if msg.protocol != Protocol::MiningProtocol {
-            info!("Rejecting connection: SetupConnection asking for other protocols than mining protocol.");
+            info!("Rejecting connection from {downstream_id}: SetupConnection asking for other protocols than mining protocol.");
             let response = SetupConnectionError {
                 flags: 0,
                 error_code: "unsupported-protocol"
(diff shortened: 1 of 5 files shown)


integration-tests/tests/pool_integration.rs
```diff
@@ -339,3 +343,58 @@ async fn pool_does_not_send_jobs_to_jdc() {
         "Pool should NOT send non-future NewExtendedMiningJob messages to JDC"
     );
 }
+
+// The test runs pool and translator, with translator sending a SetupConnection message
+// with a wrong protocol, this test asserts whether pool sends SetupConnection error or
+// not to such downstream.
+#[tokio::test]
+async fn pool_reject_setup_connection_with_non_mining_protocol() {
```
plebhash: as an alternative to tProxy + `ReplaceMessage`, this could have also been achieved with `MockDownstream`

nothing wrong with this approach, I'm not suggesting a change... just pointing out this alternative, which would have been my choice in case I was writing this test

my rationale there would have been centered around the fact that I just want to assert Pool behavior, without bringing external and unrelated factors (e.g.: tProxy sane behavior) into the mix

pool-apps/pool/src/lib/downstream/common_message_handler.rs
```diff
@@ -36,6 +37,21 @@ impl HandleCommonMessagesFromClientAsync for Downstream {
...
+            info!("Rejecting connection: SetupConnection asking for other protocols than mining protocol.");
+            let response = SetupConnectionError {
+                flags: 0,
+                error_code: "unsupported-protocol"
+                    .to_string()
+                    .try_into()
+                    .expect("error code must be valid string"),
+            };
+            let frame: Sv2Frame = AnyMessage::Common(response.into_static().into()).try_into()?;
+            _ = self.downstream_channel.downstream_sender.send(frame).await;
```
plebhash: why ignore this `Result`?

integration-tests/tests/pool_integration.rs
```diff
@@ -339,3 +343,58 @@ async fn pool_does_not_send_jobs_to_jdc() {
...
+        "0",
+        pool_addr,
+        false,
+        vec![setup_connection_replace.into()],
+        None,
+    );
+    let (_tproxy, _) =
+        start_sv2_translator(&[pool_translator_sniffer_addr], false, vec![], vec![], None).await;
+
+    std::thread::sleep(std::time::Duration::from_secs(1));
```
lucasbalieiro: shouldn't we be using the tokio sleep?

```suggestion
    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```

pool-apps/pool/src/lib/downstream/common_message_handler.rs
```diff
@@ -36,6 +37,21 @@ impl HandleCommonMessagesFromClientAsync for Downstream {
...
+
+            return Err(PoolError::Shutdown);
```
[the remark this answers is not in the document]
Shourya742: Any error occurring in the downstream context such as in the downstream common message handler, will result in the downstream being disconnected.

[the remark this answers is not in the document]
Shourya742: Agree, shutdown variant is not more with us in contextual PR.

integration-tests/tests/pool_integration.rs
Shourya742: Added the commit. Thanks

lucasbalieiro:
after running with the latests patches the test `pool_reject_setup_connection_with_non_mining_protocol` now finishes as expected.

tested @ [8129fda](https://github.com/stratum-mining/sv2-apps/pull/173/commits/8129fda6049e989ce238a04aa75acff6003668d9)

PR #184, opened 2026-01-15
add shell-style path expansion for config paths

### LINKED ISSUE #101: improve the config file parsers
while working with the Docker setup, I noticed a few opportunities to improve how the config files handle path-related values in the TOML parser.

It would be helpful if the parser could support:

#### 1. Home directory expansion

Allowing `~` in paths, e.g.:

```
unix_socket_path = "~/Library/Application Support/Bitcoin/signet/node.sock"
```

#### 2. Environment variable expansion

Supporting variables like `$HOME`, e.g.:

```
unix_socket_path = "$HOME/Library/Application Support/Bitcoin/signet/node.sock"
```

Beyond file paths, environment variable expansion could also simplify handling of sensitive values. For example:

```
authority_secret_key = "$AUTHORITY_SECRET_KEY"
```

This would let users provide secrets through environment variables instead of storing them directly in config files.

lucasbalieiro:
close #101 

Add support for `~` and environment variable expansion when deserializing paths from the TOML config.

Applies to both required and optional path fields using a shared helper, implemented via `shellexpand`. No config format changes or behavior outside config parsing.

### DIFF
--- a/stratum-apps/src/config_helpers/toml.rs
+++ b/stratum-apps/src/config_helpers/toml.rs
@@ -60,3 +60,98 @@ where
     })
     .transpose()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ext_config::{Config, File, FileFormat};
+    use serde::Deserialize;
+    use std::env;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Deserialize)]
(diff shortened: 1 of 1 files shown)


Shourya742 (approved):
ACK

PR #191, opened 2026-01-19
Add implicit SetupConnection handling to MockDownstream and MockUpstream

### LINKED ISSUE #148: `integration_tests_sv2::{MockDownstream, MockUpstream}` should handle `SetupConnection` implicitly
it's already implied that whenever we're using `MockDownstream`, it's going to send a `SetupConnection` as the first thing they do

the same goes for `MockDownstream`, which should reply with a `SetupConnection.Success` if it receives a `SetupConnection` with the correct parameters

therefore, the user shouldn't need to explicitly create a `SetupConnection`/`SetupConnection.Success` message

they should give the meaningful parameters (`protocol` + `flags`) while calling `new` or `start`, and already rest assured that the `SetupConnection` message will be properly handled

note: while it might be tempting to allow the user to insert **every single** parameter to the `SetupConnection`, most of them don't really matter for the coverage we're aiming at (e.g.: `min_version`, `max_version`, `endpoint_host`, `vendor`, etc)

so I'd focus strictly on `protocol` + `flags` for the sake of simplicity and a smoother dev experience

xyephy:
MockDownstream now automatically sends SetupConnection on start(). MockUpstream now automatically responds with SetupConnectionSuccess when it receives a SetupConnection with matching protocol.

Both now require protocol and flags parameters in new().

Closes #148

### DIFF
--- a/integration-tests/lib/mock_roles.rs
+++ b/integration-tests/lib/mock_roles.rs
@@ -1,29 +1,64 @@
 use crate::utils::{create_downstream, create_upstream, message_from_frame, wait_for_client};
 use async_channel::Sender;
-use std::net::SocketAddr;
-use stratum_apps::{
-    stratum_core::{
-        codec_sv2::StandardEitherFrame,
-        parsers_sv2::{AnyMessage, IsSv2Message},
+use std::{convert::TryInto, net::SocketAddr};
+use stratum_apps::stratum_core::{
+    codec_sv2::StandardEitherFrame,
+    common_messages_sv2::{
+        Protocol, SetupConnection, SetupConnectionError, SetupConnectionSuccess,
+        MESSAGE_TYPE_SETUP_CONNECTION,
     },
-    utils::types::Sv2Frame,
+    parsers_sv2::{AnyMessage, CommonMessages, IsSv2Message},
 };
+use stratum_apps::utils::types::Sv2Frame;
 use tokio::net::TcpStream;
 use tracing::info;
 
+pub enum WithSetup {
+    Yes(SetupConnection<'static>),
+    No,
+}
+
+impl WithSetup {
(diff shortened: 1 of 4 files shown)


xyephy:
> @xyephy are the integration tests all passing to you, locally?

yes, all tests pass locally

jbesraa (commented):
LGTM, the only nit I have is that there is some repetitive code in `MockUpstream` and generally it looks a bit too complex now. 

1) Match arms have shared code between them, so either move it before or after the match.
2) The `WithSetup::Yes` is executing `downstream_receiver.recv()` twice.

I would do something like the following:
```rust
   pub async fn start(self) -> Sender<AnyMessage<'static>> {
        let listening_address = self.listening_address;

        let (proxy_sender, proxy_receiver) = async_channel::unbounded::<AnyMessage<'static>>();

        let (downstream_receiver, downstream_sender) =
            create_downstream(wait_for_client(listening_address).await)
            .await
            .expect("Failed to connect to downstream");
       tokio::spawn(async move {
           while let Ok(mut frame) = downstream_receiver.recv().await {
               let (msg_type, msg) = message_from_frame(&mut frame);
               if msg_type == MSG_TYPE_SETUP_CONNECTION && self.require_setup() {
                   self.do_setup(...)
               }
               info!(
                   "MockUpstream: received message from downstream: {} {}",
                   msg_type, msg
               );
           }
       });

       while let Ok(message) = proxy_receiver.recv().await {
           let message_type = message.message_type();
           let frame = StandardEitherFrame::<AnyMessage<'_>>::Sv2(
               Sv2Frame::from_message(message, message_type, 0, false)
               .expect("Failed to create frame from message"),
           );
           if downstream_sender.send(frame).await.is_err() {
               break;
           }
       }

        proxy_sender
    }
}
```

PR #197, opened 2026-01-21
fix: validate username and truncate worker name at authorize time

average-gary:
The UserIdentity TLV field has a protocol-specified maximum of 32 bytes. When downstream miners send mining.authorize with long worker names, the translator would panic on unwrap().

This fix:
- Rejects mining.authorize if the username portion (before the dot) exceeds 32 bytes, as this indicates an invalid configuration
- Truncates only the worker name portion if the total length exceeds 32 bytes, preserving the username
- Adds defensive error handling in sv1_server.rs as a fallback

### DIFF
--- a/miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
@@ -3,10 +3,9 @@ use stratum_apps::stratum_core::sv1_api::{
     utils::{Extranonce, HexU32Be},
     IsServer,
 };
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::{
-    error,
     sv1::{downstream::SubmitShareWithChannelId, Sv1Server},
     utils::validate_sv1_share,
 };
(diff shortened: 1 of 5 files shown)


miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
```diff
@@ -200,11 +226,26 @@ impl IsServer<'static> for Sv1Server {
...
+        // Truncate user_identity to MAX_USER_IDENTITY_LENGTH if needed
+        // Username was already validated in handle_authorize, so only worker name gets truncated
+        let user_identity = if name.len() > MAX_USER_IDENTITY_LENGTH {
+            warn!(
+                "Worker name in '{}' exceeds {} byte limit, truncating to '{}'",
+                name,
+                MAX_USER_IDENTITY_LENGTH,
+                &name[..MAX_USER_IDENTITY_LENGTH]
+            );
+            name[..MAX_USER_IDENTITY_LENGTH].to_string()
```
gimballock: Have you tested with non-ASCII characters in worker names? The byte-slicing could panic.

GitGab19:
> (included extension negotiation)

I'm not sure about this assertion, are you?

In general, if there's an easy way to get the negotiated extensions from there, I guess we can eventually refuse the `mining.authorize` in case the username exceeds the max 32 bytes,only  when the extension is required and negotiated.

average-gary:
> I'm not sure about this assertion, are you?

`initialize_upstream()` of `TranlatorSv2` -> `try_initialize_upstream()` -> creates `new()` `Upstream` and then `start()` -> `setup_connection()` which has extension negotiation messaging. 
Once we're done awaiting `try_initialize_upstream()`  for `TranslatorSv2`, we then `sv1_server_instance.start()`.

Ah, I see, it's just a request extension message we're sending and then we handle messages.

Is there a reason we don't fully negotiate extensions in connection setup?

GitGab19:
> Is there a reason we don't fully negotiate extensions in connection setup?

I don't remember exactly why I implemented it in that way tbh.

PR #212, opened 2026-01-24
add mimalloc

bit-aloo:
This PR switches the binaries to use mimalloc instead of the default glibc allocator, which can exhibit higher fragmentation and contention under our workload.

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -1669,6 +1669,7 @@ dependencies = [
  "config",
  "hex",
  "hotpath",
+ "mimalloc",
  "serde",
  "stratum-apps",
  "tokio",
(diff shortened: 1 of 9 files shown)


GitGab19:
Should we close this @Shourya742 ? 

Or at least putting it in draft?

bit-aloo:
I don't remember now, why we stalled this PR?

GitGab19:
I don't remember exactly now, but it was not something we ever planned?

Maybe it makes sense to run again the analysis after the improvements on hotpath described [here](https://github.com/stratum-mining/sv2-apps/issues/529)?

bit-aloo:
That might be the case, once the hotpath thingy is in. I will give the numbers another shot.

PR #226, opened 2026-01-31
Add clean_queue method to Sniffer

### LINKED ISSUE #155: `integration_test_sv2::Sniffer` needs a `clean_queue` method
sometimes, we need to make sure `Sniffer` is "flushed" in the flow of an integration test

something like:

```rust
sniffer.clean_queue(MessageDirection::ToDownstream);
```

or

```rust
sniffer.clean_queue(MessageDirection::ToUpstream);
```

with a similar behavior to what happens when we call `wait_for_message_type_and_clean_queue` (but without the `wait_for_message_type` part)

xyephy:
Adds a clean_queue(MessageDirection) method that clears all messages from the specified direction's queue. This allows tests to flush the sniffer without waiting for a specific message type.

Closes #155

### DIFF
--- a/integration-tests/lib/message_aggregator.rs
+++ b/integration-tests/lib/message_aggregator.rs
@@ -50,6 +50,13 @@ impl MessagesAggregator {
             .unwrap()
     }
 
+    /// Clears all messages from the queue.
+    pub fn clear(&self) {
+        self.messages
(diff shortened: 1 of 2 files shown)


plebhash:
LGTM

is it in draft for some specific reason?

Issue #185, opened 2026-01-16
tProxy: introduce global `AGGREGATION_MODE` variable


plebhash:
to follow the same pattern as JDC, instead of having it locally on `ChannelManagerData` only

plebhash:
while working on #156 @Shourya742 drafted a potential solution

ee827184912260e9af78e1ae71c401768114ff9a

PR #229, opened 2026-02-02
Add tproxy mode

Shourya742:
Closes: #185

### DIFF
--- a/miner-apps/translator/src/lib/mod.rs
+++ b/miner-apps/translator/src/lib/mod.rs
@@ -12,7 +12,11 @@
 //! etc.) for specialized functionalities.
 #![allow(clippy::module_inception)]
 use async_channel::{unbounded, Receiver, Sender};
-use std::{net::SocketAddr, sync::Arc, time::Duration};
+use std::{
+    net::SocketAddr,
+    sync::{Arc, OnceLock},
+    time::Duration,
+};
 use stratum_apps::{
     task_manager::TaskManager, utils::types::Sv2Frame, SHUTDOWN_BROADCAST_CAPACITY,
 };
@@ -27,7 +31,7 @@ use crate::{
     error::TproxyErrorKind,
     status::{State, Status},
     sv1::sv1_server::sv1_server::Sv1Server,
-    sv2::{channel_manager::ChannelMode, ChannelManager, Upstream},
+    sv2::{ChannelManager, Upstream},
     utils::{ShutdownMessage, UpstreamEntry},
 };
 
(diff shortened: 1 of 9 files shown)


miner-apps/translator/src/lib/tproxy_mode.rs
```diff
@@ -0,0 +1,51 @@
+use std::sync::OnceLock;
+
+/// Defines the operational mode for channel management.
```
GitGab19:
```suggestion
/// Defines the operational mode for Translator Proxy.
```

miner-apps/translator/src/lib/tproxy_mode.rs
```diff
@@ -0,0 +1,51 @@
...
+///
+/// The translator can operate in two different modes that affect how
+/// downstream connections are mapped to upstream SV2 channels:
```
GitGab19:
```suggestion
/// It can operate in two different modes that affect how Sv1
/// downstream connections are mapped to the upstream Sv2 channels.
```

miner-apps/translator/src/lib/tproxy_mode.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TproxyMode {
+    /// All downstream connections share a single extended SV2 channel.
+    /// This mode uses extranonce prefix allocation to distinguish between
+    /// different downstream miners while presenting them as a single entity
+    /// to the upstream server. This is more efficient for pools with many
+    /// miners.
```
GitGab19:
```suggestion
    /// All Sv1 downstream connections share a single extended Sv2 channel.
    /// This mode uses extranonce_prefix allocation to distinguish between
    /// different downstream miners while presenting them as a single entity
    /// to the upstream server. This is more efficient for pools with many
    /// miners.
```

miner-apps/translator/src/lib/tproxy_mode.rs
```diff
@@ -0,0 +1,51 @@
...
+    Aggregated,
+    /// Each downstream connection gets its own dedicated extended SV2 channel.
+    /// This mode provides complete isolation between downstream connections
+    /// but may be less efficient for large numbers of miners.
```
GitGab19:
```suggestion
    /// Each Sv1 downstream connection gets its own dedicated extended Sv2 channel.
    /// This mode provides complete isolation between downstream connections
    /// but may be less efficient for large numbers of miners.
```

miner-apps/translator/src/lib/tproxy_mode.rs
```diff
@@ -0,0 +1,51 @@
...
+static TPROXY_MODE: OnceLock<TproxyMode> = OnceLock::new();
+
+pub fn init_tproxy_mode(mode: TproxyMode) {
+    TPROXY_MODE
+        .set(mode)
+        .expect("TPROXY_MODE initialized more than once");
+}
+
+pub fn tproxy_mode() -> TproxyMode {
+    *TPROXY_MODE.get_or_init(|| TproxyMode::Aggregated)
```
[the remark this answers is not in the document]
GitGab19: Maybe we should initialize it only on the first test though? I don't know.

The only problem I have is with the default arbitrary choice of setting this to `Aggregated` in case it's not initialized.

PR #233, opened 2026-02-03
Add machete to CI

Shourya742:
closes: #232 

This PR also removes unused dependency across the sv2-apps.

### DIFF
--- a/miner-apps/mining-device/Cargo.lock
+++ b/miner-apps/mining-device/Cargo.lock
@@ -259,17 +259,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-recursion"
-version = "0.3.2"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -119,4 +119,21 @@ jobs:
...
+        uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install cargo-machete
+        uses: taiki-e/install-action@cargo-machete
+
+      - name: Run cargo-machete
+        run:  cargo machete
```
GitGab19: Why is this different from the example workflow in https://github.com/marketplace/actions/cargo-machete?

Issue #242, opened 2026-02-06
tProxy does not update `nominal_hashrate` on upstream channel when receiving `SetTarget`


gimballock:
<img width="765" height="274" alt="Image" src="https://github.com/user-attachments/assets/8d870172-bf16-465f-9570-be22dc316bd6" />

**Summary**
The tProxy's reported server hashrate (`sv2_server_hashrate_total`) is significantly higher than the pool's reported client hashrate (`sv2_client_hashrate_total`) for the same connection, even though both sides use `get_nominal_hashrate()` to compute the value. This makes the hashrate metric non-comparable across the two applications.

**Root Cause**
When the pool's vardiff algorithm adjusts a channel's difficulty, it calls `ExtendedChannel::update_channel(new_hashrate, None)`, which updates both the channel's `target` and `nominal_hashrate` to reflect the miner's actual observed hashrate.

However, when the tProxy receives the resulting `SetTarget` message from the pool, its `handle_set_target` implementation ([mining_message_handler.rs:865](https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs#L865)) only calls `set_target()` on the upstream extended channel — it *never updates* `nominal_hashrate`.

This means the tProxy's upstream channel retains the originally declared hashrate from `OpenExtendedMiningChannel` (e.g. `min_individual_miner_hashrate` = 10 TH/s from config), while the pool has already adjusted it down to the actual observed hashrate (e.g. ~1.5 TH/s for a Bitaxe).

**Impact**
- **Monitoring:** `sv2_server_hashrate_total` on the tProxy reports a value ~6x higher than `sv2_client_hashrate_total` on the pool for the same miner, making cross-application dashboards misleading.
- **Any logic using** `get_nominal_hashrate()` on the tProxy's upstream channel will see a stale value that diverges further from reality over time as vardiff adjusts.

**Expected Behavior**
When the tProxy receives a `SetTarget` from the pool, it should also update the upstream channel's `nominal_hashrate` to stay consistent with the pool's vardiff estimate. The `ExtendedChannel::update_channel(new_hashrate, None)` method is available for this purpose — the new hashrate can be derived from the new target using the same formula vardiff uses.

**Steps to Reproduce**
[...]

PR #245, opened 2026-02-06
fix(channels): Update channel nominal hashrate from downstream vardiff estimate

gimballock:
In aggregated mode, update the upstream extended channel's nominal hashrate to match the downstream vardiff estimate before forwarding UpdateChannel. In non-aggregated mode, update the specific extended channel's nominal hashrate. This ensures monitoring reports values consistent with downstream vardiff estimates.

See [stratum-mining/sv2-apps#242](https://github.com/stratum-mining/sv2-apps/issues/242)


GitGab19 (approved):
We need to get https://github.com/stratum-mining/stratum/pull/2077 merged before proceeding with this one, otherwise CI is not gonna pass.

PR #247, opened 2026-02-09
tProxy fallbacks after receiving a `SetExtranoncePrefix` from upstream

### LINKED ISSUE #139: tProxy's`handle_set_extranonce_prefix` is wrong
the current tProxy implementation of `handle_set_extranonce_prefix` is wrong:

https://github.com/stratum-mining/sv2-apps/blob/39188c7cf60ecf7db89b54eb3128df7f118a174a/miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs#L308-L317

an extranonce prefix is a contract between client and server... it is initially established via `OpenExtendedMiningChannel.Success` and on Sv2 realm it can potentially change during the lifetime of the channel

in Sv1, we cannot dynamically change it once it's already been established, so it's fair that tProxy needs to treat this message with caution

synchronization of `extranonce_prefix` between server and client is essential for share validity

simply ignoring `SetExtranoncePrefix` will cause subsequent shares to be invalid, because now server and client are operating under different states

this should ~~trigger fallback~~ (or better: force all Sv1 clients to re-connect under a new `extranonce1`), not be ignored

GitGab19:
Closes #139

### DIFF
--- a/integration-tests/tests/translator_integration.rs
+++ b/integration-tests/tests/translator_integration.rs
@@ -1794,3 +1794,172 @@ async fn aggregated_translator_handles_downstream_connecting_during_future_job()
         .wait_for_message(&["mining.submit"], MessageDirection::ToUpstream)
         .await;
 }
+
+/// Verifies that SetExtranoncePrefix triggers fallback since Sv1 clients cannot handle
+/// dynamic extranonce changes. Ignoring this message would cause share validity issues.
+#[tokio::test]
+async fn translator_fallbacks_on_set_extranonce_prefix() {
+    start_tracing();
+
+    // First upstream server mock (will send SetExtranoncePrefix)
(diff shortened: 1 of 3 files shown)


GitGab19:
> And we can let the downstream know about the extranonce change via setExtranonce message rather than performing disconnections?

We cannot rely on the Sv1 `mining.set_extranonce` message because majority of machines doesn't support that.

I simply implemented what @plebhash suggested in the [issue description](https://github.com/stratum-mining/sv2-apps/issues/139#issue-3748888670), which is the fallback.

But sure, another approach could be what you mentioned: disconnect all the clients, clean the old channels, let them reconnect, and create new channels.

GitGab19:
Closing this as stale.

Furthermore, we should ideally disconnect all the Sv1 clients connected, instead of doing the fallback.

Issue #248, opened 2026-02-09
`sv2_client_channel_shares_per_minute` metric reports configured target, not actual share rate


gimballock:
The per-channel Prometheus metric `sv2_client_channel_shares_per_minute{client_id, channel_id, user_identity}` is intended to represent (i believe) the measured share submission rate for each channel. 

However, it actually reports the configured `expected_share_per_minute` — **a static config parameter** — rather than an observed rate.

The value comes from `ExtendedChannel::get_shares_per_minute()` / `StandardChannel::get_shares_per_minute()` in the stratum library, which simply returns `self.expected_share_per_minute`. 

This field is set once at channel creation from the application's config and never updated:
- **Pool**: set from `PoolConfig.shares_per_minute`
- **JD Client**: set from `JdClientConfig.shares_per_minute`
- **Translator Proxy**: set from `DownstreamDifficultyConfig.shares_per_minute`

As a result, every channel on a given application instance reports the same constant value.  The per-channel labels (`client_id`, `channel_id`, `user_identity`) therefore do not reflect any channel-specific behavior.

**What `expected_share_per_minute` actually represents**
This value is a **vardiff tuning parameter**, not a runtime metric. 

It represents the target share submission rate that the vardiff algorithm attempts to achieve for a miner with the declared `nominal_hashrate`.  Internally the vardiff logic (`try_vardiff` in `channels_sv2::vardiff::classic`) computes a `realized_share_per_min` from actual share counts each adjustment cycle, but this value is used transiently and not stored on the channel.

**Options for correctly tracking per channel share rate**

**Option A: Track realized share rate inside the stratum library**
Add a `realized_shares_per_minute` field to `ExtendedChannel`/`StandardChannel`.

The vardiff logic already computes this value:
```rust
let realized_share_per_min = shares_since_last_update as f64 / (delta_time as f64 / 60.0);
```
Persisting this value (one additional f32 and one assignment per vardiff cycle) would make the measured share rate available to downstream consumers.

This approach would:
- allow the existing `sv2_client_channel_shares_per_minute{...}` metric to reflect a true measured rate
- benefit non-Prometheus consumers (e.g. JSON APIs, CLI tools) that cannot derive rates externally

**Option B: Derive share rate externally from share counts**
The per-channel `*_shares_accepted` metrics already expose cumulative share counts. Monitoring systems such as Grafana can derive rates directly:
```promql
rate(sv2_client_channel_shares_accepted[5m])
```

[...]

PR #252, opened 2026-02-10
fix(monitoring): remove redundant sv2_client_channel_shares_per_minute metric

gimballock:
Closes [#248](https://github.com/stratum-mining/sv2-apps/issues/248)

Remove the `sv2_client_channel_shares_per_minute` Prometheus gauge metric and the `shares_per_minute` field from the monitoring channel info structs. This removes it from the API response (/api/v1/clients/{id}/channels) as well.

This metric is redundant — the same information can be derived using `rate(sv2_client_shares_accepted_total[1m])` on the existing per-channel shares accepted counter, which is the idiomatic Prometheus approach.

The `shares_per_minute` config parameter used by vardiff (difficulty adjustment) is unchanged.

### DIFF
--- a/stratum-apps/src/monitoring/prometheus_metrics.rs
+++ b/stratum-apps/src/monitoring/prometheus_metrics.rs
@@ -10,15 +10,13 @@ pub struct PrometheusMetrics {
     // System metrics
     pub sv2_uptime_seconds: Gauge,
     // Server metrics (upstream connection)
-    pub sv2_server_channels_extended: Option<Gauge>,
-    pub sv2_server_channels_standard: Option<Gauge>,
+    pub sv2_server_channels: Option<GaugeVec>,
     pub sv2_server_hashrate_total: Option<Gauge>,
     pub sv2_server_channel_hashrate: Option<GaugeVec>,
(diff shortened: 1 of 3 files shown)


stratum-apps/src/monitoring/prometheus_metrics.rs
```diff
@@ -10,21 +10,16 @@ pub struct PrometheusMetrics {
...
     // Clients metrics (downstream connections)
     pub sv2_clients_total: Option<Gauge>,
-    pub sv2_client_channels_total: Option<Gauge>,
-    pub sv2_client_channels_extended: Option<Gauge>,
-    pub sv2_client_channels_standard: Option<Gauge>,
+    pub sv2_client_channels: Option<GaugeVec>,
     pub sv2_client_hashrate_total: Option<Gauge>,
     pub sv2_client_channel_hashrate: Option<GaugeVec>,
     pub sv2_client_shares_accepted_total: Option<GaugeVec>,
-    pub sv2_client_channel_shares_per_minute: Option<GaugeVec>,
```
GitGab19: Why did you remove this instead of renaming it?

Issue #262, opened 2026-02-11
tProxy hashrate metrics remain stale when `enable_vardiff = false` and upstream adjusts difficulty via `SetTarget`


gimballock:
<img width="749" height="204" alt="Image" src="https://github.com/user-attachments/assets/595b087f-7c79-441d-9e35-6f3d14b2d500" />

**Related:** #242, PR #245 

When `enable_vardiff = false` (the recommended configuration when using a JDC), the tProxy does not update `nominal_hashrate` in response to upstream `SetTarget` messages.

As a result, three hashrate metrics remain stuck at their initial values, even after upstream difficulty changes.

PR #245 fixes the case where the tProxy itself generates `UpdateChannel` (i.e., when `enable_vardiff = true`).  
It does not fix the `SetTarget` path, which is the only path used when vardiff is disabled.

### Affected Metrics
 
| Metric | Why It's Incorrect |
|--------|--------------------|
| `sv2_server_hashrate_total` | `handle_set_target` updates target but not `nominal_hashrate` |
| `sv2_client_hashrate_total` | Same channel, same issue |
| `sv1_hashrate_total` | `d.hashrate` never updated when forwarding difficulty to SV1 miners |

All three metrics remain at the value declared in `OpenExtendedMiningChannel` (derived from `min_individual_miner_hashrate`), regardless of subsequent upstream difficulty adjustments.

### Reproduction
 
1. Configure tProxy with `enable_vardiff = false` (typical when using JDC):
   ```toml
   [downstream_difficulty_config]
   min_individual_miner_hashrate = 10_000_000_000_000.0
   shares_per_minute = 6.0
   enable_vardiff = false

2. Connect a miner whose actual hashrate differs from the configured value.

3. Allow upstream (JDC or pool) to adjust difficulty via `SetTarget`.

4. Observe:
- `sv2_server_hashrate_total` remains at the initial configured value
- `sv1_hashrate_total` remains at the initial configured value
- The upstream pool correctly reflects adjusted hashrate.

**What PR #245 Fixes**

When `enable_vardiff = true`, and the tProxy emits `UpdateChannel`, it now updates `nominal_hashrate`.

**What's missing:**

**1.  `handle_set_target` (SV2 path)**
When receiving `SetTarget`, the handler updates channel target(s) but not `nominal_hashrate`.

**2. SV1 downstream hashrate**
When vardiff is disabled, difficulty updates are forwarded to SV1 miners but `d.hashrate` is never updated.

[...]

PR #263, opened 2026-02-11
fix(monitoring): make nominal_hashrate Optional and omit metrics when vardiff disabled

gimballock:
Closes [#262](https://github.com/stratum-mining/sv2-apps/issues/262)
 
When enable_vardiff = false, the tProxy and JDC never generate UpdateChannel messages, so difficulty adjustments arrive exclusively as SetTarget (tProxy) or are managed internally (JDC). The existing handlers updated targets but not nominal_hashrate, leaving hashrate metrics stuck at initial values.
    
Changes:
- Make nominal_hashrate Option<f32> in server monitoring types
- Update total_hashrate() to use filter_map for Option handling
- Omit hashrate metrics from HTTP endpoints when vardiff is disabled

This ensures hashrate metrics accurately reflect when they can be reliably tracked (vardiff enabled) vs when they cannot (vardiff disabled).

### DIFF
--- a/stratum-apps/src/monitoring/server.rs
+++ b/stratum-apps/src/monitoring/server.rs
@@ -11,7 +11,8 @@ use utoipa::ToSchema;
 pub struct ServerExtendedChannelInfo {
     pub channel_id: u32,
     pub user_identity: String,
-    pub nominal_hashrate: f32,
+    /// None when vardiff is disabled and hashrate cannot be reliably tracked
+    pub nominal_hashrate: Option<f32>,
     pub target_hex: String,
     pub extranonce_prefix_hex: String,
     pub full_extranonce_size: usize,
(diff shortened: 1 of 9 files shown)


stratum-apps/src/monitoring/server.rs
```diff
@@ -11,7 +11,7 @@ use utoipa::ToSchema;
 pub struct ServerExtendedChannelInfo {
     pub channel_id: u32,
     pub user_identity: String,
-    pub nominal_hashrate: f32,
+    pub nominal_hashrate: Option<f32>,
```
GitGab19: Can you add a short comment here which clarifies the `Option`?

GitGab19 (commented):
LGTM, just a small thing.

PR #276, opened 2026-02-17
Fix/extension negotiation race

### LINKED ISSUE #264: Extension Negotiation Flow
Extension negotiation (RequestExtensions → RequestExtensionsSuccess/Error) completes asynchronously after the SV1 server has already started accepting downstream connections. This creates race conditions where extension-dependent behavior cannot be reliably determined at the time it's needed.

## Background: How We Encountered This
While investigating a panic in the translator (PR #197), we discovered:
1. A downstream SV1 miner sends mining.authorize with a username like very-long-pool-username.worker1
2. The translator attempts to create a UserIdentity TLV (Extension n.2 - Worker Specific Hashrate Tracking)  
3. The UserIdentity field has a protocol-specified maximum of 32 bytes
4. If the username exceeds 32 bytes, the translator panics on unwrap()
During PR review, GitGab19 noted: "We don't check the length at authorize time because we don't know if we need to use the extension or not there."
Investigation revealed the translator sends RequestExtensions during setup_connection() but does not wait for the response before returning. The SV1 server starts immediately, accepting downstream connections before extension negotiation completes.

Key code paths:
| Step | File | Lines | Description |
|------|------|-------|-------------|
| Send RequestExtensions | miner-apps/translator/sv2/upstream/upstream.rs | 276-300 | Sends message but doesn't await response |
| Return without waiting | miner-apps/translator/sv2/upstream/upstream.rs | 302 | Ok(()) returned immediately |
| Start SV1 server | miner-apps/translator/lib/mod.rs | 324-330 | sv1_server_instance.start() called before extensions known |
| Async message loop | miner-apps/translator/sv2/channel_manager/channel_manager.rs | 184 | handle_upstream_frame() in select! loop |
| Route to extensions handler | miner-apps/translator/sv2/channel_manager/channel_manager.rs | 244-248 | MessageType::Extensions routed here |
| Store negotiated extensions | miner-apps/translator/sv2/channel_manager/extensions_message_handler.rs | 59-61 | data.negotiated_extensions = supported |

## Proposed Solution
Extension negotiation must complete synchronously before accepting downstream connections:
1. TCP + Noise handshake with upstream
2. Send SetupConnection
3. Receive SetupConnection.Success
4. Send RequestExtensions
5. Receive RequestExtensionsSuccess/Error  ◄── WAIT HERE
6. Return from setup_connection() with known extension state in channel manager
7. Start SV1 server  ◄── EXTENSIONS ARE NOW KNOWN
8. Accept miner connections with full knowledge of capabilities

This way all asynchronous message handling is conducted with the channel-manager having full knowledge of extensions and can handle according.

## Graceful Degradation for Extensions?
Should extensions prevent mining if the Downstream doesn't comply with the extension? 
For the username extension specifically, we could probably just hash whatever string the Downstream Sv1 provides and use that Upstream with the extension since that would be 32 bytes.

average-gary:
Fixes the extension negotiation race condition where Translator and JDC would start accepting downstream connections before knowing which extensions were negotiated with the upstream server.

- Translator & JDC: setup_connection() now waits synchronously for the extension negotiation response (30s timeout) before returning
- Translator: Added username truncation to 32 bytes for UserIdentity TLV, with UTF-8 boundary respect and warning log

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/mod.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/mod.rs
@@ -354,6 +354,16 @@ impl ChannelManager {
         Ok(channel_manager)
     }
 
+    /// Sets the negotiated extensions.
+    ///
+    /// This is used after upstream connection setup to store the extensions
+    /// that were successfully negotiated with the upstream server.
+    pub fn set_negotiated_extensions(&self, extensions: Vec<u16>) {
+        self.channel_manager_data.super_safe_lock(|data| {
+            data.negotiated_extensions = extensions;
+        });
+    }
+
     // Bootstraps a group channel with the given parameters.
     // Returns a `GroupChannel` if successful, otherwise returns `None`.
     //
--- a/miner-apps/jd-client/src/lib/error.rs
+++ b/miner-apps/jd-client/src/lib/error.rs
@@ -224,9 +224,11 @@ pub enum JDCErrorKind {
     RequiredExtensionsNotSupported(Vec<u16>),
     /// Server requires extensions that the translator doesn't support
     ServerRequiresUnsupportedExtensions(Vec<u16>),
-    /// BitcoinCoreSv2TDP cancellation token activated
-    BitcoinCoreSv2TDPCancellationTokenActivated,
-    /// Failed to create BitcoinCoreSv2TDP tokio runtime
+    /// Extension negotiation timed out waiting for response
+    ExtensionNegotiationTimeout,
+    /// BitcoinCoreSv2 cancellation token activated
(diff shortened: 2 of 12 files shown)


GitGab19:
@average-gary can you solve conflicts and rebase this PR?

I'd like to take a look at it in the near future :)

GitGab19:
Any news on this @average-gary ?

We can open a new PR in the future if you think it's hard to update this one.

average-gary:
Closing as stale. Will re-open if the issue presents again, although, it may be resolved at this point.

GitGab19:
I'm pretty sure the issue is not resolved yet, but I'll try to work on it in the next milestone.

PR #292, opened 2026-02-23
feat(monitoring): track all share response outcomes per channel

gimballock:
Previously, only accepted shares were tracked via ShareAccounting from stratum-core. When a share was rejected (stale, invalid job, difficulty too low, etc.) the pool logged an error and sent SubmitSharesError but no counters were incremented — meaning there was no way to assess the rate of share rejection or diagnose the root cause of revenue loss.

This adds per-channel ShareResponseCounts that track every share submission outcome the pool produces:

  - accepted (valid + block-found)
  - stale-share
  - invalid-job-id
  - invalid-share
  - difficulty-too-low
  - duplicate-share
  - bad-extranonce-size
  - invalid-channel-id

Counters are incremented inline in the pool's share validation handlers (handle_submit_shares_standard / handle_submit_shares_extended) and stored in ChannelManagerData alongside the existing vardiff state.

The monitoring pipeline is extended end-to-end:

  - ShareResponseCounts struct added to monitoring client types
  - Pool monitoring reads counts from ChannelManagerData and maps them into the channel info structs (ExtendedChannelInfo, StandardChannelInfo)
  - JSON API exposes share_responses per channel (omitted when absent)
  - New Prometheus GaugeVec sv2_client_shares_rejected_total with labels [client_id, channel_id, user_identity, reason] enables alerting on rejection spikes and dashboarding rejection rates by error type
  - OpenAPI schema updated to include ShareResponseCounts

This enables operators to:
  - Monitor share rejection rate as a fraction of total submissions
  - Identify whether rejections are dominated by stale shares (latency), invalid job IDs (template propagation), or other causes
  - Set alerts when rejection rates exceed acceptable thresholds
  - Correlate rejection patterns with infrastructure changes

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
@@ -29,6 +29,8 @@ use stratum_apps::{
 };
 use tracing::{debug, error, info, warn};
 
+use stratum_apps::monitoring::client::ShareResponseCounts;
+
 use crate::{
     channel_manager::{
         ChannelManager, ChannelManagerChannel, SharesOrderedByDiff, FULL_EXTRANONCE_SIZE,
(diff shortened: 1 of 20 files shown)


gimballock:
Added SV1 share rejection monitoring

gimballock:
yes, it's possible we can find a smarter way to count what's important here.

Should we close with the intention to reopen and reassess once we can account for rejections in a more useful way?

PR #294, opened 2026-02-23
fix(config): make server_id, supported_extensions, required_extensions optional

### LINKED ISSUE #115: Make server_id, supported_extensions and required_extensions optional
It's a bit annoying to have to add this to all my existing config files:

```toml
server_id = 1

supported_extensions = [
]
 
required_extensions = [
]
```

Would be nice if they can be omitted if empty. The default `server_id` can just be random or `0`.

average-gary:
Add #[serde(default)] to server_id (pool), supported_extensions, and required_extensions fields across pool, translator, and jd-client configs. This allows these fields to be omitted from TOML config files, defaulting to 0 and empty vectors respectively.

Closes #115

### DIFF
--- a/docker/config/translator-proxy-config.toml.template
+++ b/docker/config/translator-proxy-config.toml.template
@@ -28,7 +28,7 @@ aggregate_channels = ${TPROXY_AGGREGATE_CHANNELS}
 # Extensions that the translator supports (will request if required by server)
 # Example: supported_extensions = [0x0002, 0x0003]
 supported_extensions = [
-    # 0x0002,  # Worker-Specific Hashrate Tracking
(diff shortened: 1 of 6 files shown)


miner-apps/jd-client/src/lib/config.rs
GitGab19: Can you clean up JDC's config example files, by removing the (already commented out) `required_extensions`?

pool-apps/pool/src/lib/config.rs
GitGab19: Can you clean up tProxy's config example files, by removing the `supported_extensions` and `required_extensions`?

Issue #145, opened 2025-12-22
fix Pool memory leaks on `HashMap`s


plebhash:
as noted [here](https://github.com/stratum-mining/sv2-apps/pull/138#discussion_r2639993297)

there's multiple` HashMap`s where we only add elements, but never clean them

plebhash:
while tackling this, we should make sure `CloseChannel` paths (in both directions) trigger the proper cleanups

arguably broader scope than just "HashMap leaks", but still closely related to be tackled here instead of a standalone issue

context: https://github.com/stratum-mining/sv2-apps/pull/310#discussion_r3109133730

PR #310, opened 2026-03-02
Adapt apps to new extranonce APIs

GitGab19:
This PR adapts Pool, JDC, and tProxy to the new extranonce APIs introduced by https://github.com/stratum-mining/stratum/pull/2098

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/mod.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/mod.rs
@@ -19,6 +19,7 @@ use stratum_apps::{
         bitcoin::{consensus, Amount, Target, TxOut},
         channels_sv2::{
             client::extended::ExtendedChannel,
+            extranonce_manager::{bytes_needed, ExtranonceAllocator},
             outputs::deserialize_outputs,
             server::{
                 group::GroupChannel,
@@ -39,10 +40,7 @@ use stratum_apps::{
         job_declaration_sv2::{
             AllocateMiningJobToken, AllocateMiningJobTokenSuccess, DeclareMiningJob,
         },
-        mining_sv2::{
-            ExtendedExtranonce, OpenExtendedMiningChannel, SetCustomMiningJob, SetTarget,
-            UpdateChannel,
-        },
+        mining_sv2::{OpenExtendedMiningChannel, SetCustomMiningJob, SetTarget, UpdateChannel},
         parsers_sv2::{AnyMessage, JobDeclaration, Mining, TemplateDistribution, Tlv},
         template_distribution_sv2::{NewTemplate, SetNewPrevHash as SetNewPrevHashTdp},
     },
@@ -74,17 +72,62 @@ mod jd_message_handler;
 mod template_message_handler;
 mod upstream_message_handler;
 
-pub const JDC_SEARCH_SPACE_BYTES: usize = 4;
-// These are only used for solo-mining, very similar to pool
-const CLIENT_SEARCH_SPACE_BYTES: usize = 16;
-pub const FULL_EXTRANONCE_SIZE: usize = JDC_SEARCH_SPACE_BYTES + CLIENT_SEARCH_SPACE_BYTES;
-// some extra bytes guaranteed for standard channels
-// allows for 65,536 standard channels on worst case
-// where worst-case means: OpenExtendedMiningChannel.Success.extranonce_size ==
-// OpenExtendedMiningChannel.min_extranonce_size
-const STANDARD_CHANNEL_ALLOCATION_BYTES: usize = 2;
-// for OpenExtendedMiningChannel.min_extranonce_size
-const MIN_EXTRANONCE_SIZE: usize = JDC_SEARCH_SPACE_BYTES + STANDARD_CHANNEL_ALLOCATION_BYTES;
+// ============================================================================
+// JDC extranonce layout
+// ============================================================================
+//
+// JDC multiplexes many downstream channels (standard + extended) over a
+// **single** upstream extended channel whose `extranonce_size` is fixed at
+// opening time — JDC cannot grow it later. Every SV2 extranonce JDC
+// produces is therefore laid out as:
+//
+//     | upstream_prefix | local_index | downstream rollable |
+//
+// * `upstream_prefix`: pool-assigned (empty in solo mining).
+// * `local_index`: JDC's per-channel slot; width = [`JDC_LOCAL_PREFIX_BYTES`].
+// * `downstream rollable`: what an extended downstream rolls (absent for standard downstreams).
+//
+// The size JDC asks the pool for at open time must cover **every** future
+// downstream, not just the one that triggered the open. See
+// [`MIN_EXTRANONCE_SIZE`] and `handle_downstream_message` for the formula.
+
+/// Maximum number of concurrent downstream channels JDC can allocate.
+/// Determines [`JDC_LOCAL_PREFIX_BYTES`] via [`bytes_needed`].
+const JDC_MAX_CHANNELS: u32 = 65_536;
+
+/// Bytes consumed by JDC's per-channel `local_index`. Derived from
+/// [`JDC_MAX_CHANNELS`] so the two stay in sync.
+const JDC_LOCAL_PREFIX_BYTES: u8 = bytes_needed(JDC_MAX_CHANNELS);
+
+/// Retroactive `extranonce_size` commitment JDC honors for extended
+/// downstreams that attach *after* the upstream channel is already open
+/// (e.g. when the first downstream was standard). Downstreams asking for
+/// more than this are only satisfiable if they arrive first — see
+/// [`MIN_EXTRANONCE_SIZE`] and `handle_downstream_message`.
+///
+/// `8` comfortably covers the common SV1 `extranonce2_size = 4` case and
+/// is easily granted by any reasonable pool.
+const MIN_DOWNSTREAM_ROLLABLE_BYTES: u8 = 8;
+
+/// Lower bound on the `min_extranonce_size` JDC sends in
+/// `OpenExtendedMiningChannel` to the pool. The actual value sent is
+/// `max(MIN_EXTRANONCE_SIZE, JDC_LOCAL_PREFIX_BYTES + M)` where `M` is the
+/// first downstream's request (or `0` for a standard first downstream).
+///
+/// ```text
+/// | local_index (2) | downstream rollable (8) |
+/// |<----- MIN_EXTRANONCE_SIZE = 10 ---------->|
+/// ```
+const MIN_EXTRANONCE_SIZE: u8 = JDC_LOCAL_PREFIX_BYTES + MIN_DOWNSTREAM_ROLLABLE_BYTES;
+
+/// Total extranonce length used by JDC in **solo mining** mode (no upstream
+/// pool). Mirrors the Pool's layout so both modes produce consistent shapes.
+///
+/// ```text
+/// | local_index (2) | downstream rollable (18) |
+/// |<----- SOLO_FULL_EXTRANONCE_SIZE = 20 ---------->|
+/// ```
+pub const SOLO_FULL_EXTRANONCE_SIZE: u8 = 20;
 
 /// A `DeclaredJob` encapsulates all the relevant data associated with a single
 /// job declaration, including its template, optional messages, coinbase output,
@@ -116,12 +159,12 @@ pub struct ChannelManagerData {
     // Mapping of `downstream_id` → `Downstream` object,
     // used by the channel manager to locate and interact with downstream clients.
     pub downstream: HashMap<DownstreamId, Downstream>,
-    // Extranonce prefix factory for **extended downstream channels**.
-    // Each new extended downstream receives a unique extranonce prefix.
-    extranonce_prefix_factory_extended: ExtendedExtranonce,
-    // Extranonce prefix factory for **standard downstream channels**.
-    // Each new standard downstream receives a unique extranonce prefix.
-    extranonce_prefix_factory_standard: ExtendedExtranonce,
+    // Unified extranonce prefix allocator shared by standard and extended
+    // downstream channels. Rebuilt with the upstream-assigned prefix whenever
+    // the upstream connection is (re)negotiated or `SetExtranoncePrefix` is
+    // received. The allocated [`ExtranoncePrefix`] is stored on the channel
+    // itself, so dropping the channel automatically releases the slot.
+    extranonce_allocator: ExtranonceAllocator,
     // Factory that generates **monotonically increasing request IDs**
     // for messages sent from the JDC.
     request_id_factory: AtomicU32,
@@ -189,19 +232,12 @@ impl ChannelManagerData {
         self.downstream_id_factory = AtomicUsize::new(0);
         self.request_id_factory = AtomicU32::new(0);
 
-        let (range_0, range_1, range_2) = {
-            let range_1 = 0..JDC_SEARCH_SPACE_BYTES;
-            (
-                0..range_1.start,
-                range_1.clone(),
-                range_1.end..FULL_EXTRANONCE_SIZE,
-            )
-        };
-        self.extranonce_prefix_factory_extended =
-            ExtendedExtranonce::new(range_0.clone(), range_1.clone(), range_2.clone(), None)
-                .expect("valid ranges");
-        self.extranonce_prefix_factory_standard =
-            ExtendedExtranonce::new(range_0, range_1, range_2, None).expect("valid ranges");
+        // Reset the allocator to its solo-mining default. When upstream
+        // reconnects with a new extranonce prefix it will be rebuilt via
+        // [`ExtranonceAllocator::from_upstream_prefix`] in the upstream handler.
+        self.extranonce_allocator =
+            ExtranonceAllocator::new(Vec::new(), SOLO_FULL_EXTRANONCE_SIZE, JDC_MAX_CHANNELS)
+                .expect("Failed to create ExtranonceAllocator with valid parameters");
 
         self.allocate_tokens.clear();
         self.upstream_channel = None;
@@ -328,27 +364,16 @@ impl ChannelManager {
         supported_extensions: Vec<u16>,
         required_extensions: Vec<u16>,
     ) -> JDCResult<Self, error::ChannelManager> {
-        let (range_0, range_1, range_2) = {
-            let range_1 = 0..JDC_SEARCH_SPACE_BYTES;
-            (
-                0..range_1.start,
-                range_1.clone(),
-                range_1.end..FULL_EXTRANONCE_SIZE,
-            )
-        };
-
-        let make_extranonce_factory = || {
-            ExtendedExtranonce::new(range_0.clone(), range_1.clone(), range_2.clone(), None)
-                .expect("Failed to create ExtendedExtranonce with valid ranges")
-        };
-
-        let extranonce_prefix_factory_extended = make_extranonce_factory();
-        let extranonce_prefix_factory_standard = make_extranonce_factory();
+        // Start with a solo-mining allocator (no upstream prefix). Once the
+        // upstream channel is opened in `handle_open_extended_mining_channel_success`
+        // this allocator is replaced with one built from the upstream prefix.
+        let extranonce_allocator =
+            ExtranonceAllocator::new(Vec::new(), SOLO_FULL_EXTRANONCE_SIZE, JDC_MAX_CHANNELS)
+                .map_err(JDCError::<error::ChannelManager>::shutdown)?;
 
         let channel_manager_data = Arc::new(Mutex::new(ChannelManagerData {
             downstream: HashMap::new(),
-            extranonce_prefix_factory_extended,
-            extranonce_prefix_factory_standard,
+            extranonce_allocator,
             downstream_id_factory: AtomicUsize::new(0),
             request_id_factory: AtomicU32::new(0),
             sequence_number_factory: AtomicU32::new(1),
@@ -409,9 +434,12 @@ impl ChannelManager {
                     data.upstream_channel
                         .as_ref()
                         .map(|channel| channel.get_full_extranonce_size())
-                        .unwrap_or(FULL_EXTRANONCE_SIZE), /* Default to FULL_EXTRANONCE_SIZE if
-                                                           * upstream channel is not present
-                                                           * (solo mining mode) */
+                        .unwrap_or(SOLO_FULL_EXTRANONCE_SIZE as usize), /* Default to
+                                                                         * SOLO_FULL_EXTRANONCE_SIZE if
+                                                                         * upstream channel is
+                                                                         * not
+                                                                         * present
+                                                                         * (solo mining mode) */
                     data.pool_tag_string.clone(),
                     data.last_future_template
                         .clone()
@@ -885,7 +913,22 @@ impl ChannelManager {
                                     .try_into()
                                     .map_err(JDCError::shutdown)?;
                                 upstream_message.request_id = 1;
-                                upstream_message.min_extranonce_size += MIN_EXTRANONCE_SIZE as u16;
+                                // The upstream extended channel is opened once and its
+                                // `extranonce_size` is fixed. Size its rollable region to fit:
+                                //   - JDC's own `local_index` (JDC_LOCAL_PREFIX_BYTES), plus
+                                //   - the larger of the downstream's request `M` and JDC's
+                                //     retroactive commitment to future downstreams
+                                //     (MIN_DOWNSTREAM_ROLLABLE_BYTES) — i.e. never fall below
+                                //     MIN_EXTRANONCE_SIZE.
+                                // Equivalently: max(MIN_EXTRANONCE_SIZE,
+                                //                   JDC_LOCAL_PREFIX_BYTES + M).
+                                let downstream_min = upstream_message.min_extranonce_size as usize;
+                                let upstream_min = std::cmp::max(
+                                    MIN_EXTRANONCE_SIZE as usize,
+                                    (JDC_LOCAL_PREFIX_BYTES as usize)
+                                        .saturating_add(downstream_min),
+                                );
+                                upstream_message.min_extranonce_size = upstream_min as u16;
                                 let upstream_message =
                                     Mining::OpenExtendedMiningChannel(upstream_message)
                                         .into_static();
@@ -940,6 +983,11 @@ impl ChannelManager {
                                 .compare_and_set(UpstreamState::NoChannel, UpstreamState::Pending)
                                 .is_ok()
                             {
+                                // The first downstream is a standard channel, which doesn't
+                                // roll the extranonce itself. Ask the pool for exactly
+                                // MIN_EXTRANONCE_SIZE so we still honor our retroactive
+                                // commitment (MIN_DOWNSTREAM_ROLLABLE_BYTES) to any later
+                                // extended downstream that attaches to this upstream.
                                 let upstream_open = OpenExtendedMiningChannel {
                                     user_identity: self.user_identity.clone().try_into().unwrap(),
                                     request_id: 1,
--- a/miner-apps/jd-client/src/lib/channel_manager/mod.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/mod.rs
@@ -88,8 +88,10 @@ mod upstream_message_handler;
 // * `downstream rollable`: what an extended downstream rolls (absent for standard downstreams).
 //
 // The size JDC asks the pool for at open time must cover **every** future
-// downstream, not just the one that triggered the open. See
-// [`MIN_EXTRANONCE_SIZE`] and `handle_downstream_message` for the formula.
+// downstream, not just the one that triggered the open. The floor on the
+// rollable region is configured by
+// [`JobDeclaratorClientConfig::reserved_downstream_rollable_extranonce_size`];
+// see `handle_downstream_message` for the exact formula.
 
 /// Maximum number of concurrent downstream channels JDC can allocate.
 /// Determines [`JDC_LOCAL_PREFIX_BYTES`] via [`bytes_needed`].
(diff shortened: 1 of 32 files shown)


bitcoin-core-sv2/Cargo.toml
```diff
@@ -21,6 +21,6 @@ async-channel = "1.5.1"
 # fetching from github enables synchronizing development workflows across sv2-apps and stratum repos
 # it MUST be changed before bitcoin-core-sv2 is published to crates.io
 # with the proper version of stratum-core being fetched from crates.io as well
-stratum-core = { git = "https://github.com/stratum-mining/stratum", branch = "main" }
+stratum-core = { git = "https://github.com/GitGab19/stratum", branch = "extranonce-allocation" }
```
plebhash: should we make this PR a draft while this fork is used?

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -424,18 +423,27 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
...
                                             messages_results.push(Ok((
                                                 *downstream_id,
                                                 Mining::SetExtranoncePrefix(SetExtranoncePrefix {
                                                     channel_id: *channel_id,
-                                                    extranonce_prefix: prefix.into(),
+                                                    extranonce_prefix: prefix_bytes
+                                                        .try_into()
+                                                        .expect(
+                                                            "Prefix will always be less than 32",
+                                                        ),
```
bit-aloo: lets propagate a shutdown error here

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -450,21 +458,30 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
...
```
bit-aloo: lets propagate shutdown here

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -75,17 +73,62 @@ mod jd_message_handler;
...
+// The size JDC asks the pool for at open time must cover **every** future
+// downstream, not just the one that triggered the open. See
+// [`MIN_EXTRANONCE_SIZE`] and `handle_downstream_message` for the formula.
+
+/// Maximum number of concurrent downstream channels JDC can allocate.
+/// Determines [`JDC_LOCAL_PREFIX_BYTES`] via `bytes_needed(JDC_MAX_CHANNELS)`.
+const JDC_MAX_CHANNELS: usize = 65_536;
+
+/// Bytes consumed by JDC's per-channel `local_index`.
+const JDC_LOCAL_PREFIX_BYTES: usize = 2;
```
bit-aloo: Can we keep it as `JDC_SEARCH_SPACE_BYTES`, I thinks its better to understand than index terminology.

```suggestion
const JDC_SEARCH_SPACE_BYTES: usize = 2;
```

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -117,12 +160,15 @@ pub struct ChannelManagerData {
...
-    extranonce_prefix_factory_standard: ExtendedExtranonce,
+    // Unified extranonce prefix allocator shared by standard and extended
+    // downstream channels. Rebuilt with the upstream-assigned prefix whenever
+    // the upstream connection is (re)negotiated or `SetExtranoncePrefix` is
+    // received.
+    extranonce_allocator: ExtranonceAllocator,
+    // Mapping of `(downstream_id, channel_id)` → `local_index` returned by the
+    // allocator when the channel was opened. Used on `CloseChannel` to free
+    // the slot via [`ExtranonceAllocator::free`].
+    extranonce_local_indexes: HashMap<(DownstreamId, ChannelId), usize>,
```
bit-aloo: Every time we receive a `close channel` message, we need to ensure that the allocated space is freed. Perhaps we can leverage lifetimes so that when a channel associated with an extranonce is dropped, the extranonce is automatically released. I bring this up because it’s very easy to forget this cleanup, which can lead to extranonces remaining blocked.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
bit-aloo: Initially we had this as 4, why are we changing to 2?

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -692,6 +722,19 @@ impl ChannelManager {
...
+                .extranonce_local_indexes
+                .keys()
+                .filter(|(ds_id, _)| *ds_id == downstream_id)
+                .cloned()
+                .collect();
+            for key in freed_keys {
+                if let Some(local_index) = cm_data.extranonce_local_indexes.remove(&key) {
+                    cm_data.extranonce_allocator.free(local_index);
+                }
+            }
```
bit-aloo: We don't need this to be cloned.

```rust
cm_data.extranonce_local_indexes.retain(|(ds_id, _), local_index| {
      if *ds_id == downstream_id {
          cm_data.extranonce_allocator.free(*local_index);
          return false;
      }
      true
  });
```

pool-apps/pool/src/lib/config.rs
```diff
@@ -39,7 +39,7 @@ pub struct PoolConfig {
     #[serde(default, deserialize_with = "opt_path_from_toml")]
     log_file: Option<PathBuf>,
     #[serde(default)]
-    server_id: u16,
+    server_id: u8,
```
bit-aloo, line 42: Isn't server_id two bytes?

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -442,6 +437,19 @@ impl ChannelManager {
...
+                .extranonce_local_indexes
+                .keys()
+                .filter(|(ds_id, _)| *ds_id == downstream_id)
+                .cloned()
+                .collect();
+            for key in freed_keys {
+                if let Some(local_index) = cm_data.extranonce_local_indexes.remove(&key) {
+                    cm_data.extranonce_allocator.free(local_index);
+                }
+            }
```
bit-aloo:
```suggestion
cm_data.extranonce_local_indexes.retain(|(ds_id, _), local_index| {
    if *ds_id == downstream_id {
        cm_data.extranonce_allocator.free(*local_index);
        return false;
    } 
   true
});
```

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -101,6 +101,12 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                 channel_manager_data
                     .vardiff
                     .remove(&(downstream_id, msg.channel_id).into());
+                if let Some(local_index) = channel_manager_data
```
bit-aloo: This covers the case where we receive a `CloseChannel` message from the downstream. In our implementation, there are also cases where we send a `CloseChannel` message to the downstream. We should probably handle extranonce deallocation in those cases as well.

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -652,17 +708,16 @@ impl ChannelManager {
...
-            .get_range2_len();
+        let (new_extranonce_prefix, new_extranonce_size) = self
+            .aggregated_extranonce_allocator
+            .super_safe_lock(|slot| {
+                let allocator = slot
+                    .as_mut()
+                    .expect("aggregated extranonce allocator must be initialised");
+                let rollable_size = allocator.rollable_extranonce_size();
+                let prefix = allocator.allocate_extended(min_extranonce_size).ok();
+                (prefix, rollable_size)
```
bit-aloo: lets not expect inside the closure, maybe return an Option and then process

miner-apps/jd-client/src/lib/channel_manager/mod.rs
bit-aloo: We can derive this from MAX_CHANNELS?

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -191,72 +183,12 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
...
-                        new_open_extended_mining_channel_success.into_static(),
-                    )
-                } else {
-                    // Extranonce size matches, use as-is
-                    self.extended_channels
-                        .insert(m.channel_id, extended_channel);
-                    Ok::<OpenExtendedMiningChannelSuccess<'static>, Self::Error>(m.into_static())
-                }
+                // Non-aggregated mode is a pure passthrough: the upstream channel
+                // is dedicated to this downstream, so whatever prefix and size
```
bit-aloo: This makes a lot of sense. Earlier, every time I saw this block, I was always puzzled about why we needed it.

pool-apps/pool/src/lib/config.rs
GitGab19: I changed it to be 1 byte now (which already allows for 256 different Pool instances), since there's no realistic need for 2 bytes.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: 2 bytes means JDC will be able to support up to 65_536 downstream channels, which is a reasonable number to me.

GitGab19: I'd push back on this, since it makes more sense to use the APIs terminology here, instead of a generic "search_space".

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: With my recent changes on https://github.com/stratum-mining/stratum/pull/2098/ (particularly on https://github.com/stratum-mining/stratum/pull/2098/commits/462f59584c15bed12c4b5d7debe580361735d450), we don't need to do any explicit cleanup anymore. Once the channel is dropped, the local_index is freed automatically.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: Done, I am now using the `byes_needed()` helper function defined in the other companion PR.

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
GitGab19: Are we already cleaning up everything else in those cases? 

I feel we don't have a clear cleaning procedure in those cases, so probably this should be tracked in a broader issue.

But maybe I'm missing something.

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
GitGab19: Done

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
GitGab19: Done

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
GitGab19: I'm now returning a shutdown error here.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -75,17 +73,63 @@ mod jd_message_handler;
...
+// **single** upstream extended channel whose `extranonce_size` is fixed at
+// opening time — JDC cannot grow it later. Every SV2 extranonce JDC
+// produces is therefore laid out as:
+//
+//     | upstream_prefix | local_index | downstream rollable |
+//
+// * `upstream_prefix`: pool-assigned (empty in solo mining).
+// * `local_index`: JDC's per-channel slot; width = [`JDC_LOCAL_PREFIX_BYTES`].
+// * `downstream rollable`: what an extended downstream rolls (zero-padded for standard
+//   downstreams).
```
plebhash:
```suggestion
// * `downstream rollable`: what an extended downstream rolls (absent for standard
//   downstreams).
```

the 0 padding happens inside `extranonce_prefix`

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -75,17 +73,63 @@ mod jd_message_handler;
...
+const MIN_EXTRANONCE_SIZE: u8 = JDC_LOCAL_PREFIX_BYTES + MIN_DOWNSTREAM_ROLLABLE_BYTES;
+
+/// Total extranonce length used by JDC in **solo mining** mode (no upstream
+/// pool). Mirrors the Pool's layout so both modes produce consistent shapes.
+///
+/// ```text
+/// | local_index (2) | downstream rollable (18) |
+/// |<----- FULL_EXTRANONCE_SIZE = 20 ---------->|
+/// ```
+pub const FULL_EXTRANONCE_SIZE: u8 = 20;
```
plebhash:
```suggestion
pub const SOLO_FULL_EXTRANONCE_SIZE: u8 = 20;
```

if this constant is restricted to solo mode only, its name should make that explicitly clear

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -826,7 +863,22 @@ impl ChannelManager {
...
+                                //     MIN_EXTRANONCE_SIZE.
+                                // Equivalently: max(MIN_EXTRANONCE_SIZE,
+                                //                   JDC_LOCAL_PREFIX_BYTES + M).
+                                let downstream_min = upstream_message.min_extranonce_size as usize;
+                                let upstream_min = std::cmp::max(
+                                    MIN_EXTRANONCE_SIZE as usize,
+                                    (JDC_LOCAL_PREFIX_BYTES as usize)
+                                        .saturating_add(downstream_min),
+                                );
+                                upstream_message.min_extranonce_size = upstream_min as u16;
```
plebhash, line 916: I think the logic here leaves room for edge cases

let's imagine this scenario:

- first downstream asks `min_extranonce_size = 4`
- JDC asks upstream for `min_extranonce_size = max(10, 2+4) = 10`
- upstream grants exactly `10`

later:
- another downstream asks `min_extranonce_size = 12`
- JDC needs `2 + 12 = 14` from the upstream channel
- since the upstream channel only has `10`, JDC has to reject the later downstream with `min-extranonce-size-too-large`

on the other hand, if the `min_extranonce_size = 12` downstream had connected first, JDC would have asked upstream for `14`, so the outcome depends on connection order, which is suboptimal.


---

this issue is not being introduced by this PR; JDC already opens only one upstream extended channel and cannot grow its `extranonce_size` later.

but this PR changes the sizing policy and codifies `MIN_DOWNSTREAM_ROLLABLE_BYTES = 8` as the hidden reserve for future downstreams.

I wonder if we should replace `MIN_DOWNSTREAM_ROLLABLE_BYTES` with an explicit JDC config parameter instead? 

this config parameter could remain optional and default to `8` when absent from the `config.toml`

but it would at least give the option for the user to expand this limit, in case there's potential downstream clients that could ask for more

then if first downstream channel request is extended, we would do:
```rust
let reserved_downstream_rollable = config.reserved_downstream_rollable_extranonce_size();
let downstream_min = upstream_message.min_extranonce_size as usize;

let upstream_min = (JDC_LOCAL_PREFIX_BYTES as usize)
    + std::cmp::max(reserved_downstream_rollable as usize, downstream_min);

upstream_message.min_extranonce_size = upstream_min as u16;
```

and if first downstream channel request is standard, we would do:
```rust
let upstream_open = OpenExtendedMiningChannel {
    ...
    min_extranonce_size: (JDC_LOCAL_PREFIX_BYTES
        + config.reserved_downstream_rollable_extranonce_size()) as u16,
}
```

this approach would still leave room for potential edge cases, but they would be under control of the JDC operator, and not ossified by JDC design... so this is arguably better UX

---

note: I see this comment:
```rust
/// `8` comfortably covers the common SV1 `extranonce2_size = 4` case and
/// is easily granted by any reasonable pool.
const MIN_DOWNSTREAM_ROLLABLE_BYTES: u8 = 8;
```

however I think different values are still possible with downstream Sv2 clients, which might or might not be translated from legacy Sv1 devices

---

in summary, I'm not hardly opinionated about the suggested change, mainly because it's hard to evaluate how realistic these kind of edge cases are

so I'd be fine to proceed without it, but we should at least be aware this could happen

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -30,10 +33,45 @@ use stratum_apps::{
...
+
+/// Maximum number of concurrent downstream extended channels the translator
+/// can multiplex over the shared upstream channel in aggregated mode.
+/// Determines [`TPROXY_LOCAL_PREFIX_BYTES`] via [`bytes_needed`].
+pub(crate) const TPROXY_MAX_CHANNELS: usize = 65_536;
+
+/// Bytes consumed by the translator's per-downstream `local_index` inside
+/// the extranonce in aggregated mode. Derived from [`TPROXY_MAX_CHANNELS`]
+/// so the two stay in sync.
+pub(crate) const TPROXY_LOCAL_PREFIX_BYTES: u8 = bytes_needed(TPROXY_MAX_CHANNELS as u32);
```
plebhash:
```suggestion
pub(crate) const AGGREGATED_TPROXY_LOCAL_PREFIX_BYTES: u8 = bytes_needed(TPROXY_MAX_CHANNELS as u32);
```

if this is only used for aggregated mode, we should make it explicitly clear in the constant name

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -30,10 +33,45 @@ use stratum_apps::{
...
+//
+// * **Non-aggregated** — each downstream gets its own upstream extended channel. The translator is
+//   a pure passthrough: the downstream's `OpenExtendedMiningChannel` is forwarded upstream
+//   unchanged, and the pool's `OpenExtendedMiningChannelSuccess` is forwarded to the downstream
+//   unchanged. No extranonce allocator is involved.
+
+/// Maximum number of concurrent downstream extended channels the translator
+/// can multiplex over the shared upstream channel in aggregated mode.
+/// Determines [`TPROXY_LOCAL_PREFIX_BYTES`] via [`bytes_needed`].
+pub(crate) const TPROXY_MAX_CHANNELS: usize = 65_536;
```
plebhash:
```suggestion
pub(crate) const AGGREGATED_TPROXY_MAX_CHANNELS: usize = 65_536;
```

if this is only used for aggregated mode, we should make it explicitly clear in the constant name

miner-apps/jd-client/src/lib/channel_manager/mod.rs
plebhash: but now we're storing the owned `ExtranoncePrefix` inside `HashMap`s, and dropping the channel does not guarantee those will be freed

so I'd say we're only trading one footgun (forgetting to call `free`) for another (forgetting to clean `HashMap`)?

plebhash: I mean, I don't know whether the `HashMap`s were part of the design before the review comment above was addressed, maybe they weren't necessarily introduced as a way to remove the need to call `free`

and I'm also not opposing the existence of the `HashMap`s either

I'm just trying to point out that there's still potential footguns here (although I don't think it's super critical)

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
plebhash: unresolving this

I agree that the code being removed in this hunk is very complex and confusing, but I'm worried that this oversimplification could cause regressions

we should remember that the notion of `min_extranonce_size` (which sets a lower boundary, but allows for bigger values) is a Sv2 concept

before this change, the user is statically setting `downstream_extranonce2_size`, which is enforced on all Sv1 clients

now, even if the user sets a specific `downstream_extranonce2_size` based on the known restrictions of their Sv1 clients, their Sv1 clients might still get a bigger `extranonce2_size` depending on the Sv2 pool policy

let's look at a concrete example:
- user chooses `downstream_extranonce2_size = 8` (after reading cgminer restrictions on the `.toml` example)
- tProxy sends `OpenExtendedMiningChannel.min_extranonce_size = 8`
- Pool replies `OpenExtendedMiningChannel.Success.extranonce_size = 12`

previously, tProxy would enforce `extranonce2 = 8` for cgminer, while now we're passing-through the pool-granted `12` and violating cgminer restrictions, despite the fact that:
- this limitation is documented on the `.toml` example
- the user chose this specific value because of the known limitation

plebhash: another example where this might go wrong: https://github.com/jstefanop/Apollo-Miner-Binaries/issues/19

despite the issue title, FutureBit Apollo II has restrictions on the sizes of `extranonce1` (which is somewhat irrelevant here) and `extranonce2` (which could break with regressions from this PR)

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
plebhash: took note on #143 and #145

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -117,12 +161,16 @@ pub struct ChannelManagerData {
...
+    // Unified extranonce prefix allocator shared by standard and extended
+    // downstream channels. Rebuilt with the upstream-assigned prefix whenever
+    // the upstream connection is (re)negotiated or `SetExtranoncePrefix` is
+    // received.
+    extranonce_allocator: ExtranonceAllocator,
+    // Owns the extranonce prefix issued by the allocator for every open
+    // channel, keyed by `(downstream_id, channel_id)`. The entry is inserted
+    // when the channel is opened and removed when it is closed (explicit
+    // `CloseChannel` or downstream disconnect).
+    extranonce_prefixes: HashMap<(DownstreamId, ChannelId), ExtranoncePrefix>,
```
bit-aloo: Do we still need this? I thought we could let the channel store this value, so when the channel is dropped, `ExtranoncePrefix` is dropped as well. We can query the extranonce prefix from the channel itself if we ever need to read it. I think this can be removed by tweaking the channel API to store `ExtranoncePrefix`.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -286,27 +328,17 @@ impl ChannelManager {
...
-        };
-
-        let extranonce_prefix_factory_extended = make_extranonce_factory();
-        let extranonce_prefix_factory_standard = make_extranonce_factory();
+        // Start with a solo-mining allocator (no upstream prefix). Once the
+        // upstream channel is opened in `handle_open_extended_mining_channel_success`
+        // this allocator is replaced with one built from the upstream prefix.
+        let extranonce_allocator =
+            ExtranonceAllocator::new(FULL_EXTRANONCE_SIZE, None, JDC_MAX_CHANNELS as u32)
+                .expect("Failed to create ExtranonceAllocator with valid parameters");
```
bit-aloo: lets propagate shutdown. I would later open a PR, to remove all expects with shutdowns.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -367,9 +399,11 @@ impl ChannelManager {
...
                         .as_ref()
                         .map(|channel| channel.get_full_extranonce_size())
-                        .unwrap_or(FULL_EXTRANONCE_SIZE), /* Default to FULL_EXTRANONCE_SIZE if
-                                                           * upstream channel is not present
-                                                           * (solo mining mode) */
+                        .unwrap_or(FULL_EXTRANONCE_SIZE as usize), /* Default to
+                                                                    * FULL_EXTRANONCE_SIZE if
+                                                                    * upstream channel is not
+                                                                    * present
+                                                                    * (solo mining mode) */
```
bit-aloo: Can we remove this way of adding comment with, double slash comment before the `unwrap_or`. Sorry for begin nitty here.

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -118,9 +124,6 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
                 let prefix_len = msg.extranonce_prefix.len();
 
                 let total_len = prefix_len + msg.extranonce_size as usize;
```
bit-aloo:
```suggestion
                let total_len = prefix_len as u8 + msg.extranonce_size;
```

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -75,17 +73,63 @@ mod jd_message_handler;
...
+// * `downstream rollable`: what an extended downstream rolls (zero-padded for standard
+//   downstreams).
+//
+// The size JDC asks the pool for at open time must cover **every** future
+// downstream, not just the one that triggered the open. See
+// [`MIN_EXTRANONCE_SIZE`] and `handle_downstream_message` for the formula.
+
+/// Maximum number of concurrent downstream channels JDC can allocate.
+/// Determines [`JDC_LOCAL_PREFIX_BYTES`] via [`bytes_needed`].
+const JDC_MAX_CHANNELS: usize = 65_536;
```
bit-aloo:
```suggestion
const JDC_MAX_CHANNELS: u32 = 65_536;
```

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -129,15 +132,14 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
...
 
-                let extranonces = match ExtendedExtranonce::from_upstream_extranonce(
-                    msg.extranonce_prefix.clone().into(),
-                    range_0,
-                    range_1,
-                    range_2,
+                let extranonce_allocator = match ExtranonceAllocator::from_upstream(
+                    msg.extranonce_prefix.to_vec(),
+                    total_len as u8,
+                    JDC_MAX_CHANNELS as u32,
```
bit-aloo: We won't need these typecast once we correct the types above.

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
bit-aloo:
```suggestion
pub(crate) const TPROXY_MAX_CHANNELS: u32 = 65_536;
```

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -52,20 +53,32 @@ use crate::{
...
-const POOL_ALLOCATION_BYTES: usize = 4;
-const CLIENT_SEARCH_SPACE_BYTES: usize = 16;
-pub const FULL_EXTRANONCE_SIZE: usize = POOL_ALLOCATION_BYTES + CLIENT_SEARCH_SPACE_BYTES;
+// Static identifier for this pool server, placed at the start of the pool's
+// extranonce allocation. One byte covers up to 256 distinct pool servers.
+const POOL_SERVER_BYTES: u8 = 1;
+// Maximum number of concurrent channels the pool can allocate. Determines
+// [`POOL_LOCAL_PREFIX_BYTES`] via [`bytes_needed`]. The internal allocation
+// bitmap uses `POOL_MAX_CHANNELS / 8` bytes of RAM.
+const POOL_MAX_CHANNELS: usize = 16_777_216;
```
bit-aloo:
```suggestion
const POOL_MAX_CHANNELS: u32 = 16_777_216;
```

miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -450,21 +461,33 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
...
                                         ) {
                                         Ok(prefix) => {
+                                            let prefix_bytes = prefix.as_bytes().to_vec();
+                                            channel_manager_data
+                                                .extranonce_prefixes
+                                                .insert((*downstream_id, *channel_id), prefix);
                                             extended_channel
-                                                .set_extranonce_prefix(prefix.clone().to_vec())
+                                                .set_extranonce_prefix(prefix_bytes.clone())
                                                 .expect("Prefix will always be less than 32");
```
bit-aloo: lets propagate shutdown here.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: The hashmaps were here since the beginning actually, but I'm currently working in the direction described by @Shourya742 [here](https://github.com/stratum-mining/sv2-apps/pull/310#discussion_r3122241696).

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: Yeah I expected this kind of comments, let me explain my rationale here.

While working on this, I looked at Braiins Hashpower and Nicehash marketplaces, and they both impose a low limit to the `extranonce2_size` (eg. it has to be bigger than 5 bytes). 

So I though: why not starting to use the `downstream_extranonce2_size` field we have in the config as the corresponding Sv2 `min_extranonce_size`? That's why I simplified the code here (I should have renamed the config field to make my choice more clear).

Then, I tested this PR with both a cpu-miner and a Bitaxe, and they were both working well.

The idea I had was to run some tests with rented hashpower once this PR is finalized (before merging it) to additionally assess this.

Of course this changes if we're already aware of some machines/firmware(s) which do have an upper limit about `extranonce2_size`. From the link above it's not clear to me if the Apollo has an upper limit on it or not though.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: After the changes introduced on the companion `stratum` PR, this is fixed as discussed during PR Review Club call.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: After the changes introduced on the companion `stratum` PR, this is fixed as discussed during PR Review Club call.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: After last changes, tProxy is now respecting the `downstream_extranonce2_size` value set in the config file, and adjusts the extranonce prefix to achieve that.

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -52,20 +53,28 @@ use crate::{
 mod mining_message_handler;
 mod template_distribution_message_handler;
 
-const POOL_ALLOCATION_BYTES: usize = 4;
-const CLIENT_SEARCH_SPACE_BYTES: usize = 16;
-pub const FULL_EXTRANONCE_SIZE: usize = POOL_ALLOCATION_BYTES + CLIENT_SEARCH_SPACE_BYTES;
+// Static identifier for this pool server, placed at the start of the pool's
```
plebhash:
```suggestion
// Size of the static identifier for this pool server, placed at the start of the pool's
```

miner-apps/jd-client/src/lib/channel_manager/mod.rs
GitGab19: Done here https://github.com/stratum-mining/sv2-apps/pull/310/commits/9d73588fc6de715b6c8c4db1cbf4f439b24f18a3

plebhash (approved):

plebhash:
companion merged

@Shourya742 all good here?

@GitGab19 should we update the `Cargo.toml`s?

Shourya742 (approved):
ACK

Shourya742:
> companion merged
> 
> @Shourya742 all good here?
> 
> @GitGab19 should we update the `Cargo.toml`s?

Once @GitGab19 updates the toml should be good

GitGab19:
> > companion merged
> > @Shourya742 all good here?
> > @GitGab19 should we update the `Cargo.toml`s?
> 
> Once @GitGab19 updates the toml should be good

Done.

PR #317, opened 2026-03-04
Remove broadcast channel

### LINKED ISSUE #219: replace `tokio::sync::broadcast` for `SV1Server`
similar to #215, but for

https://github.com/stratum-mining/sv2-apps/blob/fd0ff79c5ac2f0ea3a5c0a8f412e74f7fa9617f5/miner-apps/translator/src/lib/sv1/sv1_server/channel.rs#L26

### LINKED ISSUE #220: replace `tokio::sync::broadcast` for Pool's `ChannelManager` to `Downstream` sender
similar to #215 but for this:

https://github.com/stratum-mining/sv2-apps/blob/fd0ff79c5ac2f0ea3a5c0a8f412e74f7fa9617f5/pool-apps/pool/src/lib/mod.rs#L66-L67

### LINKED ISSUE #221: replace `tokio::sync::broadcast` on JDC's `ChannelManager` to `Downstream` sender
similar to #215 but for this:

https://github.com/stratum-mining/sv2-apps/blob/fd0ff79c5ac2f0ea3a5c0a8f412e74f7fa9617f5/miner-apps/jd-client/src/lib/mod.rs#L89-L90

Shourya742:
closes: #219 #220 #221

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
@@ -111,33 +111,36 @@ impl RouteMessageTo<'_> {
     ) -> Result<(), JDCErrorKind> {
         match self {
             RouteMessageTo::Downstream((downstream_id, message)) => {
-                _ = channel_manager_channel.downstream_sender.send((
-                    downstream_id,
-                    message.into_static(),
-                    None,
-                ));
+                let sender = channel_manager_channel
+                    .downstream_sender
+                    .super_safe_lock(|map| map.get(&downstream_id).cloned());
+                if let Some(sender) = sender {
+                    sender.send((message.into_static(), None)).await?;
+                } else {
+                    debug!("Dropping message for downstream {downstream_id}: no longer connected");
+                }
             }
             RouteMessageTo::Upstream(message) => {
                 if get_jd_mode() != JdMode::SoloMining {
                     let message_static = message.into_static();
                     let sv2_frame: Sv2Frame = AnyMessage::Mining(message_static).try_into()?;
-                    _ = channel_manager_channel
+                    channel_manager_channel
                         .upstream_sender
                         .send(sv2_frame)
-                        .await;
+                        .await?;
                 }
             }
             RouteMessageTo::JobDeclarator(message) => {
-                _ = channel_manager_channel
+                channel_manager_channel
                     .jd_sender
                     .send(message.into_static())
-                    .await;
+                    .await?;
             }
             RouteMessageTo::TemplateProvider(message) => {
-                _ = channel_manager_channel
+                channel_manager_channel
                     .tp_sender
                     .send(message.into_static())
-                    .await;
+                    .await?;
             }
         }
         Ok(())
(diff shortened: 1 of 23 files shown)


miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -111,11 +111,13 @@ impl RouteMessageTo<'_> {
...
-                    downstream_id,
-                    message.into_static(),
-                    None,
-                ));
+                if let Some(sender) = channel_manager_channel
+                    .downstream_sender
+                    .get(&downstream_id)
+                {
+                    let sender = sender.value().clone();
+                    _ = sender.send((message.into_static(), None)).await;
```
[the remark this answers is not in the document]
Shourya742: This thing here can cause deadlock as its across await. Still need some changes to be done. Just a rough sketch.

plebhash:
looks like f3e0cc96c5d95ca14f9619d923f5bc7673436852 is out of scope?

I'm not saying we should remove it, but @Shourya742 can you elaborate on what motivated you to add it here?

Shourya742:
> looks like [f3e0cc9](https://github.com/stratum-mining/sv2-apps/commit/f3e0cc96c5d95ca14f9619d923f5bc7673436852) is out of scope?
> 
> I'm not saying we should remove it, but @Shourya742 can you elaborate on what motivated you to add it here?

Yeah, we’re seeing those non-deterministic tests due to the monitoring tests. I initially thought they might be caused by my changes, but the logs were hard to interpret because the test subsystems weren’t being shut down properly, which led to panics. So, I added commit to make sure atleast logs are sane.

I can remove this commit and open a separate PR.

miner-apps/translator/src/lib/sv1/downstream/downstream.rs
```diff
@@ -326,12 +307,11 @@ impl Downstream {
...
             }
             Err(e) => {
-                let downstream_id = self.downstream_id;
                 error!(
                     "Sv1 message handler error for downstream {}: {:?}",
-                    downstream_id, e
+                    self.downstream_id, e
                 );
-                return Err(TproxyError::disconnect(e, downstream_id));
+                return Err(TproxyError::shutdown(e));
```
plebhash: is this intentional?

why shutdown instead of disconnect?

bit-aloo: I think you’re right, I may have confused `sv1_server_sender` with `sv1_server_receiver`. If the sender errors out, it likely means the `sv1_server` is shutting down, so it makes sense to shut down in that case.

In this scenario, however, if `sv1_server_receiver` errors out, it could indicate that its sender end was dropped, which can happen during an internal downstream disconnection.

Thanks, I don’t recall why I made this change, but switching it to a disconnection makes sense.

plebhash:
> > looks like [f3e0cc9](https://github.com/stratum-mining/sv2-apps/commit/f3e0cc96c5d95ca14f9619d923f5bc7673436852) is out of scope?
> > I'm not saying we should remove it, but @Shourya742 can you elaborate on what motivated you to add it here?
> 
> Yeah, we’re seeing those non-deterministic tests due to the monitoring tests. I initially thought they might be caused by my changes, but the logs were hard to interpret because the test subsystems weren’t being shut down properly, which led to panics. So, I added commit to make sure atleast logs are sane.
> 
> I can remove this commit and open a separate PR.

that's fine we can leave it here, just wanted to understand the motivations

miner-apps/translator/src/lib/sv1/downstream/downstream.rs
plebhash: let's switch back to disconnect then

everything else LGTM

bit-aloo: Should be done now. Thanks

plebhash (approved):

PR #325, opened 2026-03-09
test multiple coinbase outputs round-trip

### LINKED ISSUE #58: Integration tests against TP with multiple coinbase outputs
After #57 it would be good to test against the recently added support for multiple coinbase outputs: https://github.com/stratum-mining/sv2-tp/pull/55

It should be a round trip including submitting the solution, to make sure nothing goes wrong with serialisation.

xyephy:
Add an end-to-end test: mine a real block through TP → Pool → Translator → Minerd, then assert the accepted block's coinbase outputs and that it includes a real mempool transaction.

  Closes #58

### DIFF
--- a/integration-tests/tests/template_provider_integration.rs
+++ b/integration-tests/tests/template_provider_integration.rs
@@ -23,3 +23,76 @@ async fn tp_high_diff() {
     assert_eq!(blockchain_info.difficulty, 77761.1123986095);
     assert_eq!(blockchain_info.chain, "signet");
 }
+
+// This test verifies that coinbase outputs survive the full round-trip through the mining stack.
+// A mempool transaction triggers a witness commitment output, which must be preserved through
+// TP -> Pool -> Translator -> Minerd and back to Bitcoin Core via block submission.
+#[tokio::test]
+async fn tp_coinbase_outputs_round_trip() {
+    start_tracing();
+    let sv2_interval = Some(5);
+    let (tp, tp_addr) = start_template_provider(sv2_interval, DifficultyLevel::Low);
+    tp.fund_wallet().unwrap();
+    let current_block_hash = tp.get_best_block_hash().unwrap();
+
+    let (pool, pool_addr, _) = start_pool(sv2_tp_config(tp_addr), vec![], vec![], false).await;
+
+    // Create a mempool transaction to trigger witness commitment output
+    tp.create_mempool_transaction().unwrap();
+
+    let (translator, tproxy_addr, _) =
+        start_sv2_translator(&[pool_addr], false, vec![], vec![], None, false).await;
+    let (_minerd, _) = start_minerd(tproxy_addr, None, None, false).await;
+
+    wait_for_new_block(
+        &current_block_hash,
+        || tp.get_best_block_hash().unwrap(),
+        "Block should have been mined and accepted within 60 seconds, \
+         confirming coinbase outputs survived the round-trip",
+    )
+    .await;
+    shutdown_all!(pool, translator);
+}
+
+// This test verifies that multiple coinbase outputs survive the full round-trip.
+// Requires a custom Bitcoin Core build with -testmulticoinbase support, which adds extra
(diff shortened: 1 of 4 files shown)


GitGab19:
How is this different wrt integration tests we added in https://github.com/stratum-mining/sv2-apps/pull/321?

Aren't we already testing multiple coinbase outputs in those tests?

xyephy:
> Aren't we already testing multiple coinbase outputs in those tests?

No — #321 checks that the pool correctly maps `user_identity` to payout addresses by deserializing `NewExtendedMiningJob` at the pool boundary. This PR tests that `coinbase_tx_outputs` set by the Template Provider survive the full TP → Pool → Translator → Minerd pipeline, using Bitcoin Core block acceptance as the assertion. If any layer drops or corrupts a coinbase output, the block is consensus-invalid and the test fails.

GitGab19:
I see.

Are you gonna solve conflicts and make this PR up to date?

xyephy:
> I see.
> 
> Are you gonna solve conflicts and make this PR up to date?

yes

integration-tests/tests/template_provider_integration.rs
plebhash, line 1: I don't see how the tests on this file are asserting what this PR aims to achieve.

we simply assert that **some block was mined**, while there's no assertion with regards to how many outputs there are in the coinbase tx

integration-tests/tests/template_provider_integration.rs
```diff
@@ -23,3 +23,69 @@ async fn tp_high_diff() {
...
+    )
+    .await;
+    shutdown_all!(pool, translator);
+}
+
+// This test verifies that multiple coinbase outputs survive the full round-trip.
+// Requires a custom Bitcoin Core build with -testmulticoinbase support, which adds extra
+// OP_RETURN outputs to the coinbase. Set BITCOIN_NODE_BIN to the custom binary path.
+#[tokio::test]
+#[ignore = "requires custom Bitcoin Core build with -testmulticoinbase; set BITCOIN_NODE_BIN"]
```
plebhash: the strategy behind this `ignore` is not clear

plebhash: what does one need to do to get `testmulticoinbase` into Bitcoin Core?

and if it's not readily available on official Bitcoin Core distribution, why should it be covered by SRI CI?

integration-tests/tests/template_provider_integration.rs
xyephy: Fair point, block acceptance alone is too weak. I will replace it with a block-inspection assertion (decode coinbase tx, check count + witness commitment + reward).

integration-tests/tests/template_provider_integration.rs
xyephy: You would need a local Bitcoin Core patch since there's no upstream path. I patched it locally when I was working on stratum-mining/sv2-tp#55, and the upstream PR that would have made it testable without a patch, bitcoin/bitcoin#33890, was closed unmerged.

Agreed it shouldn't be in SRI CI. Dropping the test. Looking for an SRI-compatible alternative for issue #58, will propose before re-pushing.

GitGab19: Given that the idea proposed in https://github.com/bitcoin/bitcoin/pull/33890 wasn't useful for Sv2, and that it got closed, I'm wondering if issue #58 makes sense at all at this point.

As I was already pointing out in [this comment](https://github.com/stratum-mining/sv2-apps/pull/325#issuecomment-4429750618) we already have a test which covers the case of having multiple outputs in a coinbase tx, it's called `pool_solo_mining_partial_donation()`.

Maybe the thing which is missing there is simply the block submission, which could be eventually added very easily without adding anything else.

xyephy:
Updated: dropped the `testmulticoinbase` test (no upstream path after bitcoin/bitcoin#33890 closed). Now a single end-to-end test as suggested; a real block is mined and its coinbase outputs asserted directly. The test also requires the block to include a mempool transaction, so job serialization with real txdata is covered too.

PR #328, opened 2026-03-12
client share accounting refinement

plebhash:
companion https://github.com/stratum-mining/stratum/pull/2114

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -87,9 +87,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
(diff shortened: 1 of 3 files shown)


gimballock:
For when the monitoring/metrics commits come back (via #2119): the earlier rename of `validated_shares` to `submitted_shares` in the app layer collides with the existing `shares_submitted` field on `ServerExtendedChannelInfo` (which counts shares *sent* to upstream — a different concept). Recommend keeping `validated_shares` consistently across both layers to avoid confusion in dashboards and PromQL queries.

PR #352, opened 2026-03-20
refine how JDS handles downstream disconnections

### LINKED ISSUE #351: JDS needs refinements around downstream disconnections
on #299 we left some inefficiencies on how JDS handles downstream disconnections, namely:
- we don't kill the corresponding task spawned via `spawn_io_tasks`, leaving a zombie task behind, wasting the runtime executor resources
- we don't call `JobDeclarator::cleanup_downstream` when the spawned loop of `Downstream::start` breaks

plebhash:
follow-up to #299 

close #351


Shourya742 (approved):
ACK

PR #342, opened 2026-03-18
Add dependabot to sv2

Shourya742:
This PR adds Dependabot, a bot that automatically keeps project dependencies up to date. It introduces a `dependabot.yml` configuration that checks daily for updates to `stratum-core`. If an update is available, it opens a PR to bump the dependency.

Currently, it will open three separate PRs for `pool-apps`, `miner-apps`, and `integration-test`. Once we adopt a more structured workspace layout after the JDS merge, this will be consolidated into a single PR.

more on dependabot here: https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/dependabot-quickstart-guide

PR #353, opened 2026-03-20
ci: replace stratum freshness check with event-driven stratum-core sync

plebhash:
alternative to #342

companion https://github.com/stratum-mining/stratum/pull/2121

replaces the old `stratum-freshness` fail-only check with an event-driven lockfile update workflow for `stratum-core`


plebhash:
> Have you had the chance to test this on your fork, or in other ways?

I contemplated the possibility of testing it on my fork, but then realized it would not be too trivial and moved on to other stuff

so in other words: this solution is untested

PR #355, opened 2026-03-21
make `mining_device` into submodule of `integration_tests_sv2` crate

### LINKED ISSUE #354: `mining_device` is blocking the publish of `integration_tests_sv2` to crates.io
we're currently unable to publish `integration_tests_sv2` to crates.io: https://github.com/stratum-mining/sv2-apps/actions/runs/23310666268/job/67796813420

I can see a few options ahead:
1. publish `mining_device` to crates.io
2. integrate `mining_device` into `integration_tests_sv2` as a submodule

Peronally speaking, I lean towards 2, for a few reasons:
- we haven't maintained `mining_device` in a long time. The crate is full of technical debt and in my personal opinion, we shouldn't prioritize refactoring it.
- publishing to crates.io is irreversible, which makes it an undesirable strategy for workarounds.
- [Mujina already supports CPU mining](https://github.com/256foundation/mujina/blob/main/docs/cpu-mining.md), and Sv2 support is on its way... it would be great if `integration_tests_sv2` could use mujina as a dependency and we do all integration tests with it.

plebhash:
this is one potential (and opinionated) approach to close #354

one alternative is to start publishing `mining_device` to crates.io


xyephy (changes requested):
`cargo publish --dry-run` still fails after this change:

```
error: all dependencies must have a version specified when publishing.
dependency `key-utils` does not specify a version
```

These git deps need version specifiers:

```toml
key-utils = { version = "1.2.0", git = "https://github.com/stratum-mining/stratum", rev = "v1.5.0" }
stratum-common = { version = "4.0.1", git = "https://github.com/stratum-mining/stratum", rev = "v1.5.0", features = ["with_network_helpers"] }
```

Both versions match the v1.5.0 tag and exist on crates.io.

plebhash:
> `cargo publish --dry-run` still fails after this change:
> 
> ```
> error: all dependencies must have a version specified when publishing.
> dependency `key-utils` does not specify a version
> ```
> 
> These git deps need version specifiers:
> 
> ```toml
> key-utils = { version = "1.2.0", git = "https://github.com/stratum-mining/stratum", rev = "v1.5.0" }
> stratum-common = { version = "4.0.1", git = "https://github.com/stratum-mining/stratum", rev = "v1.5.0", features = ["with_network_helpers"] }
> ```
> 
> Both versions match the v1.5.0 tag and exist on crates.io.

thanks, should be addressed

PR #368, opened 2026-03-24
Add shared and sharedMap to stratum apps

### LINKED ISSUE #255: need to add a `RwLock` variant to `stratum_apps::custom_mutex::Mutex`
on #83 I proposed to completely remove `stratum_apps::custom_mutex::Mutex`, but got some pushback together with the acknowledgement that a `RwLock` variant to `Mutex` could be a good improvement

so this issue aims to replace #83

bit-aloo:
This PR is based on the discussion at https://github.com/stratum-mining/sv2-apps/discussions/340 regarding the addition of wrapper types around our synchronous primitives to provide a safer and more robust abstraction over leaky async blocking primitives.

To see the application of the introduced wrapper types, refer to https://github.com/stratum-mining/sv2-apps/pull/304, where we are migrating Pool to use DashMap via these wrappers.

This separate PR has been opened to reduce the review surface of the Pool PR.

closes: https://github.com/stratum-mining/sv2-apps/issues/255

### DIFF
--- a/stratum-apps/src/lib.rs
+++ b/stratum-apps/src/lib.rs
@@ -90,5 +90,8 @@ pub mod payout;
 #[cfg(feature = "fallback-coordinator")]
 pub mod fallback_coordinator;
 
+/// Share synchronous primitives
+pub mod sync;
+
 /// Shared async channel cleanup helpers.
 pub mod channel_utils;
--- a/stratum-apps/src/sync.rs
+++ b/stratum-apps/src/sync.rs
@@ -1,0 +1,304 @@
+use std::{
+    hash::Hash,
+    sync::{Arc, Mutex, MutexGuard, PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard},
+};
+
+use dashmap::DashMap;
+
+type SharedLockResult<'a, T, R> = Result<R, PoisonError<MutexGuard<'a, T>>>;
+type SharedRwReadResult<'a, T, R> = Result<R, PoisonError<RwLockReadGuard<'a, T>>>;
+type SharedRwWriteResult<'a, T, R> = Result<R, PoisonError<RwLockWriteGuard<'a, T>>>;
+
+/// Thread-safe shared mutable value using `Mutex` for exclusive access.
+#[derive(Debug)]
+pub struct SharedLock<T>(Arc<Mutex<T>>);
+
+impl<T> Clone for SharedLock<T> {
+    fn clone(&self) -> Self {
+        SharedLock(Arc::clone(&self.0))
+    }
+}
+
+impl<T> SharedLock<T> {
+    /// Create a new shared value.
+    pub fn new(v: T) -> Self {
+        SharedLock(Arc::new(Mutex::new(v)))
+    }
+
+    /// Execute a closure with mutable access to the inner value.
+    ///
+    /// Caution: `f` runs while the mutex is held. Avoid re-entering this
+    /// `SharedLock` from inside the closure.
+    pub fn with<F, R>(&self, f: F) -> SharedLockResult<'_, T, R>
+    where
+        F: FnOnce(&mut T) -> R,
+    {
+        let mut lock = self.0.lock()?;
+        Ok(f(&mut *lock))
+    }
+
+    /// Get a cloned snapshot of the value.
+    pub fn get(&self) -> SharedLockResult<'_, T, T>
+    where
+        T: Clone,
+    {
+        self.with(|v| v.clone())
+    }
+
+    /// Replace the inner value.
+    pub fn set(&self, value: T) -> SharedLockResult<'_, T, ()> {
+        self.with(|v| *v = value)?;
+        Ok(())
+    }
+}
+
+/// Thread-safe shared value using `RwLock` for concurrent reads and exclusive writes.
+#[derive(Debug)]
+pub struct SharedRw<T>(Arc<RwLock<T>>);
+
+impl<T> Clone for SharedRw<T> {
+    fn clone(&self) -> Self {
+        SharedRw(Arc::clone(&self.0))
+    }
+}
+
+impl<T> SharedRw<T> {
+    /// Create a new shared value.
+    pub fn new(v: T) -> Self {
+        SharedRw(Arc::new(RwLock::new(v)))
+    }
+
+    /// Execute a closure with read-only access.
+    ///
+    /// Caution: `f` runs while the read lock is held. Avoid re-entering this
+    /// `SharedRw` from inside the closure, especially for writes.
+    pub fn read<F, R>(&self, f: F) -> SharedRwReadResult<'_, T, R>
+    where
+        F: FnOnce(&T) -> R,
+    {
+        let guard = self.0.read()?;
+        Ok(f(&*guard))
+    }
+
+    /// Execute a closure with mutable access.
+    ///
+    /// Caution: `f` runs while the write lock is held. Avoid re-entering this
+    /// `SharedRw` from inside the closure.
+    pub fn write<F, R>(&self, f: F) -> SharedRwWriteResult<'_, T, R>
+    where
+        F: FnOnce(&mut T) -> R,
+    {
+        let mut guard = self.0.write()?;
+        Ok(f(&mut *guard))
+    }
+
+    /// Get a cloned snapshot of the value.
+    pub fn get(&self) -> SharedRwReadResult<'_, T, T>
+    where
+        T: Clone,
+    {
+        self.read(|v| v.clone())
+    }
+
+    /// Replace the inner value.
+    pub fn set(&self, value: T) -> SharedRwWriteResult<'_, T, ()> {
+        self.write(|v| *v = value)?;
+        Ok(())
+    }
+}
+
+/// Concurrent map wrapper over `DashMap` providing ergonomic scoped access.
+pub struct SharedMap<K: Eq + Clone + Hash, V>(Arc<DashMap<K, V>>);
+
+impl<K: Eq + Clone + Hash, V> Clone for SharedMap<K, V> {
+    fn clone(&self) -> Self {
+        SharedMap(Arc::clone(&self.0))
+    }
+}
+
+impl<K: Eq + Hash + Clone, V> SharedMap<K, V> {
+    /// Create a new concurrent map.
+    pub fn new() -> Self {
+        SharedMap(Arc::new(DashMap::new()))
+    }
+
+    /// Read a value for a key using a closure.
+    ///
+    /// Caution: `f` runs while the entry guard is held. Avoid re-entering this
+    /// `SharedMap` from inside the closure.
+    pub fn with<F, R>(&self, key: &K, f: F) -> Option<R>
+    where
+        F: FnOnce(&V) -> R,
+    {
+        let guard = self.0.get(key)?;
+        let result = f(guard.value());
+        drop(guard);
+        Some(result)
+    }
+
+    /// Mutate a value for a key using a closure.
+    ///
+    /// Caution: `f` runs while the entry guard is held. Avoid re-entering this
+    /// `SharedMap` from inside the closure.
+    pub fn with_mut<F, R>(&self, key: &K, f: F) -> Option<R>
+    where
+        F: FnOnce(&mut V) -> R,
+    {
+        let mut guard = self.0.get_mut(key)?;
+        let result = f(guard.value_mut());
+        Some(result)
+    }
+
+    /// Iterate over all entries immutably.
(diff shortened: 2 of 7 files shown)


stratum-apps/src/shared.rs
```diff
@@ -0,0 +1,258 @@
...
+    pub fn new(v: T) -> Self {
+        Shared(Arc::new(Mutex::new(v)))
+    }
+
+    /// Execute a closure with mutable access to the inner value.
+    pub fn with<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut T) -> R,
+    {
+        let mut lock = self.0.lock().unwrap();
```
plebhash: is this `unwrap` an intentional design decision?

if yes, what's the rationale behind it?

(same question for other `unwrap`s on this file)

bit-aloo: The only way this closure would ever fail on `unwrap` is if the lock gets poisoned. If lock poisoning happens, we either want to recover from it by moving from a corrupt state back to a stable state, or panic and let the application terminate.

I want to take the second route just panic and terminate the application. That was the rationale behind this.

I originally wanted to use `parking_lot` because the notion of poisoning does not exist in its `Mutex` or `RwLock` implementations. With poisoning, you effectively give the system a chance to recover, meaning we need logic to unpoison the lock and restore a stable state, which is kind of a headache. In a non-poisoning setup, if something panics while holding the lock, the panic still happens, but the lock itself is not poisoned, so the lock remains usable afterward, but we still need to ensure our critical sections do not leave the protected state inconsistent if they panic.

At the same time, if we just `unwrap` and something goes wrong, we might only poison a subset of tasks in an async context while the application itself continues running like we saw in the Pool VPS issue, where the pool kept running but stopped accepting new connections. So I guess, in the case of `std` locks, we can propagate the poison error and ensure it carries a shutdown action with it. Though the ergonomics would be a bit weird, with chains of `map_err`s everywhere, but we can improve that later.

Another thing we can do is establish a convention where we do not allow any computation inside the closure that could potentially panic. We obviously cannot enforce this at compile time, but we can expect users to rely on dedicated get/set APIs whenever they need to perform more complex operations.

stratum-apps/src/shared.rs
```diff
@@ -0,0 +1,258 @@
...
+    where
+        F: FnOnce(&mut V) -> R,
+    {
+        let mut guard = self.0.get_mut(key)?;
+        let result = f(guard.value_mut());
+        Some(result)
+    }
+
+    /// Iterate over all entries immutably.
+    pub fn for_each<F, Ret>(&self, mut f: F)
```
plebhash, line 156: from `gpt-5.5`:

> this exposes for_each, for_each_mut, and try_for_each_mut callbacks while DashMap entry guards are held. That still leaves callers able to re-enter the same SharedMap from inside the closure, for example calling remove, insert, retain, or nested with_mut, which DashMap documents as a possible deadlock when any map reference is held. Since this module is meant to make shared synchronous primitives safer, the iteration API should either avoid exposing callbacks under DashMap guards or clearly constrain/rework mutation patterns before downstream code migrates to it.

stratum-apps/src/shared.rs
bit-aloo: Added a commit: 1a2a651, to remove unwraps on std locks

stratum-apps/src/shared.rs
bit-aloo: Removed for_each and all variants that could lead to re-entrancy issues. I still kept `with_mut`, as I want updation to happen atomically.

stratum-apps/src/shared.rs
```diff
@@ -0,0 +1,254 @@
+use std::{
+    hash::Hash,
+    sync::{Arc, Mutex, MutexGuard, PoisonError, RwLock, RwLockReadGuard, RwLockWriteGuard},
+};
+
+use dashmap::DashMap;
+
+type SharedResult<'a, T, R> = Result<R, PoisonError<MutexGuard<'a, T>>>;
```
plebhash:
```suggestion
type SharedLockResult<'a, T, R> = Result<R, PoisonError<MutexGuard<'a, T>>>;
```

stratum-apps/src/shared.rs
```diff
@@ -0,0 +1,254 @@
...
+type SharedReadResult<'a, T, R> = Result<R, PoisonError<RwLockReadGuard<'a, T>>>;
```
plebhash:
```suggestion
type SharedRwReadResult<'a, T, R> = Result<R, PoisonError<RwLockReadGuard<'a, T>>>;
```

stratum-apps/src/shared.rs
```diff
@@ -0,0 +1,254 @@
...
+type SharedWriteResult<'a, T, R> = Result<R, PoisonError<RwLockWriteGuard<'a, T>>>;
```
plebhash:
```suggestion
type SharedRwWriteResult<'a, T, R> = Result<R, PoisonError<RwLockWriteGuard<'a, T>>>;
```

stratum-apps/src/shared.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Thread-safe shared mutable value using `Mutex` for exclusive access.
+#[derive(Debug)]
+pub struct Shared<T>(Arc<Mutex<T>>);
```
plebhash:
```suggestion
pub struct SharedLock<T>(Arc<Mutex<T>>);
```

stratum-apps/src/shared.rs
plebhash: ok took some time to digest the review comment above

so IIUC the danger is in the generic execution of the `f: F` callback/closure while the entry lock is held

in case the user inadvertently called e.g. `self.remove()` or `self.insert()` from inside `f`, there would be room for deadlock

---

the mitigation we're taking here is simply getting rid of `for_each`, `for_each_mut`, and `try_for_each_mut`, in an attempt to reduce the surface for the so-called re-entrancy deadlocks

---

unfortunately I'm afraid to say this mitigation strategy is not completely eliminating the problem... more specifically, the following is not 100% true:

> Removed all variants that could lead to re-entrancy issues.

`with` and `with_mut` suffer from the same class of callback/closure vulnerability:

in case the user inadvertently called e.g. `self.remove()` or `self.insert()` from inside `f`, there would be room for deadlock

so we should at least write a very explicit warning on Rust Docs above `with` and `with_mut` to let the user know that trying to mutate `self` inside those closures is a bad idea

with this approach, we could potentially re-introduce `for_each` + `for_each_mut` + `try_for_each_mut` with similar warning docs

(or keep them out for simplicity, no hard opinions)

stratum-apps/src/shared.rs
plebhash: I think this has more semantic clarity

`SharedResult` makes me ask:

> is this a `Result` that happens to be a shared resource, or is this a `Result` about some specific shared resource?

if we rename `Shared` into `SharedLock`, we make it explicitly clear what the underlying shared resource is, and ad the same time `SharedLockResult` feels less semantically ambiguous

stratum-apps/src/shared.rs
plebhash: `SharedReadResult` makes me ask:

> is this the `Result` of reading `Shared`? or the `Result` of reading `SharedRw`?

`SharedRwReadResult` feels less semantically ambiguous

especially if we also rename `Shared` into `SharedLock`, which establishes a clear semantic contrast against `SharedRw`

stratum-apps/src/shared.rs
plebhash: `SharedRwWriteResult` makes me ask:

> is this the `Result` of writing `Shared`? or the `Result` of writing `SharedRw`?

`SharedRwWriteResult` feels less semantically ambiguous

especially if we also rename `Shared` into `SharedLock`, which establishes a clear semantic contrast against `SharedRw`

stratum-apps/src/shared.rs
plebhash: `SharedLock` establishes a clear semantic contrast against `SharedRw`

plebhash: `SharedMutex` is another viable option

stratum-apps/src/shared.rs
plebhash: `SharedMutex` is another viable option

stratum-apps/src/shared.rs
bit-aloo: TBH, I would prefer to revert the clone-based solution and instead add a comment mentioning the caution. It gives us a much cleaner surface.

With the clone-based approach, the issue is that changes won’t be atomic: we first clone, then make the change, and finally reinsert the value, which makes it possible to overwrite updates from some other concurrent change. I’m okay with re-entrancy as long as we are careful about it. Let me know what you think

plebhash: > to revert the clone-based solution

to be honest I hadn't noticed that this had been introduced here (I only remember discussing it on the call)

sorry for missing that on my review

and I agree, it's not good to break atomicity, we shouldn't clone inside those methods

also agree that docs should be enough to warn users about deadlocks-due-to-reentrancy footguns on `SharedMap`

about  `for_each` + `for_each_mut` + `try_for_each_mut`, I'd lean towards keeping out of the scope:
- their added complexity make things harder to reason about
- we're not sure we'll really need iterators
- if we need, it can be achieved with some inline boilerplate around `with` and `with_mut`

bit-aloo: If you look at tproxy, where we already use DashMap, we have a couple of instances where we rely on iterators. That was the motivation behind adding them. If you think they are not really that complex, they are very similar to the with or with_mut methods defined on standard locks.

Regarding the suggestion to use with or with_mut as accessors: these are only helpful for a hashmap when we already know the key. Otherwise, the workflow would involve first retrieving all the keys (which requires cloning them), and then iterating over those keys to process their associated values, which increases the overall complexity.

bit-aloo: I have reverted the cloned changes and add a word of caution for re-entrancy issue

plebhash (approved):
@Shourya742 can we squash all commits into 1?

bit-aloo:
> @Shourya742 can we squash all commits into 1?

Done. Once this is merged, I will replace dashmap in tproxy with sharedMap and replace custom_mutex with its shared variant.

stratum-apps/src/lib.rs
```diff
@@ -90,5 +90,8 @@ pub mod payout;
 #[cfg(feature = "fallback-coordinator")]
 pub mod fallback_coordinator;
 
+/// Share synchronous primitives
+pub mod shared;
```
GitGab19: Can this module be named in a clearer way?

stratum-apps/src/sync.rs
```diff
@@ -0,0 +1,290 @@
...
+        let mut guard = self.0.get_mut(key)?;
+        let result = f(guard.value_mut());
+        Some(result)
+    }
+
+    /// Iterate over all entries immutably.
+    ///
+    /// Caution: `f` runs while an iterator entry guard is held. Avoid
+    /// re-entering this `SharedMap` from inside the closure.
+    pub fn for_each<F, Ret>(&self, mut f: F)
```
GitGab19, line 156: Shouldn't we have a `try_for_each` as we do for `try_for_each_mut`?

bit-aloo: I have not really seen a use case of this in our current dashmap tproxy implementation. We can add this if we see the usecase in our other crates.

stratum-apps/src/lib.rs
bit-aloo: All ears...

stratum-apps/src/sync.rs
GitGab19: It could be useful in cases where the internal callback can fail and you want to stop iteration and return the error?

stratum-apps/src/lib.rs
GitGab19: Maybe `sync`?

stratum-apps/src/sync.rs
bit-aloo: For sure, I am mentioning we don't see such scenario in our tproxy dashmap. For symmetry, I will add it. ;)

stratum-apps/src/lib.rs
bit-aloo: took the suggestion, named the module `sync`.

stratum-apps/src/sync.rs
bit-aloo: Should be there now.

GitGab19 (approved):

Issue #382, opened 2026-03-27
bitaxes failing to connect to `translator_sv2`


plebhash:
during community testing after v0.3.1, which means the deployment happened via
```
➜  ~ docker run --rm \                   
  --name sv2-ui \
  -p 8080:8080 \
  -e HOST_HOME=$HOME \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v sv2-config:/app/data/config \
  stratumv2/sv2-ui:v0.1.0
```

 we're getting some community members reporting that they cannot connect bitaxes to `translator_sv2`

I initially suspected this could be related to the docker deployments, but @lucasbalieiro mentioned he was able to reproduce it without docker

<img width="758" height="540" alt="Image" src="https://github.com/user-attachments/assets/1de3a575-e0e2-474e-aedb-1e356cfa76e1" /> 

<img width="907" height="621" alt="Image" src="https://github.com/user-attachments/assets/3c2c92e0-52bc-47f6-9ffa-c8aa2cfef173" />

<img width="1007" height="623" alt="Image" src="https://github.com/user-attachments/assets/fc0b47fd-9a5a-4f57-91fc-19493b4614e2" />

plebhash:
potentially related to https://github.com/stratum-mining/sv2-apps/issues/383

lucasbalieiro:
Another weird behavior:

If try to connect the bitaxe to tproxy and it loops in the: `SV1 handshake not complete, caching mining.notify`

If you CTRL-C the tproxy, and keep looking at the bitaxe logs.

Wait for the second retry on the primary pool (which will be your killed tproxy).

After the second retry, start your tproxy again.
The time my tproxy connects upstreams and starts listening is the time that bitaxe connects quickly to the ckpool.solo. Then the bitaxe somehow understands that primary pool got back and connects again to your tproxy. But now, is able to send shares, receives jobs.

Also, another scenario that makes the bitaxe got back working again is making a quick switch between the different tproxys: 
example: i was quickly switching between `--release` bin -> docker or to normal build (without `--release`) and eventually the bitaxe is able to receive jobs and sending shares.

Maybe, my comment above about only happening in `--release` was not the problem. But the reconnection with the tproxy eventually left the bitaxe in a state that it was able to receive jobs and send shares.


in the logs below you can see that the bitaxe fallsback to cksolo, and right after I start my tproxy. The bitaxe detects the primary pool is live again and reconnects, but now it is receiving work:

```text
[...]

PR #387, opened 2026-03-28
Fix translator handshake flow

Shourya742:
closes: #382

### DIFF
--- a/miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
@@ -584,10 +584,6 @@ impl Sv1Server {
                                 downstream_id
                             );
 
-                            // Set flag to indicate we're processing queued responses
-                            downstream
-                                .processing_queued_sv1_handshake_responses
-                                .store(true, Ordering::SeqCst);
                             let downstream_sv1_sender = downstream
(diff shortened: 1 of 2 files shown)


miner-apps/translator/src/lib/sv1/sv1_server/sv1_server.rs
```diff
@@ -370,13 +370,12 @@ impl Sv1Server {
...
-                            if let Err(e) = downstream.handle_sv1_handshake_completion().await {
-                                error!("Down: Failed to handle handshake completion: {:?}", e);
-                                return Err(TproxyError::disconnect(e, downstream_id));
-                            }
+                    if matches!(&downstream_message, json_rpc::Message::StandardRequest(r) if r.method == "mining.authorize")
+                    {
+                        info!("Down: Handling mining.authorize after handshake completion");
+                        if let Err(e) = downstream.handle_sv1_handshake_completion().await {
+                            error!("Down: Failed to handle handshake completion: {:?}", e);
+                            return Err(TproxyError::disconnect(e, downstream_id));
```
plebhash: please let's stay focused and avoid introducing cosmetic changes into patch PRs

not only this is unnecessary and orthogonal to #382, now this macro is far less obvious to read and it takes review effort to evaluate why this is actually needed

PR #398, opened 2026-03-31
add config option to start JDC in solo mining mode

### LINKED ISSUE #371: need to add config option to start JDC in solo mining mode
Currently there is no way to explicitly start JDC in SOLO MINING mode without first going through the upstream connection attempt loop.

We probably need to add a config parameter for that.

By starting JDC in SOLO MINING mode, users can do a fully sovereign solo mining with their Bitcoin Core node, without having to connect to any Pool/JDS.

We've already planned to enable this feature on the `sv2-ui` once it lands here:

Image

Tracker on `sv2-ui`: https://github.com/stratum-mining/sv2-ui/issues/36

xyephy:
When `upstreams` is omitted from the TOML config, JDC starts directly in solo mining mode. No upstream connection attempts, no retry delay. Templates come from the Template Provider and blocks are submitted via SubmitSolution.

Unblocks stratum-mining/sv2-ui#36

Closes #371

### DIFF
--- a/miner-apps/jd-client/src/lib/config.rs
+++ b/miner-apps/jd-client/src/lib/config.rs
@@ -45,7 +45,7 @@ pub struct JobDeclaratorClientConfig {
     shares_per_minute: SharesPerMinute,
     /// share batch size
     share_batch_size: SharesBatchSize,
-    /// JDC mode: FullTemplate or CoinbaseOnly
+    /// JDC mode: FullTemplate, CoinbaseOnly, or SoloMining
     #[serde(deserialize_with = "deserialize_jdc_mode", default)]
     pub mode: ConfigJDCMode,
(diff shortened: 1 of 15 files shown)


plebhash:
let's aim to get this merged for v0.3.3

miner-apps/jd-client/src/lib/config.rs
```diff
@@ -31,6 +31,7 @@ pub struct JobDeclaratorClientConfig {
     template_provider_type: TemplateProviderType,
     /// A list of upstream Job Declarator Servers (JDS) that this JDC can connect to.
     /// JDC can fallover between these upstreams.
+    #[serde(default)]
     upstreams: Vec<Upstream>,
```
Shourya742: this will make config file to parse even without upstream sub field. I would keep it, even though we don't add any upstream.

Issue #234, opened 2026-02-03
need to remove main status loop


plebhash:
with #215 came the realization that the status loop spawned on `{PoolSv2,TranslatorSv2JobDeclaratorClient}::start` is somewhat redundant

it introduces a queue of status events into a unified pipeline, which arguably causes unnecessary delays on different subsystems behavior, given that everything needs to be bottleneck'd into this unified loop

in summary, whenever some edge case event happens somewhere, it should trigger the response behavior directly into the affected areas, without a centralizing pipeline on this main status loop

PR #426, opened 2026-04-13
Remove status main loop from Pool

Shourya742:
Part of: #234, for now only migrates Pool


pool-apps/pool/src/lib/downstream/mod.rs
```diff
@@ -181,7 +175,20 @@ impl Downstream {
...
+                    self.downstream_connection_token.cancel();
+                }
+                Action::Log => {
+                    warn!(
+                        "Log-only error from downstream {}: {:?}",
+                        self.downstream_id, e.kind
+                    );
+                }
+            }
             return;
```
plebhash, line 193:
```suggestion
            }
            remove_downstream(self.downstream_id);
            return;
```

otherwise, if the client is rejected right after handshake (bad frame, or bad `SetupConnection`), it will leave behind a zombie entry in the HashMap

bit-aloo: Ah!! thanks. Let me update it.

bit-aloo: should be there now. Thanks

plebhash:
can we clean up commit history?

we're doing things (e.g.: f8d22439dc6d60f84c361d347845fe5920466a5a), regretting and then doing something else later (e.g.: 0e6476c616b8c40e6dd69a8109666ca7ffa71051)

Shourya742:
> can we clean up commit history?
> 
> we're doing things (e.g.: [f8d2243](https://github.com/stratum-mining/sv2-apps/commit/f8d22439dc6d60f84c361d347845fe5920466a5a)), regretting and then doing something else later (e.g.: [0e6476c](https://github.com/stratum-mining/sv2-apps/commit/0e6476c616b8c40e6dd69a8109666ca7ffa71051))

Done. Though I regret the match action as well ;>, couldn't come up with better atm.

plebhash (approved):
tACK

Issue #430, opened 2026-04-14
TranslatorSv2::start() panics on second call due to process-wide OnceLock statics (breaks sequential integration tests)


gimballock:
## Problem

`TPROXY_MODE` and `VARDIFF_ENABLED` are declared as process-wide `static OnceLock<_>` in `miner-apps/translator/src/lib/mod.rs`:

```rust
static TPROXY_MODE: OnceLock<TproxyMode> = OnceLock::new();
static VARDIFF_ENABLED: OnceLock<bool> = OnceLock::new();
```

`TranslatorSv2::start()` initializes them unconditionally on every call:

```rust
TPROXY_MODE
    .set(self.config.aggregate_channels.into())
    .expect("TPROXY_MODE initialized more than once");
VARDIFF_ENABLED
    .set(self.config.downstream_difficulty_config.enable_vardiff)
    .expect("VARDIFF_ENABLED initialized more than once");
```

`nextest.toml` sets `test-threads = 1`, so all integration tests run sequentially in the **same process**. The first test that calls `start()` on a `TranslatorSv2` sets the `OnceLock`s successfully. Every subsequent test that creates a new `TranslatorSv2` and calls `start()` causes `OnceLock::set()` to return `Err` — the `.expect()` panics. Since `start()` is invoked inside `tokio::spawn`, the panic is silently swallowed, the translator exits immediately, and the test framework only sees a downstream symptom (e.g. "Connection refused" on the monitoring or SV1 port).

## Observed symptom

When running the monitoring integration tests with `--test-threads=1`, the first tProxy test passes and the remaining 9 fail with "Connection refused" on the monitoring endpoint. There are 10 tProxy tests total, producing a consistent 10/9 pass/fail split. Confirmed reproducible on `3773dd5` (upstream main before PR #373).

## Fix

`TPROXY_MODE` and `VARDIFF_ENABLED` should not be process-wide singletons. Options:

1. **Make them instance fields on `TranslatorSv2`** (preferred) — remove the statics, store the values in the struct, and pass them through to callers.
2. **Use `get_or_init` everywhere instead of `set` + `expect`** — this silently ignores the second initialization, which is safe if all instances are configured identically, but masks misconfiguration.

Option 1 is the correct fix as it allows multiple `TranslatorSv2` instances with different configs to coexist in the same process (required for integration tests and any future multi-instance deployment).

Shourya742:
Thats a bummer, IT should have flagged earlier. Let me check

PR #431, opened 2026-04-15
Fix/430 Refactor: Remove OnceLock statics from TranslatorSv2 (#430)

gimballock:
Replace process-wide static OnceLock variables (TPROXY_MODE, VARDIFF_ENABLED) 
with instance fields to fix panics during sequential integration tests.

Changes:
- Add tproxy_mode and report_hashrate fields to ChannelManager and Sv1Server
- Add is_aggregated() and is_non_aggregated() instance methods
- Update all call sites to use self.is_aggregated() instead of global functions
- Remove static OnceLocks and global helper functions from mod.rs
- Add integration test verifying multiple TranslatorSv2 instances can run
- Fix test naming mismatch in test_handle_set_target_without_vardiff_non_aggregated

This enables multiple TranslatorSv2 instances to be created sequentially
without panicking due to OnceLock re-initialization.


Shourya742:
superseded here: https://github.com/stratum-mining/sv2-apps/pull/432

PR #373, opened 2026-03-24
Add integration and unit tests for monitoring JSON API endpoints

gimballock:
Closes #329

Exercise every JSON REST API endpoint against live SV2 topologies and strengthen unit-test coverage for edge cases.

Integration tests (monitoring_integration.rs):
- 15 new tests covering /, /api/v1/global, /api/v1/server, /api/v1/server/channels, /api/v1/clients, /api/v1/clients/{id}, /api/v1/clients/{id}/channels, /api/v1/sv1/clients, and /api/v1/sv1/clients/{id} for both Pool and tProxy topologies
- Topology setup helpers (PoolWithSv2Miner, TproxyWithSv1Miner) eliminate duplicated boilerplate across tests
- 404 endpoints assert both HTTP status code and JSON error body via new assert_api_not_found helper

Assertion helpers (prometheus_metrics_assertions.rs):
- fetch_api_json: parse JSON API response
- fetch_api_with_status: return (status_code, json) without panicking on non-2xx, enabling 404 testing
- poll_until_api_field_gte: poll JSON endpoint until a field reaches a threshold (analogous to poll_until_metric_gte for Prometheus)
- assert_api_root, assert_api_global: reusable structure checks
- assert_api_not_found: verify HTTP 404 + error field
- POLL_TIMEOUT constant to reduce repetition

HTTP helper (utils.rs):
- make_get_request_with_status: returns (status, body) without panicking on 4xx responses, unlike make_get_request

Unit tests (http_server.rs):
- 11 new tests for pagination boundaries (limit=0, offset beyond total, limit exceeding MAX_LIMIT), missing data sources returning 404, invalid/non-existent client IDs, SV1 data in global endpoint, and Prometheus metrics with no sources

Dependencies:
- Add serde_json to integration-tests for JSON parsing

PR #432, opened 2026-04-15
Remove statics from tproxy and jdc

### LINKED ISSUE #430: TranslatorSv2::start() panics on second call due to process-wide OnceLock statics (breaks sequential integration tests)

Shourya742:
This PR removes the use of static variables that were previously used to specify the operating mode (in JDC: `FullTemplate` / `CoinbaseOnly` / `SoloMining`, and in Tproxy: `Aggregated` / `NonAggregated`). While static approach works fine in production where a single binary simulates the process, it does not work well in integration tests (IT), where these components are used as libraries. In such cases, the test harness is built once, and static global variables are placed in the binary’s data section, remaining valid for the entire runtime. However, for testing, we need the ability to change this value dynamically in accordance to start method invocation. Although it is technically possible to mutate these values (as done in JDC), doing so can introduce non-deterministic behavior. When multiple such tasks run in parallel, they may modify the shared state concurrently without proper coordination, leading to unpredictable results.

This PR updates both Tproxy and JDC to eliminate statics and instead initialize the mode locally within the `start` method.

closes: #430

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
@@ -34,7 +34,6 @@ use crate::{
         ChannelManager, ChannelManagerChannel, SharesOrderedByDiff, SOLO_FULL_EXTRANONCE_SIZE,
     },
     error::{self, JDCError, JDCErrorKind},
-    jd_mode::{get_jd_mode, JdMode},
     utils::{add_share_to_cache, create_close_channel_msg},
 };
 
@@ -121,14 +120,12 @@ impl RouteMessageTo<'_> {
                 }
             }
             RouteMessageTo::Upstream(message) => {
-                if get_jd_mode() != JdMode::SoloMining {
-                    let message_static = message.into_static();
-                    let sv2_frame: Sv2Frame = AnyMessage::Mining(message_static).try_into()?;
-                    channel_manager_channel
-                        .upstream_sender
-                        .send(sv2_frame)
-                        .await?;
-                }
+                let message_static = message.into_static();
+                let sv2_frame: Sv2Frame = AnyMessage::Mining(message_static).try_into()?;
+                channel_manager_channel
+                    .upstream_sender
+                    .send(sv2_frame)
+                    .await?;
             }
             RouteMessageTo::JobDeclarator(message) => {
                 channel_manager_channel
--- a/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
@@ -99,8 +99,8 @@ impl RouteMessageTo<'_> {
     /// The routing is handled as follows:
     /// - [`RouteMessageTo::Downstream`] → Sends the mining message to the specified downstream
     ///   client.
-    /// - [`RouteMessageTo::Upstream`] → Sends the mining message upstream, unless in
-    ///   [`JdMode::SoloMining`].
+    /// - [`RouteMessageTo::Upstream`] → Forwards mining message upstream. In solo mode,
+    ///   upstream-directed messages should not be produced.
     /// - [`RouteMessageTo::JobDeclarator`] → Sends the job declaration message to the JDS.
     /// - [`RouteMessageTo::TemplateProvider`] → Sends the template distribution message to the
     ///   template provider.
--- a/miner-apps/jd-client/src/lib/jd_mode.rs
+++ b/miner-apps/jd-client/src/lib/jd_mode.rs
@@ -3,59 +3,77 @@
 //! This module defines different operating modes for the Job Declarator
 //! and provides atomic accessors for setting and retrieving the current mode.
 //!
-//! Modes are stored in a global [`AtomicU8`] to allow safe concurrent access
+//! Mode state is stored in `JDMode::inner` (`Arc<AtomicU8>`) to allow safe concurrent access
 //! across threads.
-use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::{
+    atomic::{AtomicU8, Ordering},
+    Arc,
+};
 
-/// Operating modes for the Job Declarator.
-#[repr(u8)]
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum JdMode {
-    /// Runs in Coinbase only mode.
-    CoinbaseOnly = 0,
-    /// Runs in Full template mode,
-    FullTemplate = 1,
-    /// Runs in solo mining mode,
-    SoloMining = 2,
+use crate::config::ConfigJDCMode;
+
+#[derive(Clone, Debug)]
+pub struct JDMode {
+    inner: Arc<AtomicU8>,
+    // Currently, how JDC works is the mode in config
+    // only gets activated once an upstream connection
+    // is made.
+    config_mode: ConfigJDCMode,
 }
 
-impl From<u8> for JdMode {
-    fn from(val: u8) -> Self {
-        match val {
-            0 => JdMode::CoinbaseOnly,
-            1 => JdMode::FullTemplate,
-            2 => JdMode::SoloMining,
-            _ => JdMode::SoloMining,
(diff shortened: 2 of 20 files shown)


Shourya742:
I think Gimballock's AI explained the issue here: <https://github.com/stratum-mining/sv2-apps/pull/373#issuecomment-4247996017>

miner-apps/jd-client/src/lib/channel_manager/downstream_message_handler.rs
plebhash: 1cd54bb32b4819c816f522000c0f9d84f415a25d makes this doc outdated

plebhash: for some reason my diff suggestion doesn't render properly on github UI, but here's my attempt:

```diff
-    /// - [`RouteMessageTo::Upstream`] → Sends the mining message upstream, unless in
-    ///   [`JdMode::SoloMining`].
+    /// - [`RouteMessageTo::Upstream`] → Forwards mining message upstream. In solo mode,
+    ///   upstream-directed messages should not be produced.
```

ideally this should be squashed into 1cd54bb32b4819c816f522000c0f9d84f415a25d

miner-apps/jd-client/src/lib/jd_mode.rs
```diff
@@ -5,57 +5,75 @@
 //!
 //! Modes are stored in a global [`AtomicU8`] to allow safe concurrent access
 //! across threads.
```
plebhash:
```suggestion
//! Mode state is stored in `JDMode::inner` (`Arc<AtomicU8>`) to allow safe concurrent access
//! across threads.
```

plebhash:
a similar integration test to avoid JDC regressions: ea9e7f3ee63fb48152dbebe0b6f3b9b339a9a08f

Shourya742:
> a similar integration test to avoid JDC regressions: [ea9e7f3](https://github.com/stratum-mining/sv2-apps/commit/ea9e7f3ee63fb48152dbebe0b6f3b9b339a9a08f)

Picked, thanks

miner-apps/jd-client/src/lib/jd_mode.rs
```diff
@@ -3,59 +3,77 @@
 //! This module defines different operating modes for the Job Declarator
 //! and provides atomic accessors for setting and retrieving the current mode.
 //!
-//! Modes are stored in a global [`AtomicU8`] to allow safe concurrent access
+//! MMode state is stored in `JDMode::inner` (`Arc<AtomicU8>`) to allow safe concurrent access
```
plebhash:
```suggestion
//! Mode state is stored in `JDMode::inner` (`Arc<AtomicU8>`) to allow safe concurrent access
```

plebhash (approved):

miner-apps/jd-client/src/lib/jd_mode.rs
bit-aloo: Removed

miner-apps/jd-client/src/lib/job_declarator/message_handler.rs
```diff
@@ -32,16 +30,7 @@ impl HandleCommonMessagesFromServerAsync for JobDeclarator {
...
-        // successful handshake.
-        let jd_mode = match self.mode {
-            ConfigJDCMode::CoinbaseOnly => JdMode::CoinbaseOnly,
-            ConfigJDCMode::FullTemplate => JdMode::FullTemplate,
-            ConfigJDCMode::SoloMining => JdMode::SoloMining,
-        };
-
-        set_jd_mode(jd_mode);
-
+        self.mode.activate();
```
GitGab19, line 33: It's still not clear to me why we need to set JD mode only here, and not while bootstrapping JDC.

bit-aloo: because its set to SoloMining initially and only change the mode if JDS connects.

miner-apps/jd-client/src/lib/jd_mode.rs
```diff
@@ -3,59 +3,77 @@
...
-            _ => JdMode::SoloMining,
+    /// This activates mode based on config file once
+    /// upstream connection is made.
+    pub fn activate(&self) {
+        match self.config_mode {
+            ConfigJDCMode::CoinbaseOnly => self.set_coinbase_only(),
+            ConfigJDCMode::FullTemplate => self.set_full_template(),
+            ConfigJDCMode::SoloMining => self.set_solo_mining(),
         }
     }
```
GitGab19, line 40: If this is called only when JDC receives a `SetupConnection.Success` from JDS, why could it lead to setting JDmode to solo mining?

miner-apps/jd-client/src/lib/job_declarator/message_handler.rs
bit-aloo: We will change this behavior soon....

miner-apps/jd-client/src/lib/jd_mode.rs
bit-aloo: its more about exhaustiveness of enum, we would never really set that, it all be dependent on what mode we add config_mode (which is set via config)

miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs
```diff
@@ -120,6 +125,8 @@ impl ChannelManager {
         status_sender: Sender<Status>,
         supported_extensions: Vec<u16>,
         required_extensions: Vec<u16>,
+        tproxy_mode: TproxyMode,
+        #[cfg(feature = "monitoring")] report_hashrate: bool,
```
GitGab19, line 166: why did you call this `report_hashrate` instead of `enable_vardiff`?

bit-aloo: its only used in monitoring, and there we call it report hashrate. Considering the context is related to report hashrate which would only be used in monitoring.

GitGab19 (approved):

Issue #168, opened 2026-01-07
Non critical Error evaluation (JDC)


Shourya742:
We have a few places in JDC where errors are treated as non-critical and silently ignored. With the contextual error mechanism being integrated into JDC, these cases are now explicitly represented, but we are still effectively just logging and ignoring them. This issue tracks a thorough evaluation of those cases and proposes patches where a more appropriate handling strategy can be applied.

plebhash:
while https://github.com/stratum-mining/sv2-apps/pull/132 is laying down some foundational systems for error handling, it's not realistic to review in-detail every single instance where this new system is being used in every single app at once... that was the motivation for creating this issue here, in the spirit of breaking things down in smaller steps

but I would propose the scope here should actually be much broader, also accounting for the re-evaluation of every error case.

if we restricted this task only on re-evaluating log-and-skip, while assuming that https://github.com/stratum-mining/sv2-apps/pull/132 has already correctly taken care of every other scenario, we are leaving too much room for loose ends and blindspots that will most likely come back to bite our asses in the future.

this should be an opportunity for a fine toothcomb review that makes correct usage of the foundations from #132 without leaving anything behind

---

TLDR: IMHO we should re-evaluate **EVERYTHING** that's making usage of #132 (as well as current usage of `unwrap()`, `expect()`, `if let Err`, `if let Some`, etc), not only non-critical or log-and-skip, as the original description suggests

nulllpc:
Hi @Shourya742 @plebhash I'm reviewing error handling for JDC. I have a few observations so far, please let me know if I'm on the right track

1. `JDCError` context is lost, leading to blind retries
The `try_initialize_single` function [strips](https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/jd-client/src/lib/mod.rs#L725) the rich `JDCError` (with `kind` and `action`) down to a simple `JDCErrorKind`. This goes against the goal of the new error system, as it forces [the main retry loop](https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/jd-client/src/lib/mod.rs#L663) to be blind, treating all failures generically. While a separate `cancellation_token` check currently [ensures shutdowns are handled](https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/jd-client/src/lib/upstream/mod.rs#L116), the loop itself can't act on other important actions like `Disconnect` or fail fast on non-retryable errors

[...]

PR #456, opened 2026-04-25
refactor(jd-client): robust error handling for initialization and retry logic

nulllpc:
https://github.com/stratum-mining/sv2-apps/issues/168

This PR attempts to improve error handling for JDC by addressing some low-hanging improvements:
1. Replaces several `expect()` and `unwrap()` calls in `JobDeclaratorClient::start` with proper error propagation using `Result`.
2. Refactors the JDC error handling by making the upstream initialization logic "action-aware."

Changes:
- Updated `start()` signature to return `JDCError<JobDeclarationClient>`.
- Introduced `InitializationError(String)` variant to `JDCErrorKind` to preserve context from formerly panicking calls.
- Updated `main.rs` and integration tests to handle the new `Result` type
- `try_initialize_single` is updated to inspect the `Action` from sub-component errors and return a full `JDCError` with either a `Shutdown` or `Fallback` action.
- The retry loop in `initialize_jd` now checks `e.action`. It immediately propagates fatal `Shutdown` errors up to `start()` while continuing to retry on `Fallback` errors.
- The `Action` enum now derives `PartialEq` and `Eq` to enable this logic.
- Function signatures now consistently use the `JDCResult` type alias.

### DIFF
--- a/integration-tests/lib/mod.rs
+++ b/integration-tests/lib/mod.rs
@@ -274,7 +274,11 @@ pub fn start_jdc(
     );
     let ret = jd_client_sv2::JobDeclaratorClient::new(jd_client_proxy);
     let ret_clone = ret.clone();
-    tokio::spawn(async move { ret_clone.start().await });
+    tokio::spawn(async move {
+        if let Err(e) = ret_clone.start().await {
+            panic!("JDC failed to start: {e}");
+        }
+    });
     (ret, jdc_address, monitoring_address)
 }
 
--- a/miner-apps/jd-client/src/lib/error.rs
+++ b/miner-apps/jd-client/src/lib/error.rs
@@ -58,13 +58,16 @@ pub struct Upstream;
 pub struct Downstream;
 
 #[derive(Debug)]
+pub struct JobDeclaratorClient;
+
+#[derive(Debug)]
 pub struct JDCError<Owner> {
     pub kind: JDCErrorKind,
     pub action: Action,
     _owner: PhantomData<Owner>,
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Action {
     Log,
     Disconnect(DownstreamId),
@@ -84,12 +87,14 @@ impl CanDisconnect for ChannelManager {}
 impl CanFallback for Upstream {}
 impl CanFallback for JobDeclarator {}
 impl CanFallback for ChannelManager {}
+impl CanFallback for JobDeclaratorClient {}
 
 impl CanShutdown for ChannelManager {}
 impl CanShutdown for TemplateProvider {}
 impl CanShutdown for Downstream {}
 impl CanShutdown for Upstream {}
 impl CanShutdown for JobDeclarator {}
+impl CanShutdown for JobDeclaratorClient {}
 
 impl<O> JDCError<O> {
     pub fn log<E: Into<JDCErrorKind>>(kind: E) -> Self {
@@ -256,6 +261,10 @@ pub enum JDCErrorKind {
     InvalidKey,
     /// Upstream not found
     UpstreamNotFound,
+    /// JDC initialization error with details
+    InitializationError(String),
+    /// No upstream is available after all attempts have failed, fallback to solo mining.
+    NoUpstreamAvailable,
 }
 
 impl std::error::Error for JDCErrorKind {}
@@ -395,6 +404,12 @@ impl fmt::Display for JDCErrorKind {
             CouldNotInitiateSystem => write!(f, "Could not initiate subsystem"),
             InvalidKey => write!(f, "Invalid key used during noise handshake"),
             UpstreamNotFound => write!(f, "Upstream not found"),
+            InitializationError(ref err) => {
+                write!(f, "Cannot initialize JD client: {err:?}")
+            }
+            NoUpstreamAvailable => {
+                write!(f, "No upstream is available after all attempts have failed")
+            }
         }
     }
 }
--- a/miner-apps/jd-client/src/lib/mod.rs
+++ b/miner-apps/jd-client/src/lib/mod.rs
@@ -22,7 +22,7 @@ use tracing::{debug, error, info, warn};
 use crate::{
     channel_manager::ChannelManager,
     config::JobDeclaratorClientConfig,
-    error::JDCErrorKind,
+    error::{Action, JDCError, JDCErrorKind, JDCResult},
     jd_mode::JDMode,
     job_declarator::JobDeclarator,
     template_receiver::{
@@ -68,7 +68,7 @@ impl JobDeclaratorClient {
     }
 
     /// Starts the Job Declarator Client (JDC) main loop.
-    pub async fn start(&self) {
+    pub async fn start(&self) -> JDCResult<(), error::JobDeclaratorClient> {
         info!(
             "Job declarator client starting... setting up subsystems, User Identity: {}",
             self.config.user_identity()
@@ -77,17 +77,28 @@ impl JobDeclaratorClient {
         let miner_coinbase_outputs = vec![self.config.get_txout()];
         let mut encoded_outputs = vec![];
         let mode = JDMode::new(self.config.mode);
+        let task_manager = Arc::new(TaskManager::new());
+        let mut bitcoin_core_sv2_join_handle: Option<JoinHandle<()>> = None;
+        let mut bitcoin_core_sv2_cancellation_token: Option<CancellationToken> = None;
 
-        if let Err(e) = miner_coinbase_outputs.consensus_encode(&mut encoded_outputs) {
-            error!(error = ?e, "Invalid coinbase output in config");
+        if miner_coinbase_outputs
+            .consensus_encode(&mut encoded_outputs)
+            .is_err()
+        {
             self.cancellation_token.cancel();
-            self.shutdown_notify.notify_waiters();
-            self.is_alive.store(false, Ordering::Relaxed);
-            return;
+            self.graceful_shutdown(
+                task_manager,
+                bitcoin_core_sv2_join_handle,
+                bitcoin_core_sv2_cancellation_token,
+            )
+            .await;
+
+            return Err(JDCError::shutdown(JDCErrorKind::InitializationError(
+                "Invalid coinbase output in config".to_string(),
+            )));
         }
 
         let mut fallback_coordinator = FallbackCoordinator::new();
-        let task_manager = Arc::new(TaskManager::new());
 
         let (channel_manager_to_upstream_sender, channel_manager_to_upstream_receiver) =
             unbounded();
@@ -123,11 +134,15 @@ impl JobDeclaratorClient {
         {
             Ok(channel_manager) => channel_manager,
             Err(e) => {
-                error!(error = ?e, "Failed to initialize channel manager");
                 self.cancellation_token.cancel();
-                self.shutdown_notify.notify_waiters();
-                self.is_alive.store(false, Ordering::Relaxed);
-                return;
+                self.graceful_shutdown(
+                    task_manager,
+                    bitcoin_core_sv2_join_handle,
+                    bitcoin_core_sv2_cancellation_token,
+                )
+                .await;
+
+                return Err(JDCError::shutdown(e.kind));
             }
         };
 
@@ -191,8 +206,6 @@ impl JobDeclaratorClient {
         }
 
         let initial_channel_manager = channel_manager.clone();
-        let mut bitcoin_core_sv2_join_handle: Option<JoinHandle<()>> = None;
-        let mut bitcoin_core_sv2_cancellation_token: Option<CancellationToken> = None;
 
         match self.config.template_provider_type().clone() {
             TemplateProviderType::Sv2Tp {
@@ -213,9 +226,14 @@ impl JobDeclaratorClient {
                     Err(e) => {
                         error!(error = ?e, "Failed to initialize SV2 template receiver");
                         self.cancellation_token.cancel();
-                        self.shutdown_notify.notify_waiters();
-                        self.is_alive.store(false, Ordering::Relaxed);
-                        return;
+                        self.graceful_shutdown(
+                            task_manager,
+                            bitcoin_core_sv2_join_handle,
+                            bitcoin_core_sv2_cancellation_token,
+                        )
+                        .await;
+
+                        return Err(JDCError::shutdown(e.kind));
                     }
                 };
 
@@ -228,9 +246,14 @@ impl JobDeclaratorClient {
                 {
                     error!(error = ?e, "Failed to start SV2 template receiver");
                     self.cancellation_token.cancel();
-                    self.shutdown_notify.notify_waiters();
-                    self.is_alive.store(false, Ordering::Relaxed);
-                    return;
+                    self.graceful_shutdown(
+                        task_manager,
+                        bitcoin_core_sv2_join_handle,
+                        bitcoin_core_sv2_cancellation_token,
+                    )
+                    .await;
+
+                    return Err(JDCError::shutdown(e.kind));
                 }
 
                 info!("Sv2 Template Provider setup done");
@@ -246,13 +269,15 @@ impl JobDeclaratorClient {
                 ) {
                     Some(unix_socket_path) => unix_socket_path,
                     None => {
-                        error!(
-                                "Could not determine Bitcoin data directory. Please set data_dir in config."
-                            );
                         self.cancellation_token.cancel();
-                        self.shutdown_notify.notify_waiters();
-                        self.is_alive.store(false, Ordering::Relaxed);
-                        return;
+                        self.graceful_shutdown(
+                            task_manager,
+                            bitcoin_core_sv2_join_handle,
+                            bitcoin_core_sv2_cancellation_token,
+                        )
+                        .await;
+
+                        return Err(JDCError::shutdown(JDCErrorKind::InitializationError("Could not determine Bitcoin data directory. Please set data_dir in config.".to_string())));
                     }
                 };
 
@@ -367,8 +392,22 @@ impl JobDeclaratorClient {
                     _ = initial_channel_manager.allocate_tokens(2).await;
                 }
                 Err(e) => {
-                    tracing::error!("Failed to initialize upstream: {:?}", e);
+                    error!("Failed to initialize upstream: {:?}", e);
+
+                    if e.action == Action::Shutdown {
+                        self.cancellation_token.cancel();
+                        self.graceful_shutdown(
+                            task_manager,
+                            bitcoin_core_sv2_join_handle,
+                            bitcoin_core_sv2_cancellation_token,
+                        )
+                        .await;
+
+                        return Err(e);
+                    }
+
                     mode.set_solo_mining();
+                    info!("Fallback to solo mining mode");
                 }
             };
         }
@@ -403,12 +442,12 @@ impl JobDeclaratorClient {
         info!("Spawning status listener task...");
         let mut fallback_token = fallback_coordinator.token();
 
-        loop {
+        let loop_error = loop {
             tokio::select! {
                 biased;
 
                 _ = self.cancellation_token.cancelled() => {
-                    break;
+                    break None;
                 }
                 _ = fallback_token.cancelled() => {
                     warn!("Upstream/Job Declarator connection dropped — attempting reconnection...");
@@ -456,7 +495,7 @@ impl JobDeclaratorClient {
                         Err(e) => {
                             error!(error = ?e, "Failed to initialize channel manager during fallback");
                             self.cancellation_token.cancel();
-                            break;
+                            break Some(JDCError::shutdown(e.kind));
                         }
                     };
 
@@ -511,7 +550,14 @@ impl JobDeclaratorClient {
                             _ = channel_manager.allocate_tokens(2).await;
                         }
                         Err(e) => {
-                            tracing::error!("Failed to initialize upstream: {:?}", e);
+                            error!("Failed to initialize upstream: {:?}", e);
+
+                            if e.action == Action::Shutdown {
+                                info!("Shutdown requested while initializing upstream during fallback, initiate graceful shutdown");
+                                self.cancellation_token.cancel();
+                                break Some(e);
+                            }
+
                             channel_manager
                                 .upstream_state
                                 .set(UpstreamState::SoloMining);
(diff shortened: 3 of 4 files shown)


miner-apps/jd-client/src/lib/error.rs
```diff
@@ -84,12 +87,14 @@ impl CanDisconnect for ChannelManager {}
 impl CanFallback for Upstream {}
 impl CanFallback for JobDeclarator {}
 impl CanFallback for ChannelManager {}
+impl CanFallback for JobDeclarationClient {}
```
bit-aloo: This should never happen. At the top level, we should only receive a shutdown error action and nothing else. If the fallback variant propagates, that indicates an error and should be handled in a different way.

Shourya742 (commented):
Overall approach looks good.

miner-apps/jd-client/src/lib/error.rs
nulllpc: https://github.com/stratum-mining/sv2-apps/pull/456/changes#diff-4bb69fad525b1fa19981d967feeba9ea788a91e2fbe67727688e9f51715f0d8aR672

Implementing CanFallback is for `initialize_jd` to be able to distinguish fallback from true shutdown signal so we can fail fast instead of retrying blindly. I agree that fallback should not generally exist on this level, but does this finer shutdown handling justify the trait for `JobDeclarationClient`?

nulllpc: Hi @Shourya742 how can I proceed with this PR? Is the reasoning above good enough for you? I try to keep the PR lean so we can merge a few improvements at a time instead of a giant PR. Please let me know what you think

bit-aloo: I will have a look again tomorrow, complete missed it.

bit-aloo: You might need to update this PR

nulllpc:
Hi @Shourya742 I updated my PR accordingly, I reviewed the changes made in #467 and I think my changes still align well with changes there and the comment you made here

https://github.com/stratum-mining/sv2-apps/pull/467#issuecomment-4352306763

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -78,13 +78,17 @@ impl JobDeclaratorClient {
...
+            .consensus_encode(&mut encoded_outputs)
+            .map_err(|_| {
+                self.cancellation_token.cancel();
+                self.shutdown_notify.notify_waiters();
+                self.is_alive.store(false, Ordering::Relaxed);
+
+                JDCError::shutdown(JDCErrorKind::InitializationError(
+                    "Invalid coinbase output in config".to_string(),
+                ))
+            })?;
```
bit-aloo: We shouldn't be having side effects inside the combinator, kindly revert to if let Err,

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -120,16 +124,13 @@ impl JobDeclaratorClient {
...
-                return;
-            }
-        };
+        .map_err(|err| {
+            self.cancellation_token.cancel();
+            self.shutdown_notify.notify_waiters();
+            self.is_alive.store(false, Ordering::Relaxed);
+
+            JDCError::shutdown(err.kind)
+        })?;
```
bit-aloo: Same for this we should not perform side effects inside the closure

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -204,16 +205,13 @@ impl JobDeclaratorClient {
...
-                        return;
-                    }
-                };
+                .map_err(|err| {
+                    self.cancellation_token.cancel();
+                    self.shutdown_notify.notify_waiters();
+                    self.is_alive.store(false, Ordering::Relaxed);
+
+                    JDCError::shutdown(err.kind)
+                })?;
```
bit-aloo: Same for this

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -230,20 +228,17 @@ impl JobDeclaratorClient {
...
-                };
+                ).ok_or_else(|| {
+                    self.cancellation_token.cancel();
+                    self.shutdown_notify.notify_waiters();
+                    self.is_alive.store(false, Ordering::Relaxed);
+
+                    JDCError::shutdown(
+                        JDCErrorKind::InitializationError("Could not determine Bitcoin data directory. Please set data_dir in config.".to_string())
+                    )
+                })?;
```
bit-aloo: Same for this

Shourya742 (changes requested):
Changes look okay to me. As a rule of thumb, avoid adding side effects in combinators. Once we fix this, I will approve it.

nulllpc:
I addressed all comments, please review again @Shourya742

miner-apps/jd-client/src/main.rs
```diff
@@ -25,5 +25,8 @@ async fn inner_main() {
     });
 
     init_logging(jdc_config.log_file());
-    JobDeclaratorClient::new(jdc_config).start().await;
+    if let Err(e) = JobDeclaratorClient::new(jdc_config).start().await {
+        eprintln!("Job Declarator Client failed to start: {e}");
```
bit-aloo: As we have event subscriber already initialized at this point.
```suggestion
        tracing::error!("Job Declarator Client failed to start: {e}");
```

integration-tests/lib/mod.rs
```diff
@@ -265,7 +265,11 @@ pub fn start_jdc(
     );
     let ret = jd_client_sv2::JobDeclaratorClient::new(jd_client_proxy);
     let ret_clone = ret.clone();
-    tokio::spawn(async move { ret_clone.start().await });
+    tokio::spawn(async move {
+        if let Err(e) = ret_clone.start().await {
+            panic!("Integration test JDC failed to start: {e:?}");
```
bit-aloo:
```suggestion
            panic!("Integration test JDC failed to start: {e}");
```

Shourya742 (commented):
Few nits

Shourya742:
@nulllpc once you fix the clippy and fmt nits, I’ll approve. Nice work.

nulllpc:
All done! Let's get this merged, this is my first PR for sv2. Thanks for your reviews @Shourya742

Shourya742:
@nulllpc can you squash the commits.

nulllpc:
Done! @Shourya742 I double-checked commit message as well

Shourya742 (approved):
Thanks! ACK

nulllpc:
Hi @Shourya742 can we merge this PR, are we waiting for sth?

GitGab19:
@nulllpc CI is failing here, I guess you need to rebase it and fix issues.

nulllpc:
Hi @GitGab19, I updated the PR, it's ready for review now

integration-tests/lib/mod.rs
```diff
@@ -274,7 +274,11 @@ pub fn start_jdc(
     );
     let ret = jd_client_sv2::JobDeclaratorClient::new(jd_client_proxy);
     let ret_clone = ret.clone();
-    tokio::spawn(async move { ret_clone.start().await });
+    tokio::spawn(async move {
+        if let Err(e) = ret_clone.start().await {
+            panic!("Integration test JDC failed to start: {e}");
```
GitGab19:
```suggestion
            panic!("JDC failed to start: {e}");
```

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -797,7 +811,10 @@ async fn try_initialize_single(
         config.required_extensions().to_vec(),
     )
     .await
-    .map_err(|error| error.kind)?;
+    .map_err(|error| match error.action {
+        Action::Shutdown => JDCError::shutdown(error.kind),
+        _ => JDCError::fallback(error.kind),
+    })?;
```
GitGab19, line 888: Do we really need this match?

cc @Shourya742

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -811,7 +828,10 @@ async fn try_initialize_single(
         task_manager.clone(),
     )
     .await
-    .map_err(|error| error.kind)?;
+    .map_err(|error| match error.action {
+        Action::Shutdown => JDCError::shutdown(error.kind),
+        _ => JDCError::fallback(error.kind),
+    })?;
```
GitGab19, line 905: Do we really need this match?

cc @Shourya742

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -736,11 +743,18 @@ impl JobDeclaratorClient {
...
                                 info!("Shutdown requested after upstream initialization failure, aborting retries");
-                                return Err(JDCErrorKind::CouldNotInitiateSystem);
+                                return Err(
+                                    JDCError::shutdown(JDCErrorKind::CouldNotInitiateSystem)
+                                );
                             }
                             _ = tokio::time::sleep(Duration::from_secs(1)) => {}
                         }
 
+                        if e.action == Action::Shutdown {
```
GitGab19, line 824: Can you add a short comment here explaining why we need this `if` statement?

miner-apps/jd-client/src/lib/mod.rs
bit-aloo: Kinda yes. The methods we invoke during fallback in an error path can result in either a fallback or a shutdown. Earlier, before this change, regardless of whether the error variant indicated a fallback or a shutdown, we would always retry, even in the shutdown case. This match statement simply lets the fallback handler know that if something went terribly wrong (i.e., an error with the shutdown variant), it should immediately stop the process and exit gracefully instead of retrying. I think we can improve this further, but for now this is okay.

miner-apps/jd-client/src/lib/mod.rs
bit-aloo, line 552: You probably need to check the action on error here, whether it’s fallback or shutdown. Right now, we’re ignoring the error and defaulting to fallback on solo mining. If the error action is shutdown, we should shut down.

GitGab19, line 411: Probably the same check explained in https://github.com/stratum-mining/sv2-apps/pull/456/changes#r3225801560 should be done here?

What do you think @Shourya742 ?

bit-aloo: Yes

miner-apps/jd-client/src/lib/mod.rs
nulllpc: So basically `try_initialize_single` can return either a fallback or shutdown signal. The rationale here is to catch the shutdown signal and then fail fast, instead of retrying again. Before, it was passing only `JDCErrorKind` so we wouldn't be able to differentiate between a shutdown and a fallback signal. This is the same as @Shourya742 explained here: https://github.com/stratum-mining/sv2-apps/pull/456/changes#r3224976497

miner-apps/jd-client/src/lib/mod.rs
nulllpc: Nice catch, I'm on it

nulllpc: I realized that when we fallback to solo mining here, we only set the mode but we do not set the `upstream_state`. Is this intentional or a bug? @Shourya742

bit-aloo: It is already set in the fallback branch for the channel manager, so we don't need an assignment here

nulllpc: Ok, I see `ChannelManager::new()` initiates with upstream state as solo mining so technically it still works, but I think it's better to be explicit here. What do you think @Shourya742 @GitGab19

bit-aloo: Not really. I’m going to refactor the startup and fallback code soon anyway to make it more streamlined, so it’s not really needed right now.

nulllpc: Cool, thanks for the explanation

nulllpc:
I resolved all comments @Shourya742 @GitGab19 Please check again when you have time

nulllpc:
Tests are failing because they still expect setup to fallback to solo mining, I'll adjust the tests accordingly

miner-apps/jd-client/src/lib/mod.rs
GitGab19: Ok, can you write a short comment explaining that though?

nulllpc: Oh, you mean a comment in the code. Sorry for misunderstanding, I'll add it

nulllpc:
Hello @Shourya742 @GitGab19, the failed test comes from this [PR](https://github.com/stratum-mining/sv2-apps/pull/467/changes#diff-cf3e0827bd6d0169d5c4d543295cbee56aedf1dc36286e651fef3cad1e075319R3): `jdc_fallback_to_solo_mines_block_with_bitcoin_core_ipc`. Now that we expect the JDC to shutdown properly when given a shutdown signal as commented [here](https://github.com/stratum-mining/sv2-apps/pull/456#discussion_r3225801560), how should I adjust the test? The JDC now shutdown instead of falling back to solo mining, so I'm thinking about triggering a proper fallback signal? So far I have not found anything that can trigger a fallback here

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -775,7 +801,7 @@ impl JobDeclaratorClient {
         }
 
         tracing::error!("All upstreams failed after {} retries each", MAX_RETRIES);
-        Err(JDCErrorKind::CouldNotInitiateSystem)
+        Err(JDCError::shutdown(JDCErrorKind::CouldNotInitiateSystem))
```
bit-aloo: This should trigger a fallback. If we are unable to connect to any upstream, it means the JDC should run in solo mining mode. This change should make the test pass. I would suggest using a different error kind for this.

bit-aloo:
```suggestion
        Err(JDCError::fallback(JDCErrorKind::CouldNotInitiateSystem))
```

nulllpc: Changed the error kind to `SetupConnectionError`. I also see `UpstreamNotFound` but it's currently used in a upstream channel missing case so I think it's a different context. Let me know if I should use another error kind or create a new one

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -775,7 +802,7 @@ impl JobDeclaratorClient {
         }
 
         tracing::error!("All upstreams failed after {} retries each", MAX_RETRIES);
-        Err(JDCErrorKind::CouldNotInitiateSystem)
+        Err(JDCError::fallback(JDCErrorKind::SetupConnectionError))
```
bit-aloo: We can have lets say fallback variant to this error

nulllpc: Done!

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -367,8 +370,14 @@ impl JobDeclaratorClient {
                     _ = initial_channel_manager.allocate_tokens(2).await;
                 }
                 Err(e) => {
-                    tracing::error!("Failed to initialize upstream: {:?}", e);
+                    error!("Failed to initialize upstream: {:?}", e);
+
+                    if e.action == Action::Shutdown {
+                        return Err(e);
+                    }
```
bit-aloo, line 407: We need to make sure the task here do cancel and follow the shutdown procedure.

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -511,7 +520,12 @@ impl JobDeclaratorClient {
                             _ = channel_manager.allocate_tokens(2).await;
                         }
                         Err(e) => {
-                            tracing::error!("Failed to initialize upstream: {:?}", e);
+                            error!("Failed to initialize upstream: {:?}", e);
+
+                            if e.action == Action::Shutdown {
+                                return Err(e);
```
bit-aloo: This should be break, considering we want the shutdown to happen gracefully

miner-apps/jd-client/src/lib/mod.rs
nulllpc: I introduced a new fn graceful_shutdown to encapsulate task canceling and shutdown procedure. The fn might try canceling even if there's no task but I don't think that's a problem

https://github.com/stratum-mining/sv2-apps/pull/456/changes#diff-4bb69fad525b1fa19981d967feeba9ea788a91e2fbe67727688e9f51715f0d8aR650

I believe the task wasn't cancelled properly before here, it should work with the new fn
https://github.com/stratum-mining/sv2-apps/pull/456/changes#diff-4bb69fad525b1fa19981d967feeba9ea788a91e2fbe67727688e9f51715f0d8aL232

bit-aloo (approved):
Thanks!

miner-apps/jd-client/src/lib/mod.rs
GitGab19, line 352: Shouldn't we return a shutdown error here, and call the graceful_shutdown?

cc @Shourya742

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -511,7 +520,14 @@ impl JobDeclaratorClient {
                             _ = channel_manager.allocate_tokens(2).await;
                         }
                         Err(e) => {
-                            tracing::error!("Failed to initialize upstream: {:?}", e);
+                            error!("Failed to initialize upstream: {:?}", e);
+
+                            if e.action == Action::Shutdown {
+                                info!("Shutdown requested while initializing upstream during fallback, initiate graceful shutdown");
+                                self.cancellation_token.cancel();
+                                break;
```
plebhash: this `break` would lead to `start` returning `Ok(())`, which is inconsistent with what we're doing on the L372-L377 block above

in terms of overall functionality, the behavior is still arguably correct

but since we're taking this opportunity to make error handling robust and consistent on the crate, we should look into it

one could naively replace `break` with a `return Err(e)` on L528, but that would bypass the common teardown logic

so a better way to address this would be to make the L414 loop return `Option<JDCError>`, and propagate it on the end of `start` in case it's `Some`:

```diff
diff --git a/miner-apps/jd-client/src/lib/mod.rs b/miner-apps/jd-client/src/lib/mod.rs
index 033cd2553..60fecd9da 100644
--- a/miner-apps/jd-client/src/lib/mod.rs
+++ b/miner-apps/jd-client/src/lib/mod.rs
@@ -411,13 +411,12 @@ impl JobDeclaratorClient {

         info!("Spawning status listener task...");
         let mut fallback_token = fallback_coordinator.token();
-
-        loop {
+        let loop_error = loop {
             tokio::select! {
                 biased;

                 _ = self.cancellation_token.cancelled() => {
-                    break;
+                    break None;
                 }
                 _ = fallback_token.cancelled() => {
                     warn!("Upstream/Job Declarator connection dropped — attempting reconnection...");
@@ -465,7 +464,7 @@ impl JobDeclaratorClient {
                         Err(e) => {
                             error!(error = ?e, "Failed to initialize channel manager during fallback");
                             self.cancellation_token.cancel();
-                            break;
+                            break None;
                         }
                     };

@@ -525,7 +524,7 @@ impl JobDeclaratorClient {
                             if e.action == Action::Shutdown {
                                 info!("Shutdown requested while initializing upstream during fallback, initiate graceful shutdown");
                                 self.cancellation_token.cancel();
-                                break;
+                                break Some(e);
                             }

                             channel_manager
@@ -625,10 +624,10 @@ impl JobDeclaratorClient {
                 _ = tokio::signal::ctrl_c() => {
                     info!("Ctrl+C received — initiating graceful shutdown...");
                     self.cancellation_token.cancel();
-                    break;
+                    break None;
                 }
             }
-        }
+        };

         if let Some(bitcoin_core_sv2_cancellation_token) = bitcoin_core_sv2_cancellation_token {
             bitcoin_core_sv2_cancellation_token.cancel();
@@ -644,6 +643,10 @@ impl JobDeclaratorClient {

         self.graceful_shutdown(task_manager).await;

+        if let Some(e) = loop_error {
+            return Err(e);
+        }
+
         Ok(())
     }
```

miner-apps/jd-client/src/lib/mod.rs
bit-aloo: hmm... thats right to question, but I didn't flag it because we anyway don't loop and directly shutdown later. I have a better design for this, in mind, I will implement it later. It gets confusing considering its easy to miss these conditions.

miner-apps/jd-client/src/lib/mod.rs
bit-aloo: Thats a nice catch.

bit-aloo: @nulllpc you can get this in

miner-apps/jd-client/src/lib/mod.rs
GitGab19: I have to confess that the changes on this PR caused me more confusion than clarity. 

I might have to completely grasp how all the new error handling works, but from my perspective we are introducing many places where we can easily forget to call/handle things.

bit-aloo: Yup, I can understand the reasoning behind this, especially since earlier we were not handling errors in the start method properly  we were either unwrapping them or ignoring them entirely. This PR ensures that we now return proper errors for such cases, and in a way, it also highlights that the start method itself could be designed better.

Since we no longer have a central status main loop, the start method has effectively become just a sequence of state transitions (State A -> State B -> State A -> State B -> State C) and nothing more. We also have some code duplication because of this. That said, this now gives us room to improve the ergonomics in ways that were not really feasible before, and hopefully should be intuitive.

We should be able to make it much cleaner soon.

miner-apps/jd-client/src/lib/mod.rs
nulllpc: Hi @Shourya742, based on the comments of @plebhash and @GitGab19, I'm thinking about refactoring the start function so that we can have 1 clean shutdown sequence for all exits instead of handling shutdown in each branching logic. I also found another place where we didn't clean up the `BitcoinCoreSv2TDP` so that's one more reason for refactoring. I think that'd be much safer and also make the start fn easier to read. I'll push a commit for it today or tomorrow, if it doesn't look good, then we can revert and keep the changes in this PR minimal

nulllpc: I decided to leave refactoring to another PR as it's pretty large. It can be a fun learning task for me if you guys don't mind. Please let me know if you have some ideas on refactoring @Shourya742. Other than that, I added the suggested change from @plebhash

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -367,8 +367,17 @@ impl JobDeclaratorClient {
                     _ = initial_channel_manager.allocate_tokens(2).await;
                 }
                 Err(e) => {
-                    tracing::error!("Failed to initialize upstream: {:?}", e);
+                    error!("Failed to initialize upstream: {:?}", e);
+
+                    if e.action == Action::Shutdown {
```
nulllpc, line 397: @Shourya742 I think we have a similar problem to what you suggested here:
https://github.com/stratum-mining/sv2-apps/pull/456#discussion_r3283324316

At this point, `bitcoin_core_sv2_join_handle` is present, so if we return here, we need to clean it up as well.  Overall, due to the current structure of the function, it's hard to ensure the shutdown sequence is done correctly. But I think we can leave this for another PR where we refactor the start function, what do you think?

bit-aloo: Graceful should be handling clean exit to bitcoin-core-sv2 handle, its not currently doing that,

nulllpc: Ok, so I should add that in this PR as well right?

bit-aloo: Yes, considering we are abstracting shutdown into a separate method, we should do it incrementally. First, update the shutdown handler state for TaskManager, and then progressively extend it to include bitcoin-sv2-handler, exposing shutdown only once all of them are integrated

nulllpc: Done! Can you give it a review please @Shourya742

plebhash:
as discussed on the dev call today, closing in favor of https://github.com/stratum-mining/sv2-apps/issues/526

#168 remains relevant outside the scope of `JobDeclaratorClient::start`

PR #458, opened 2026-04-26
[prototype] Add ASIC monitoring APIs to stratum-apps

pavlenex:
## Summary

This is a work-in-progress prototype for exposing ASIC monitoring data from `stratum-apps` through the existing monitoring HTTP server.

It depends on the matching sv2-ui https://github.com/stratum-mining/sv2-ui/pull/126 and is intended only for prototype/testing use. The goal is to show engineers how ASIC telemetry and control data could be surfaced to the UI, and to exercise the dashboard/add-miner/manage-miner flows end to end.

## Scope

- Adds optional `asic-rs` integration behind the existing monitoring feature path.
- Exposes ASIC discovery through the monitoring API.
- Exposes per-miner telemetry lookup by management IP.
- Exposes SV1-client ASIC telemetry lookup using the client peer endpoint.
- Adds ASIC action endpoints for supported miner operations.
- Adds API response types for miner identity, telemetry, hashboards, fans, pools, messages, and capabilities.
- Keeps the integration scoped to monitoring APIs rather than adding miner-control logic to sv2-ui’s Node server.

## Notes

This is not intended to ship as-is. It is a prototype integration point for review and local testing while the sv2-ui UX is developed in parallel.

Known limitations:
- Rented/proxied hashpower usually cannot expose ASIC telemetry because only the SV1 stratum connection is visible.
- Docker Desktop can mask real miner peer IPs, so local testing may require explicit scan targets or Docker network setup.
- Discovery/control depends on what `asic-rs` supports for the target miner or simulator.


docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ###
 # ---- Builder Stage ----
 ###
-FROM rust:1.85-slim-bookworm AS builder
+FROM rust:1.88-slim-bookworm AS builder
```
pavlenex, line 4: This was a requirement, and a potential blocker we have so we should discuss if we're up for bumping it

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.85.0"
-components = [ "rustfmt", "clippy", "rust-analyzer" ]
\ No newline at end of file
+channel = "1.88.0"
```
pavlenex, line 2: potential blocker as it's a requirement in asics-rs

PR #452, opened 2026-04-23
Remove status main loop from JDC

Shourya742:
part of #234

PR #461, opened 2026-04-27
Remove status main loop from tproxy

### LINKED ISSUE #234: need to remove main status loop

bit-aloo:
closes: #234


miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -347,18 +385,15 @@ impl Sv1Server {
...
+                                let sv1_server = self.clone();
                                 Downstream::run_downstream_tasks(
                                     downstream,
                                     cancellation_token.clone(),
                                     fallback_coordinator.clone(),
-                                    status_sender,
                                     task_manager.clone(),
+                                    move || async move {
+                                        sv1_server.handle_downstream_disconnect(downstream_id).await;
+                                    },
```
GitGab19, line 413: Why this change?

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -161,10 +154,59 @@ pub struct Downstream {
...
+                warn!(
+                    downstream_id = self.downstream_id,
+                    action = ?other,
+                    error_kind = ?e.kind,
+                    "{context} returned an unhandled action"
+                );
+                LoopControl::Continue
+            }
+        }
+    }
```
GitGab19, line 227: Why are we replicating this in multiple places? We have this in every component (Sv1Server, Downstream, ChannelManager, Upstream).

Why can't we have this logic in a single place?

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -204,13 +246,16 @@ impl Downstream {
...
         cancellation_token: CancellationToken,
         fallback_coordinator: FallbackCoordinator,
-        status_sender: StatusSender,
         task_manager: Arc<TaskManager>,
-    ) {
+        on_disconnect: F,
+    ) where
+        F: FnOnce() -> Fut + Send + 'static,
+        Fut: Future<Output = ()> + Send + 'static,
+    {
```
GitGab19, line 277: What is this?

miner-apps/translator/src/lib/mod.rs
```diff
@@ -232,38 +229,15 @@ impl TranslatorSv2 {
...
                                 info!("All components finished fallback cleanup");
 
-                                // Drain any buffered status messages from old components
-                                while let Ok(old_status) = status_receiver.try_recv() {
-                                    debug!("Draining buffered status message: {:?}", old_status.state);
-                                }
-
                                 // Create a fresh FallbackCoordinator for the reconnection attempt
                                 fallback_coordinator = FallbackCoordinator::new();
+                                fallback_token = fallback_coordinator.token();
```
GitGab19, line 236: Why do we have this here? Where is it used?

bit-aloo: This is used to refresh the fallback token. Each time a fallback occurs, we create a new fallback_coordinator, which then refreshes the token used in the select encompassing this block.

miner-apps/translator/src/lib/sv1/downstream.rs
bit-aloo: As the closure needs to handle async.

miner-apps/translator/src/lib/sv1/downstream.rs
bit-aloo: If we try to generalize this, the method becomes quite ugly. It’s okay to keep this handler simple at the call site with a specific action strategy. You can read more in the discussion we had on JDC here: https://github.com/stratum-mining/sv2-apps/pull/452#pullrequestreview-4166846787

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
bit-aloo: `handle_downstream_disconnect` in sv1_server is async, so this is required. It is async because we send a `CloseChannel` message when a downstream disconnects.

plebhash: tbh I assume @GitGab19 is actually looking for a deeper context?

we should clarify the callback pattern, independently of `handle_downstream_disconnect` being `async`

---

we first introduced this pattern of cleaning up downstream clients via callbacks on Pool via #426

the problem at hand was: 

since we no longer have a `status` loop, there's no central place with global context to apply the necessary cleanups with a client disconnects

so `Downstream` needs a way to ask its owner (`ChannelManager`) to clean things up, but `Downstream` cannot own/know a `ChannelManager` instance... it's kind of a chicken/egg problem

the solution is to make `Downstream::start` take a `remove_downstream` callback function, which is meant to be called at the appropriate time when `Downstream` "dies":

https://github.com/stratum-mining/sv2-apps/blob/2d545a7de335d9c390e9b2adaa777748d6d4f150/pool-apps/pool/src/lib/downstream/mod.rs#L200-L205

here we can see the body of `ChannelManager::start_downstream_server` calling `Downstream::start` and passing `move |downstream_id| this.remove_downstream(downstream_id)` as the callback, where `this` is a reference to `ChannelManager`'s `self`

https://github.com/stratum-mining/sv2-apps/blob/2d545a7de335d9c390e9b2adaa777748d6d4f150/pool-apps/pool/src/lib/channel_manager/mod.rs#L367-L373

inside the body of `Downstream::start`, the `remove_downstream` callback can be potentially called on two different occasions:
- if the initial downstream `SetupConnection` handshake fails (L219)
- if the `Downstream` loop breaks (L260)

https://github.com/stratum-mining/sv2-apps/blob/2d545a7de335d9c390e9b2adaa777748d6d4f150/pool-apps/pool/src/lib/downstream/mod.rs#L206-L262

---

the only thing that `handle_downstream_disconnect` being `async` really explains is the fact that we're doing `move || async move { ... }` instead of `move || { ... }`, but that is meaningless without the context explained above

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -204,13 +246,16 @@ impl Downstream {
...
     /// and proper error reporting.
-    pub(super) fn run_downstream_tasks(
+    pub(super) fn run_downstream_tasks<F, Fut>(
         self,
         cancellation_token: CancellationToken,
         fallback_coordinator: FallbackCoordinator,
-        status_sender: StatusSender,
         task_manager: Arc<TaskManager>,
-    ) {
+        on_disconnect: F,
```
plebhash, line 273:
```suggestion
        remove_downstream: F,
```

since we already established this pattern on other apps, we should name the variables in a consistent way so the pattern becomes obvious

miner-apps/translator/src/lib/sv1/downstream.rs
plebhash: tbh I don't think that's what @GitGab19 is asking, he's not familiar with this callback pattern

please refer to https://github.com/stratum-mining/sv2-apps/pull/461/changes#r3157709518 for a detailed explanation

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -204,13 +246,16 @@ impl Downstream {
     /// The task will continue running until a cancellation signal is received or
     /// an unrecoverable error occurs. It ensures graceful cleanup of resources
     /// and proper error reporting.
-    pub(super) fn run_downstream_tasks(
+    pub(super) fn run_downstream_tasks<F, Fut>(
```
plebhash:
```suggestion
    pub(super) fn start<F, Fut>(
```

there's no reason for this to be called `Downstream::run_downstream_tasks` while JDC and Pool have an equivalent called `Downstream::start`

slightly out of scope, so not a blocking suggestion for this PR

if this is applied, it deserves a standalone commit

miner-apps/translator/src/lib/sv1/downstream.rs
plebhash: yeah this used to be way worse... we had `match e.action` repeated zillions of times

I tried to unify it into one single `handle_error_action` method per app, but it ended up extremely convoluted, as we had to cater for contextual data that differs for each subsystem

so having one dedicated `handle_error_action` per subsystem was the middle-ground

not perfect, but tbh I cannot think of anything better with the current `Action`-based infrastructure

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
plebhash: please note that on tProxy, `Downstream::run_downstream_tasks` is the equivalent of Pool's `Downstream::start`

not sure why they don't have identical names... IMHO they should because that would make patterns like this more obvious to fresh eyes, but that's a different discussion

miner-apps/translator/src/lib/sv2/upstream/mod.rs
```diff
@@ -453,17 +491,15 @@ impl Upstream {
...
-                                    handle_error(&status_sender, e).await;
+                                    error!("Upstream: failed to send sv2 frame: {e:?}");
+                                    fallback_token.cancel();
+                                    break;
                                 }
                             }
                             Err(e) => {
                                 error!("Upstream: channel manager receiver closed: {e}");
-                                handle_error(&status_sender, TproxyError::<error::Upstream>::shutdown(e)).await;
+                                cancellation_token.cancel();
```
plebhash:
```suggestion
                                if fallback_token.is_cancelled() {
                                    info!("Upstream: channel manager receiver closed during fallback cleanup: {e}");
                                } else {
                                    error!("Upstream: channel manager receiver closed unexpectedly: {e}");
                                    cancellation_token.cancel();
                                }
```

otherwise this can accidentally turn fallback cleanup into global shutdown

---

during fallback, the old `ChannelManager` closes its channels as part of normal cleanup.

if this `recv()` error branch wins the `tokio::select!` race before the `fallback_token.cancelled()` branch, `cancellation_token.cancel()` will cancel the global token.... then the reconnect path aborts immediately because `Upstream::new` sees the cancelled token and returns `CouldNotInitiateSystem`.

miner-apps/translator/src/lib/sv1/downstream.rs
plebhash: > Why are we replicating this in multiple places?

one key detail to note here (which can be easily overlooked), is that the each instance of `handle_error_action` is slightly different from the others... they're not identical copies of eachother

which connects to the comment above

GitGab19: I think it's fine to keep things as they are here, thanks for explaining this.

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: Thanks @plebhash :)

miner-apps/translator/src/lib/sv1/downstream.rs
plebhash: btw `on_disconnect` does sound slightly more intuitive, given that it's a callback function

but if we go with that naming, we should add commits renaming them on Pool + JDC

miner-apps/translator/src/lib/sv2/upstream/mod.rs
bit-aloo: There is a possibility of a race condition here. For example, if a shutdown or fallback occurs and the select block in another subsystem prioritizes a non-cancellation or non-shutdown branch, we might end up polling the channel receiver or sender. This could result in errors if the other end of the channel has already been dropped during the fallback or shutdown, potentially triggering a secondary fallback or shutdown.

I’m considering giving higher priority to the fallback and shutdown branches in the select block. Additionally, we could add a clause in handler_error_action such that if a shutdown or fallback is already in progress, it simply returns with continue, allowing the subsystem’s lifecycle branch to handle the token changes. I believe a similar approach should be applied to the Pool and JDC components as well.

Regarding the upstream system, I’ll align it with how we structure other subsystems. For instance, since we currently have recv and send directly in the select block, I’ll refactor them into a separate function.

miner-apps/translator/src/lib/sv1/downstream.rs
bit-aloo: Renamed `remove_downstream` with `on_disconnect` for JDC + Pool here: [e9bdb07](https://github.com/stratum-mining/sv2-apps/pull/461/commits/e9bdb0712a4eb6b3298caf1043e81429b06ab601)

miner-apps/translator/src/lib/sv1/downstream.rs
bit-aloo: Done here: 7d83a40

miner-apps/translator/src/lib/sv2/upstream/mod.rs
plebhash: while syncing with @Shourya742 on this topic, we're taking the approach on https://github.com/stratum-mining/sv2-apps/pull/461/changes/f36225d231534e0551ac40499d2b7b2bd0b8fe9b

this was already implemented on JDC via https://github.com/stratum-mining/sv2-apps/pull/467/changes/15c48e1c60d146e510a5b95270afe86a7ae2ebf6 of #467

it's a bit verbose and has some code repetition, but that's a tradeoff we've discussed recently

I don't love it but I'm fine to stick with it for the time being, as the issue raised above makes sense:

once cancellation token (either global or fallback) is triggered, we shouldn't be trying to process lingering tasks anymore... just clean things up and move on to either complete shutdown, or more importantly, restart subsystems (if fallback)

plebhash (approved):
small conflict to be solved but LGTM

PR #462, opened 2026-04-27
Fix tProxy share validation and server monitoring when running in AGGREGATED mode

### LINKED ISSUE #455: tProxy aggregated monitoring mixes upstream and downstream channel share accounting
while working on #450 I think I found a potential bug on tProxy

on this context:
https://github.com/stratum-mining/sv2-apps/blob/73a9f5d77ea47e1e5ec5c2cea097e9ee0cc5a6e0/miner-apps/translator/src/lib/monitoring.rs#L14-L21

we are looping over ALL `self.extended_channels` (including `AGGREGATED_CHANNEL_ID`):

https://github.com/stratum-mining/sv2-apps/blob/73a9f5d77ea47e1e5ec5c2cea097e9ee0cc5a6e0/miner-apps/translator/src/lib/monitoring.rs#L46-L58

so the reported "server channel" metrics are blending the upstream aggregated channel with the "internal" downstream channels

GitGab19:
This PR changes aggregated-mode tProxy monitoring so that the channel opened with the server reports share accounting data from the real aggregated upstream channel instead of mixing it with downstream virtual-channel data.

In order to do that, I had to change the way shares are validated when running aggregated mode, by validating shares directly against the `AGGREGATED_CHANNEL_ID` channel, instead of doing that against the individual downstream channels. 

Closes #455

### DIFF
--- a/miner-apps/translator/src/lib/monitoring.rs
+++ b/miner-apps/translator/src/lib/monitoring.rs
@@ -32,27 +32,7 @@ impl ServerMonitoring for ChannelManager {
                         aggregated_extended_channel.get_rollable_extranonce_size();
                     let version_rolling = aggregated_extended_channel.is_version_rolling();
                     let nominal_hashrate = aggregated_extended_channel.get_nominal_hashrate();
-
-                    let mut shares_acknowledged: u32 = 0;
-                    let mut shares_submitted: u32 = 0;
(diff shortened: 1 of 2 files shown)


plebhash (approved):

Shourya742 (approved):
ACK

Issue #391, opened 2026-03-30
`pool_sv2`: poisoned mutex in downstream accept path makes new connections fail immediately after Noise handshake


plebhash:
## Summary

After ~4 days uptime, hosted `pool_sv2` stopped accepting new downstream connections.

The confirmed direct cause is a Tokio worker panic on a poisoned mutex in `super_safe_lock()`:

https://github.com/stratum-mining/sv2-apps/blob/a0f270372faf622c2994a9fb6f89e8d5ec0f325b/stratum-apps/src/custom_mutex.rs#L58-L64

Once that mutex is poisoned, every new downstream connection can fail immediately after Noise handshake, before `SetupConnection` is processed.

## Confirmed Behavior

For failing connections, we observed this sequence:

1. TCP connect succeeds
2. Noise handshake completes
3. `pool_sv2` panics on a Tokio worker
4. the pool process closes the accepted socket
5. the client never reaches `SetupConnection`

This is **not** a client-side disconnect and **not** a `SetupConnection` rejection.

### Pool logs

Typical failure pattern:

```text
INFO  pool_sv2::channel_manager: New downstream connection socket_address=...
DEBUG stratum_apps::network_helpers::noise_stream: First handshake message received
DEBUG stratum_apps::network_helpers::noise_stream: Second handshake message sent
ERROR pool_sv2::io_task: Reader error error=SocketClosed
ERROR pool_sv2::downstream: Failed to set up downstream connection e=PoolError { kind: ChannelRecv(RecvError), ... }
```

At the same time, already-connected downstreams were still submitting valid shares successfully.

## tcpdump

Packet capture on the pool host shows the server closing first:

- client sends first Noise message
- server sends second Noise message
- server immediately sends FIN
- client only then sends its first post-Noise encrypted payload
- server responds with RST

So the server closes before the first encrypted SV2 frame is processed.

## strace

`strace` on the running `pool_sv2` process shows:
- r`ecvfrom(...)` client Noise message
- `sendto(...)` server Noise reply
- panic written to `stderr`
- `shutdown(fd, SHUT_WR)`
- `close(fd)`

Captured panic:
```
thread 'tokio-rt-worker' panicked at /home/sri/sv2-apps/stratum-apps/src/custom_mutex.rs:64:35:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```

## Impact
- New downstream connections fail immediately after Noise handshake
- Existing miners may continue working, masking the issue
[...]

PR #472, opened 2026-05-01
Dont expect in get negotiated extension with client

Shourya742:
This solves the panic we observed on the pool: https://github.com/stratum-mining/sv2-apps/issues/391#issuecomment-4358222584, though I am pretty sure there may be others. For now, this fixes the specific panic on the VPS.


GitGab19 (approved):

plebhash:
I'm syncing with @Shourya742 about this

so it seems that we're fixing one possible root cause, but we shouldn't necessarily close #391 with this PR

plebhash (approved):

PR #473, opened 2026-05-01
add connection timeout to tProxy

GitGab19:
This PR adds a timeout to the connection attempts of tProxy, which unblocks it in case the upstream is not listening at all.

PS: we are already doing this in JDC

### DIFF
--- a/miner-apps/translator/src/lib/sv2/upstream/mod.rs
+++ b/miner-apps/translator/src/lib/sv2/upstream/mod.rs
@@ -10,7 +10,7 @@ use async_channel::{unbounded, Receiver, Sender};
 use std::{net::SocketAddr, sync::Arc};
 use stratum_apps::{
     fallback_coordinator::FallbackCoordinator,
-    network_helpers::{self, connect_with_noise, resolve_host},
+    network_helpers::{self, connect_with_noise, resolve_host, TCP_CONNECT_TIMEOUT},
     stratum_core::{
         binary_sv2::Seq064K,
         common_messages_sv2::{Protocol, SetupConnection},
@@ -139,7 +139,10 @@ impl Upstream {
                 TproxyError::fallback(TproxyErrorKind::NetworkHelpersError(e.into()))
             })?;
 
-        match TcpStream::connect(resolved_addr).await {
+        match tokio::time::timeout(TCP_CONNECT_TIMEOUT, TcpStream::connect(resolved_addr))
+            .await
+            .map_err(TproxyError::fallback)?
+        {
             Ok(socket) => {
                 info!("Connected to upstream at {}", resolved_addr);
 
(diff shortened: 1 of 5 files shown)


miner-apps/translator/src/lib/sv2/upstream/mod.rs
```diff
@@ -29,6 +29,8 @@ use tokio::net::TcpStream;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, warn};
 
+const UPSTREAM_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+
```
bit-aloo: lets add this to `stratum-apps` and use in JDC as well.

Shourya742 (commented):
LGTM. One suggestion

miner-apps/translator/src/lib/sv2/upstream/mod.rs
GitGab19: Done

integration-tests/tests/translator_integration.rs
```diff
@@ -171,6 +171,50 @@ async fn test_translator_fallback_on_setup_connection_error() {
     shutdown_all!(translator, pool_2, pool_1);
 }
 
+// Demonstrates the scenario where TProxy falls back to the secondary upstream
+// after the primary upstream is not listening at all.
+#[tokio::test]
+async fn test_translator_fallback_on_tcp_connect_error() {
```
bit-aloo: Can we add a test that hits the TCP connect timeout path? This test uses an unused localhost port, so the OS immediately returns `ConnectionRefused`. That verifies fallback on connect error, but it would pass even without the timeout changes in this PR.

The timeout behavior only triggers when the TCP handshake is underway and does not complete within `TCP_CONNECT_TIMEOUT`.

```rust
2026-05-01T12:19:55.362778Z  INFO translator_sv2: Trying upstream 1 of 2: 127.0.0.1:43853
2026-05-01T12:19:55.362796Z  INFO translator_sv2: Connection attempt 1/3...
2026-05-01T12:19:56.364718Z  INFO translator_sv2::sv2::upstream: Trying to connect to upstream at 127.0.0.1:43853
2026-05-01T12:19:56.365272Z ERROR translator_sv2::sv2::upstream: Failed to connect to 127.0.0.1:43853: Connection refused (os error 111).
2026-05-01T12:19:56.365446Z  WARN translator_sv2: Attempt 1/3 failed for 127.0.0.1:43853: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
2026-05-01T12:19:56.365557Z  INFO translator_sv2: Connection attempt 2/3...
2026-05-01T12:19:57.366224Z  INFO translator_sv2::sv2::upstream: Trying to connect to upstream at 127.0.0.1:43853
2026-05-01T12:19:57.366838Z ERROR translator_sv2::sv2::upstream: Failed to connect to 127.0.0.1:43853: Connection refused (os error 111).
2026-05-01T12:19:57.366969Z  WARN translator_sv2: Attempt 2/3 failed for 127.0.0.1:43853: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
2026-05-01T12:19:57.367055Z  INFO translator_sv2: Connection attempt 3/3...
2026-05-01T12:19:58.368043Z  INFO translator_sv2::sv2::upstream: Trying to connect to upstream at 127.0.0.1:43853
2026-05-01T12:19:58.368337Z ERROR translator_sv2::sv2::upstream: Failed to connect to 127.0.0.1:43853: Connection refused (os error 111).
2026-05-01T12:19:58.368386Z  WARN translator_sv2: Attempt 3/3 failed for 127.0.0.1:43853: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
```

GitGab19: The point about this PR (and this test) is that without a timeout, tProxy is waiting for the connection to be established forever, and so it's not trying to do fallback either.

That's what the PR is fixing, and what the test wants to assert.

GitGab19:
@plebhash we need to add this patch to `release/v0.3.5` once merged.

integration-tests/tests/translator_integration.rs
bit-aloo: If we remove the timeout changes, the above test would still pass. What I mean is that the connection will be broken since the TCP handshake is never initiated.

To test the timeout, we would need to initiate the handshake but not complete it.

GitGab19: > To test the timeout, we would need to initiate the handshake but not complete it.

That would test the noise handshake timeout, not the TCP connection one which this PR is solving.

GitGab19: One way could be to connect tProxy to a random hostname/IP, but depending on external networking things is not a good idea for tests.

I'm gonna simply drop it from this PR.

bit-aloo: > > To test the timeout, we would need to initiate the handshake but not complete it.
> 
> That would test the noise handshake timeout, not the TCP connection one which this PR is solving.

Noise handshake happens after the TCP stream is established. I am mentioning about the TCP handshake in the comment above.

> One way could be to connect tProxy to a random hostname/IP, but depending on external networking things is not a good idea for tests.

> I'm gonna simply drop it from this PR.

Can't we run a local server and not allow it to complete the TCP handshake?

GitGab19: > Can't we run a local server and not allow it to complete the TCP handshake?

I don't think this complexity is really worth it tbh.

bit-aloo: Cool, I tested this locally, and the timeout mechanism works. I will approve it.

Shourya742 (approved):
tACK

plebhash:
> @plebhash we need to add this patch to `release/v0.3.5` once merged.

added the following to `release/v0.3.5` branch:
- b32adfbfe889303e681bd7335d8e7b6a7694f9aa
- 7fd93d15a174c774ed6bc2a024b57281712bef7e
- 97674b0b699c5ef398dbac89dc1178e4c70105cd

commits are not cherry-picked from this PR, since `v0.3.4` tProxy has already drifted away from `main`

so they are just 1:1 adaptations to on top of `v0.3.4` structure

PR #476, opened 2026-05-04
Ignore handle error during shutdown

bit-aloo:
refer: https://github.com/stratum-mining/sv2-apps/pull/461#discussion_r3167925310


GitGab19:
It's not clear to me why this is not part of #461 ?

Shourya742:
> It's not clear to me why this is not part of #461 ?

Because it touches the Pool and the context is somewhat different, it is part of the status main loop cleanup on tproxy PR. So, I didn’t want to mix contexts and make things confusing if we revisit the commit later.

plebhash (approved):

plebhash:
merged #461 @Shourya742 can you rebase?

PR #480, opened 2026-05-05
Make sure to drain pending channels states before lifecycle change

bit-aloo:

Issue #267, opened 2026-02-13
unify cleanup patterns


plebhash:
on #225 we added https://github.com/stratum-mining/sv2-apps/pull/225/changes/98f7ce75bb74833c7ea05638491af0e513397e8f which modularizes `Sv1Server` cleanup

there's other similar components that deserve a similar approach, ideally in a uniform way across the entire codebase

a non-comprehensive list with examples that come to mind:
- `ChannelManager`: which has slightly different implementations across Pool, JDC and tProxy
- `MonitoringServer`: which has a unified implementation in `stratum-apps`

plebhash:
during #484 we came to the realization that there's some pending discussions on this area

we need to decide the best strategies on cleanup (primarily after fallback triggers, but ideally also over shutdown triggers)

a few questions that come to mind (taking `ChannelManager` as an example, but also extensible to other subsystems):
- should we completely drop `ChannelManager` on fallback?
  - if yes:
    - what should be covered by `Drop` trait implementations?
    - which channels should be `close_and_drain`?
    - which channels should remain intact?
    - is there a unifying pattern with shared logic with other subsystems?
  - if no:
    - which channels should be `close_and_drain`?
    - which channels should remain intact?
    - is there a unifying pattern with shared logic with other subsystems?

PR #484, opened 2026-05-05
Make sure to drain channel on drop

bit-aloo:
This PR ensures that we drain the receiver end of the channel so that any sender awaiting on a send operation can be unblocked. This was one of the major causes of the translator fallback hanging.

### DIFF
--- a/integration-tests/lib/mock_roles.rs
+++ b/integration-tests/lib/mock_roles.rs
@@ -1,6 +1,6 @@
 use crate::utils::{create_downstream, create_upstream, message_from_frame, wait_for_client};
 use async_channel::Sender;
-use std::{convert::TryInto, net::SocketAddr};
+use std::{convert::TryInto, net::SocketAddr, time::Duration};
 use stratum_apps::{
     stratum_core::{
         codec_sv2::StandardEitherFrame,
@@ -126,6 +126,7 @@ impl MockDownstream {
 pub struct MockUpstream {
     listening_address: SocketAddr,
     setup: WithSetup,
+    disconnect_after_setup: Option<Duration>,
 }
 
 impl MockUpstream {
@@ -133,7 +134,13 @@ impl MockUpstream {
         Self {
             listening_address,
             setup,
+            disconnect_after_setup: None,
         }
+    }
+
+    pub fn disconnect_after_setup_connection_success(mut self, delay: Duration) -> Self {
+        self.disconnect_after_setup = Some(delay);
+        self
     }
 
     pub async fn start(self) -> Sender<AnyMessage<'static>> {
@@ -183,6 +190,13 @@ impl MockUpstream {
                                 "MockUpstream: sent SetupConnectionSuccess with flags {}",
                                 flags
                             );
+
+                            if let Some(delay) = self.disconnect_after_setup {
+                                tokio::time::sleep(delay).await;
+                                downstream_sender.close();
+                                downstream_receiver.close();
+                                return;
+                            }
                         } else {
                             let error = AnyMessage::Common(CommonMessages::SetupConnectionError(
                                 SetupConnectionError {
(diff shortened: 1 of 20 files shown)


plebhash:
IIUC this is meant to replace #480?

bit-aloo:
> IIUC this is meant to replace #480?

Yup, just mentioned there.

plebhash (approved):
as discussed on today dev call, I'm ok with merging as-is

but taking note for follow-up work here: https://github.com/stratum-mining/sv2-apps/issues/267#issuecomment-4381734069

bit-aloo:
> as discussed on today dev call, I'm ok with merging as-is
> 
> but taking note for follow-up work here: [#267 (comment)](https://github.com/stratum-mining/sv2-apps/issues/267#issuecomment-4381734069)

Not so soon... integration test is acting funny...

plebhash:
@Shourya742 if you replace 5601b1cc2844c3d7f388141d5540c52a80dca8d5 in favor of 15d9cebb308e71596dbc4b6e3f179bcec670083e it should unblock CI

bit-aloo:
> @Shourya742 if you replace [5601b1c](https://github.com/stratum-mining/sv2-apps/commit/5601b1cc2844c3d7f388141d5540c52a80dca8d5) in favor of [15d9ceb](https://github.com/stratum-mining/sv2-apps/commit/15d9cebb308e71596dbc4b6e3f179bcec670083e) it should unblock CI

Added, thanks.

bit-aloo:
I tested the possible combinations involving fallback on the tproxy and JDC, and everything seems to work fine

integration-tests/lib/mock_roles.rs
```diff
@@ -126,16 +126,23 @@ impl MockDownstream {
...
 impl MockUpstream {
     pub fn new(listening_address: SocketAddr, setup: WithSetup) -> Self {
         Self {
             listening_address,
             setup,
+            disconnect_after_setup: None,
         }
     }
 
+    pub fn disconnect_after_setup(mut self, delay: Duration) -> Self {
```
GitGab19:
```suggestion
    pub fn disconnect_after_setup_connection_success(mut self, delay: Duration) -> Self {
```

integration-tests/lib/mock_roles.rs
GitGab19, line 1: How the `disconnect_after_setup()` change introduced by @plebhash commit/suggestion unblocked the test?

Why wasn't the `shutdown()` enough?

GitGab19 (commented):
I tested multiple times the fallback, and it always worked as expected, great work @Shourya742!

I only left a couple of minor suggestions before merging it.

integration-tests/lib/mock_roles.rs
bit-aloo: Done

GitGab19 (approved):

integration-tests/lib/mock_roles.rs
bit-aloo: Initially, the test spawned two pools and randomly shut 1st one down after a few messages, expecting the tproxy to fallback to the other pool. However, for some reason, even after the pool was shut down, the tproxy continued receiving submit share messages, and the task responsible for triggering the fallback was neither being triggered nor polled by the executor.

@plebhash suggested using a mocked upstream instead, since the primary goal of this test is to verify whether fallback is functioning correctly. We still do not fully understand why this behavior occurred, so we decided to track it separately in #267.

That said, I tried testing all possible occurrences locally, and everything passed successfully. So, this likely seems related to task scheduling behavior within the integration tests.

GitGab19: Thanks for the explanation!

plebhash: yeah my rationale there was mostly around a divide-and-conquer mindset

by removing pool shutdown from the picture, we can concentrate on tproxy fallback behavior (which was the CI PITA at hand)

especially given that ITF is still mixing apps under the same runtime umbrella, we should aim to mock apps as much as possible in order to separate concerns

PR #469, opened 2026-04-29
Add monitoring workflow validation and notification

lucasbalieiro:
closes #460 

adds compile-time OpenAPI schema generation for the monitoring crate with CI validation and creation of issues in `sv2-ui` when a PR is merged.

```bash
cargo run --bin generate-openapi --features monitoring > stratum-apps/src/monitoring/openapi.json
```

PR #491, opened 2026-05-08
adapt `sv2-apps` with `channels_sv2::server::share_accounting` keeping track of rejected shares

plebhash:
companion https://github.com/stratum-mining/stratum/pull/2149

---

adds two new fields to `ExtendedChannelInfo` and `StandardChannelInfo`:

* `pub shares_rejected: u32`
* `pub shares_rejected_by_reason: HashMap<String, u32>`


stratum-apps/src/monitoring/snapshot_cache.rs
plebhash, line 1: disclaimer: I'm not 100% confident about what's going on in this file

stratum-apps/src/monitoring/prometheus_metrics.rs
plebhash, line 1: disclaimer: I'm not 100% confident about what's going on in this file

stratum-apps/src/monitoring/openapi.json
plebhash, line 1: it seems this we're touching the conventions established via #469?

am I doing anything stupid?

cc @lucasbalieiro @GitGab19

gimballock:
All of these changes seem to follow the correct conventions and processes, only nit i found was in the companion PR regarding use of &str instead of String to save some memory allocations.

stratum-apps/src/monitoring/openapi.json
lucasbalieiro: from what I can tell from the code, this change is expected, since we are adding two new fields to `ExtendedChannelInfo` and `StandardChannelInfo`:

* `pub shares_rejected: u32`
* `pub shares_rejected_by_reason: HashMap<String, u32>`

So the CI requesting changes here is not a false positive and in the context of the PR this change is expected. So I think you are not doing anything stupid.

If you're concerned about going against the conventions, I’d say the only guideline that might technically apply here is:

> `Add new data via optional fields or new endpoints - Don't modify existing response structures`

That said, in this PR the change seems entirely reasonable. The guideline itself might simply be too restrictive and could probably be phrased more clearly. Something along these lines would make more sense:

> `Request new data through optional payload/params fields or new endpoints when appropriate. Avoid removing existing response data`

adding new fields to a response is generally backward compatible because clients can safely ignore unknown fields. On the other hand, requiring new request data, especially if it is not optional, is much more likely to create compatibility issues for users.

plebhash:
> All of these changes seem to follow the correct conventions and processes, only nit i found was in the companion PR regarding use of &str instead of String to save some memory allocations.

thanks @gimballock !

so I assume this also includes `prometheus_metrics.rs` and `snapshot_cache.rs` right?

gimballock:
> > All of these changes seem to follow the correct conventions and processes, only nit i found was in the companion PR regarding use of &str instead of String to save some memory allocations.
> 
> thanks @gimballock !
> 
> so I assume this also includes `prometheus_metrics.rs` and `snapshot_cache.rs` right?

You're following the conventions correctly:
- Add GaugeVec field to PrometheusMetrics
- Register with the extra label dimension
- Update in update_metrics() loop with with_label_values
- Track label sets in PreviousPrometheusLabelSets and clean stale ones

But let me deploy this now to verify the metrics actually show up in grafana..
- The rejection metric is lazy loading so i don't see it, idk how to trigger a rejection right now
- Also it could be useful to include a shares submitted metric (sum of accepted + rejected)

The lazy-loaded metrics will cause issues for alerting and make associated grafana panels fail to load until first share rejection, so that one is high priority.

stratum-apps/src/monitoring/snapshot_cache.rs
```diff
@@ -58,8 +58,13 @@ use super::{
...
     /// Labels for server per-channel GaugeVecs: [channel_id, user_identity]
     server_channel_labels: HashSet<[String; 2]>,
+    /// Labels for server per-rejection GaugeVecs: [channel_id, user_identity, error_code]
+    server_rejected_share_labels: HashSet<[String; 3]>,
     /// Labels for client per-channel GaugeVecs: [client_id, channel_id, user_identity]
     client_channel_labels: HashSet<[String; 3]>,
+    /// Labels for client per-rejection GaugeVecs: [client_id, channel_id, user_identity,
+    /// error_code]
+    client_rejected_share_labels: HashSet<[String; 4]>,
 }
```
GitGab19, line 63: I'm not sure we need to touch this: can you please take a look at this @gimballock ?

GitGab19 (commented):
LGTM, apart from a minor thing on the Prometheus metrics cache, for which I'd like to get a confirmation from @gimballock on.

stratum-apps/src/monitoring/snapshot_cache.rs
plebhash: @gimballock did you mean the thumbs up reaction as:
- these lines should be removed or
- you'll take a look and report back?

gimballock: I will report back, i still didn't see the metrics appear initially on deploy and am verifying some changes before I make any suggestions

gimballock:
Heads up: `sv2_*_shares_rejected_total` lazy-loads and never appears in `/metrics` until the first rejection (verified empirically on testnet4 — panels and alerts silently broken with zero rejections). Opened https://github.com/plebhash/sv2-apps/pull/3 against this branch with a +122/-0 fix that pre-seeds the spec-defined `ERROR_CODE_SUBMIT_SHARES_*` constants to 0, plus a regression test. Verified live on testnet4 after rebuild.

stratum-apps/src/monitoring/snapshot_cache.rs
gimballock: the changes are small, I verified locally as you can see from the image, i don't have a way to generate rejected shares but this is what i wanted to see. This pr also includes a regression test that you can keep or not.

Screenshot 2026-05-15 at 11 12 30 AM

https://github.com/plebhash/sv2-apps/pull/3

plebhash:
thanks @gimballock 

cherry-picked the commit from https://github.com/plebhash/sv2-apps/pull/3 with a small adaptation:

`SPEC_REJECTION_CODES` as a variable name is a bit misleading, soon the spec will no longer suggest error codes, see https://github.com/stratum-mining/sv2-spec/pull/194

renamed to `SHARE_REJECTION_CODES`

GitGab19 (approved):
LGTM

PR #495, opened 2026-05-12
use `error_code` constants

plebhash:
companion https://github.com/stratum-mining/stratum/pull/2151

Issue #487, opened 2026-05-06
version bumps for `v0.5.0`


GitGab19:
Similar to #359

plebhash:
#495 is bumping:
- `stratum-apps`: `0.4.0` -> `0.5.0`
- `bitcoin_core_sv2`: `0.2.0` -> `0.3.0`
- `pool_sv2`: `0.3.0` -> `0.4.0`
- `jd_server_sv2`: `0.2.0` -> `0.3.0`
- `jd_client_sv2`:`0.2.0` -> `0.3.0`
- `translator_sv2`: `0.2.5` -> `0.3.0`
- `integration_tests_sv2`: `0.2.0` -> `0.3.0`

PR #497, opened 2026-05-13
enable payout verification in tProxy in case of SOLO mining

### LINKED ISSUE #308: tProxy should have a SOLO mining mode and check for Bitcoin address presence in the coinbase transaction
After the work on https://github.com/stratum-mining/sv2-apps/issues/135, we will start to host a SOLO mining endpoint on SRI VPS.

Home miners could start to use it as a SOLO pool endpoint (as a SECURE alternative of cksolo, publicpool, and similars).

In case they would need to run a tProxy (e.g. they are running Sv1-fw miners), it would be very good to have a functionality in tProxy which checks for the presence of the Bitcoin address specified by the user in the coinbase outputs of the job which is received from upstream.

By doing that, we can have tProxy completely preventing possible scams related to this.

GitGab19:
Closes #308


stratum-apps/src/payout.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+fn output_total_for_script(outputs: &[TxOut], script_pubkey: ScriptBuf) -> u64 {
+    outputs
+        .iter()
+        .filter(|output| !output.script_pubkey.is_op_return())
+        .filter(|output| output.script_pubkey.as_bytes() == script_pubkey.as_bytes())
+        .map(|output| output.value.to_sat())
+        .sum()
+}
```
bit-aloo: Can we make them inline, considering these are used only once. We don't require helper for these.

miner-apps/translator/src/lib/config.rs
```diff
@@ -129,6 +136,17 @@ impl TranslatorConfig {
...
+        &self,
+    ) -> Result<Option<PayoutMode>, PayoutModeError> {
+        if !self.verify_payout {
+            return Ok(None);
+        }
+
+        let payout_mode = PayoutMode::try_from(self.user_identity.as_str())?;
+        Ok(payout_mode.has_miner_payout().then_some(payout_mode))
+    }
+
```
bit-aloo, line 161: In case someone mistypes the user_identity and the payout_mode falls back to FullDonation with verify_payout enabled, there will be no miner_payouts. The current setup would still allow the tproxy to run, and the user would not even realize that their user_identity is incorrect. They might think everything is working fine. We should either prevent this entirely or at least warn the user that verify_payout is enabled while the system is running in FullDonation mode.

stratum-apps/src/payout.rs
GitGab19: Done

miner-apps/translator/src/lib/config.rs
GitGab19: I pushed some changes to remove this edge-case and improve things.

tProxy now uses a strict payout verification parser: when `verify_payout = true`, the configured `user_identity` must contain a verifiable payout address. Full donation / arbitrary pool usernames now fail startup with a clear config error instead of silently resulting in no verification, and address-looking typos are reported as invalid payout addresses.

Let me know what you think of it!

stratum-apps/src/payout.rs
```diff
@@ -0,0 +1,663 @@
...
+            Ok(payout_mode) => payout_mode,
+            Err(PayoutModeError::NoPayoutMode(_)) => {
+                let address = address_part_from_user_identity(user_identity);
+                if looks_like_bitcoin_address(address) {
+                    return Err(script_from_address(address)
+                        .expect_err("address-like identity should be invalid here"));
+                }
+
+                return Err(PayoutModeError::NoPayoutMode(user_identity.to_string()));
+            }
```
bit-aloo: Maybe something much simpler. We can rather just check whether the payout mode is `FullDonation` and `verify_payout` is enabled. In that case, we should error out, saying either disable verification or provide a valid Bitcoin address in `user_identity`, on the callsite.

stratum-apps/src/payout.rs
```diff
@@ -0,0 +1,663 @@
...
+        let outputs = Vec::<TxOut>::consensus_decode(&mut &outputs_bytes[..])
+            .map_err(|e| PayoutValidationError::DecodeCoinbaseOutputs(e.to_string()))?;
+
+        self.validate_coinbase_outputs(&outputs)
+    }
+
+    /// Returns true when this mode includes a miner payout address that can be verified.
+    pub fn has_miner_payout(&self) -> bool {
+        self.miner_script_pubkey().is_some()
+    }
```
bit-aloo: We can remove this helper method

GitGab19: Removed

stratum-apps/src/payout.rs
GitGab19: Simplified this by removing this helper from here, and putting the logic in tProxy itself.

plebhash (approved):
did a quick overview of the 4 commits, overall good patterns and clever usage of `MockUpstream` to assert tProxy falls back under some scammy solo pool

no findings from `gpt-5.3-codex`

plebhash:
conflict on `integration-tests/Cargo.toml` is due to version bumps introduced via #495 

see https://github.com/stratum-mining/sv2-apps/issues/487#issuecomment-4434542921

bit-aloo (approved):
ACK

PR #501, opened 2026-05-13
 tproxy + jdc: per-upstream user_identity

### LINKED ISSUE #485: translator: support per-upstream user_identity in [[upstreams]] config
## Problem

`user_identity` is a single global field on `TranslatorConfig` (`config.rs:40`). When the translator falls back to a secondary upstream, it sends the same identity string — built for the primary pool — to every pool in the list.

Different pools enforce different identity conventions:
- SRI uses `sri/solo/<btc_address>` (slash-delimited, no suffix)
- BlitzPool expects `<btc_address>.<worker_name>` (dot-delimited with worker suffix)

When SRI is the primary and BlitzPool is the fallback, the fallback connection sends `sri/solo/<addr>` to BlitzPool. BlitzPool immediately closes the socket or returns `OpenMiningChannelError(unknown-user)`, making the fallback always fail — even after the writer-task fix  lands.

This was reproduced locally and confirmed by logs shared on sv2-ui PR #119.

## Root cause

**`config.rs` lines 64–72** — The `Upstream` struct has no `user_identity` field:
```rust
pub struct Upstream {
    pub address: String,
    pub port: u16,
    pub authority_pubkey: Secp256k1PublicKey,
    // no user_identity
}
```

**`sv1/sv1_server/mod.rs` lines 943–946** — Identity is read from the global config, not from the active upstream:
```rust
let user_identity = if self.config.user_identity.starts_with("sri/") {
    self.config.user_identity.clone()
} else {
    format!("{}.miner{}", self.config.user_identity, miner_id)
};
```

This means whichever upstream is active, it always receives the identity built for the primary.

## Proposed fix

**1. Add optional `user_identity` to the `Upstream` struct (`config.rs`)**

```rust
pub struct Upstream {
    pub address: String,
    pub port: u16,
    pub authority_pubkey: Secp256k1PublicKey,
    pub user_identity: Option<String>,  // overrides global if set
}
```

TOML stays backward-compatible — `user_identity` is optional per upstream and falls back to the global value if not set:

```toml
user_identity = "default-identity"   # global fallback

[[upstreams]]
address = "75.119.150.111"
port = 3333
authority_pubkey = "9auq..."
user_identity = "sri/solo/bc1q..."   # SRI format for this upstream

[[upstreams]]
address = "blitzpool.yourdevice.ch"
port = 3333
authority_pubkey = "9bCo..."
user_identity = "bc1q....bitaxe"     # BlitzPool format for this upstream
```

**2. Thread the active upstream through to `Sv1Server` (`lib/mod.rs`)**

`initialize_upstream()` already returns once a connection succeeds. It should also return the matched `UpstreamEntry` so the caller knows which upstream was selected. The fallback mechanism already recreates `Sv1Server` on each failover (`lib/mod.rs:248`), so passing the active upstream at construction time is a clean fit.

**3. Use the per-upstream identity in `sv1_server/mod.rs:943–946`**

```rust
let base_identity = active_upstream
    .user_identity
    .as_deref()
    .unwrap_or(&self.config.user_identity);

let user_identity = if base_identity.starts_with("sri/") {
    base_identity.to_string()
} else {
    format!("{}.miner{}", base_identity, miner_id)
};
```

`channel_manager` does not need changes — identity flows through the `OpenExtendedMiningChannel` message already.

Without this fix, fallback between pools with different identity conventions always fails at the `OpenExtendedMiningChannel` step, silently crashing the translator after the fallback mechanism itself is triggered correctly.

xyephy:
This PR adds a `user_identity` field to each `[[upstreams]]` entry in the tproxy and jdc configs. When set, it is sent as-is in `OpenExtendedMiningChannel` and `AllocateMiningJobToken` instead of the global identity, allowing each upstream pool to receive the identity format it expects. Includes  ITF tests for both identity sending and fallback switching.

Closes #485

### DIFF
--- a/integration-tests/lib/mod.rs
+++ b/integration-tests/lib/mod.rs
@@ -208,10 +208,31 @@ pub fn start_jdc(
     enable_monitoring: bool,
     jdc_mode: Option<ConfigJDCMode>,
 ) -> (JobDeclaratorClient, SocketAddr, Option<SocketAddr>) {
+    let pool_with_ids: Vec<_> = pool
+        .iter()
+        .map(|&(p, j)| (p, j, "IT_TEST".to_string()))
+        .collect();
+    start_jdc_with_user_identities(
+        &pool_with_ids,
+        template_provider_config,
+        supported_extensions,
+        required_extensions,
+        enable_monitoring,
+        jdc_mode,
+    )
+}
+
+pub fn start_jdc_with_user_identities(
+    pool: &[(SocketAddr, SocketAddr, String)], /* (pool_address, jds_address,
+                                                * per_upstream_user_identity) */
+    template_provider_config: TemplateProviderType,
+    supported_extensions: Vec<u16>,
+    required_extensions: Vec<u16>,
+    enable_monitoring: bool,
+    jdc_mode: Option<ConfigJDCMode>,
+) -> (JobDeclaratorClient, SocketAddr, Option<SocketAddr>) {
     use jd_client_sv2::config::{JobDeclaratorClientConfig, PoolConfig, ProtocolConfig, Upstream};
     let jdc_address = get_available_address();
-    let max_supported_version = 2;
-    let min_supported_version = 2;
     let authority_public_key = Secp256k1PublicKey::try_from(
         "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72".to_string(),
     )
@@ -228,26 +249,19 @@ pub fn start_jdc(
     .unwrap();
     let upstreams = pool
         .iter()
-        .map(|(pool_addr, jds_addr)| {
+        .map(|(pool_addr, jds_addr, per_upstream_id)| {
             Upstream::new(
                 authority_pubkey,
                 pool_addr.ip().to_string(),
                 pool_addr.port(),
                 jds_addr.ip().to_string(),
                 jds_addr.port(),
+                per_upstream_id.clone(),
             )
         })
         .collect();
     let pool_config = PoolConfig::new(authority_public_key, authority_secret_key);
-    let protocol_config = ProtocolConfig::new(
-        max_supported_version,
-        min_supported_version,
-        coinbase_reward_script,
-    );
-    let shares_per_minute = 10.0;
-    let shares_batch_size = 1;
-    let user_identity = "IT-test".to_string();
-    let jdc_signature = "JDC".to_string();
+    let protocol_config = ProtocolConfig::new(2, 2, coinbase_reward_script);
     let monitoring_address = if enable_monitoring {
         Some(get_available_address())
     } else {
@@ -257,14 +271,13 @@ pub fn start_jdc(
     let jd_client_proxy = JobDeclaratorClientConfig::new(
         jdc_address,
         protocol_config,
-        user_identity,
-        shares_per_minute,
-        shares_batch_size,
+        10.0,
+        1,
         pool_config,
         3600,
         template_provider_config,
         upstreams,
-        jdc_signature,
+        "JDC".to_string(),
         jdc_mode,
         supported_extensions,
         required_extensions,
@@ -349,17 +362,89 @@ pub async fn start_sv2_translator(
     job_keepalive_interval_secs: Option<u16>,
     enable_monitoring: bool,
 ) -> (TranslatorSv2, SocketAddr, Option<SocketAddr>) {
-    start_sv2_translator_with_user_identity(
-        upstreams,
+    let upstreams: Vec<(SocketAddr, String)> = upstreams
+        .iter()
+        .map(|&addr| (addr, "user_identity".to_string()))
+        .collect();
+    start_sv2_translator_with_user_identities(
+        &upstreams,
         aggregate_channels,
         supported_extensions,
         required_extensions,
         job_keepalive_interval_secs,
-        "user_identity".to_string(),
-        false,
         enable_monitoring,
     )
     .await
+}
+
+pub async fn start_sv2_translator_with_user_identities(
+    upstreams: &[(SocketAddr, String)],
+    aggregate_channels: bool,
+    supported_extensions: Vec<u16>,
+    required_extensions: Vec<u16>,
+    job_keepalive_interval_secs: Option<u16>,
+    enable_monitoring: bool,
+) -> (TranslatorSv2, SocketAddr, Option<SocketAddr>) {
+    let job_keepalive_interval_secs = job_keepalive_interval_secs.unwrap_or(60);
+    let upstream_authority_pubkey = Secp256k1PublicKey::try_from(
+        "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72".to_string(),
+    )
+    .expect("failed to parse test authority pubkey");
+
+    let upstreams = upstreams
+        .iter()
+        .map(|(addr, per_upstream_id)| {
+            translator_sv2::config::Upstream::new(
+                addr.ip().to_string(),
+                addr.port(),
+                upstream_authority_pubkey,
+                per_upstream_id.clone(),
+            )
+        })
+        .collect();
+
+    let listening_address = get_available_address();
+    let listening_port = listening_address.port();
+
+    let minerd_process = MinerdProcess::new(SocketAddr::from((Ipv4Addr::LOCALHOST, 0)), false)
+        .await
+        .unwrap();
+    let min_individual_miner_hashrate = minerd_process.measure_hashrate().await.unwrap() as f32;
+
+    let downstream_difficulty_config = translator_sv2::config::DownstreamDifficultyConfig::new(
+        min_individual_miner_hashrate,
+        SHARES_PER_MINUTE,
+        true,
+        job_keepalive_interval_secs,
+    );
+
+    let monitoring_address = if enable_monitoring {
+        Some(get_available_address())
+    } else {
+        None
+    };
+    let monitoring_cache_refresh_secs = if enable_monitoring { Some(1) } else { None };
+    let config = translator_sv2::config::TranslatorConfig::new(
+        upstreams,
+        listening_address.ip().to_string(),
+        listening_port,
+        downstream_difficulty_config,
+        2,
+        2,
+        4,
+        false,
+        aggregate_channels,
+        supported_extensions,
+        required_extensions,
+        monitoring_address,
+        monitoring_cache_refresh_secs,
+    );
+    let translator_v2 = translator_sv2::TranslatorSv2::new(config);
+    let clone_translator_v2 = translator_v2.clone();
+    tokio::spawn(async move {
+        clone_translator_v2.start().await;
+    });
+    (translator_v2, listening_address, monitoring_address)
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -388,6 +473,7 @@ pub async fn start_sv2_translator_with_user_identity(
                 upstream_address,
                 upstream_port,
                 upstream_authority_pubkey,
+                user_identity.clone(),
             )
         })
         .collect();
@@ -423,7 +509,6 @@ pub async fn start_sv2_translator_with_user_identity(
         2,
         2,
         downstream_extranonce2_size,
-        user_identity,
         verify_payout,
         aggregate_channels,
         supported_extensions,
--- a/integration-tests/tests/translator_integration.rs
+++ b/integration-tests/tests/translator_integration.rs
@@ -2,6 +2,7 @@
 use integration_tests_sv2::{
     interceptor::{IgnoreMessage, MessageDirection, ReplaceMessage},
     mock_roles::{MockUpstream, WithSetup},
+    start_sv2_translator_with_user_identities,
     sv1_sniffer::SV1MessageFilter,
     template_provider::DifficultyLevel,
     utils::get_available_address,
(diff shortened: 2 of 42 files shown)


miner-apps/translator/src/lib/mod.rs
```diff
@@ -449,6 +450,12 @@ impl TranslatorSv2 {
                 .await
                 {
                     Ok(()) => {
+                        let (identity, per_upstream) = match upstream_entry.user_identity.clone() {
+                            Some(id) => (id, true),
+                            None => (self.config.user_identity.clone(), false),
+                        };
+                        sv1_server_instance.set_active_user_identity(identity, per_upstream);
+
```
bit-aloo: I think it’s better to assign an index to the upstream in the sv1_server instance, instead of relying on identity and per_upstream configurations. If the index is invalid, we can fall back to the global user_identity or to the user identity referenced by that index.

GitGab19:
Why would we want to not attach a specific `user_identity` to a specific `upstream` entry?

I'm not convinced about having this as an `Option` tbh.

From the issue and our past discussions it seems very clear that we want users to be aware of the specific `user_identity` to set for any possible fallback upstream in the list.

Also, why are you introducing `Arc<RwLock<String>>` for it?

miner-apps/translator/src/lib/config.rs
```diff
@@ -328,4 +333,30 @@ mod tests {
...
+    #[test]
+    fn test_upstream_user_identity_toml_absent() {
+        let toml = r#"
+            address = "127.0.0.1"
+            port = 4444
+            authority_pubkey = "9bDuixKmZqAJnrmP746n8zU1wyAQRrus7th9dxnkPg6RzQvCnan"
+        "#;
+        let upstream: Upstream = toml::from_str(toml).unwrap();
+        assert_eq!(upstream.user_identity, None);
+    }
```
bit-aloo, line 506: Can you add an example with  multiple upstream?

miner-apps/translator/src/lib/config.rs
```diff
@@ -69,6 +69,10 @@ pub struct Upstream {
     pub port: u16,
     /// The Secp256k1 public key used to authenticate the upstream authority.
     pub authority_pubkey: Secp256k1PublicKey,
+    /// Per-upstream identity sent as-is in `OpenExtendedMiningChannel`; falls back to the
+    /// top-level `user_identity` when absent.
+    #[serde(default)]
+    pub user_identity: Option<String>,
```
bit-aloo: It would be better if we can have user_identity per upstream otherwise it gonna be confusing.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -945,13 +956,16 @@ impl Sv1Server {
...
-        // SRI patterns use `/`-delimited segments for payout mode parsing, so appending
-        // a suffix would break pool-side validation.
-        // See: https://github.com/stratum-mining/sv2-apps/issues/369
-        let user_identity = if self.config.user_identity.starts_with("sri/") {
-            self.config.user_identity.clone()
+        let (base_identity, per_upstream) = self
+            .active_user_identity
+            .read()
+            .expect("active_user_identity lock")
+            .clone();
```
bit-aloo: These can be removed if we follow the indexed access

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -445,6 +445,17 @@ impl ChannelManager {
...
+        *self.user_identity.write().expect("user_identity lock") = identity;
+    }
+
+    fn read_identity(&self) -> String {
+        self.user_identity
+            .read()
+            .expect("user_identity lock")
+            .clone()
+    }
+
```
bit-aloo: We can remove these once we incorporate indexed access.

bit-aloo (commented):
@xyephy I dont see any of the suggested changes? are you yet to push them?

xyephy:
> @xyephy I dont see any of the suggested changes? are you yet to push them?

Hey @Shourya742, the changes are pushed under commits b892940b (tproxy) and 8b831a09 (jdc).

  Addressed:
  - `user_identity`: `Option<String>` per upstream in config for both tproxy and JDC
  - Multi-upstream TOML unit tests added in `config.rs` for both crates
  - Empty string treated as absent (filtered before use)
  - Doc comment on `Arc<RwLock>` explaining why it's needed

Indexed access not implemented because `UpstreamEntry` has a `tried_or_flagged` field that gets mutated as connections are attempted. After the first fallback, the iteration order diverges from `config.upstreams` indices, so there is no stable index to store that reliably maps back to the config array. Storing the resolved identity string at the moment of success via `Arc<RwLock> `avoids this entirely and is the minimal correct solution. Adding `original_index` to `UpstreamEntry` and threading it through all mutation paths would be a larger refactor with no correctness benefit.

The `Arc<RwLock>` stays for the same reason: `Sv1Server` and `ChannelManager` are `Clone`, the identity is unknown at construction, and the write must propagate to already-cloned task instances after the upstream connects.

bit-aloo:
We don't want the user_identity to be an Option, each upstream should be associated to an identity.

bit-aloo:
Arc<RwLock> is not the correct abstraction considering now during the fallback, we reset the states, a oncelock should be sufficient.

xyephy:
let me 

> We don't want the user_identity to be an Option, each upstream should be associated to an identity.

let me update it

xyephy:
@Shourya742 I've pushed updates addressing review comments and also rebased the PR.

miner-apps/translator/src/lib/config.rs
bit-aloo: We need to remove this, considering now each upstream has its own user identity,

miner-apps/translator/src/lib/config.rs
```diff
@@ -74,6 +74,8 @@ pub struct Upstream {
     pub port: u16,
     /// The Secp256k1 public key used to authenticate the upstream authority.
     pub authority_pubkey: Secp256k1PublicKey,
+    #[serde(default)]
```
bit-aloo: This shouldn't be default, we want user to provide user_identity.

miner-apps/translator/src/lib/mod.rs
```diff
@@ -123,9 +123,24 @@ impl TranslatorSv2 {
...
+        let has_missing = upstream_addresses
+            .iter()
+            .any(|u| u.user_identity.is_empty());
+        if has_any_per_upstream && has_missing {
+            warn!(
+                "Some upstream entries have `user_identity` set and some do not. \
+                 Upstreams without a per-upstream identity will use the global `user_identity`, \
+                 which may be rejected by pools expecting a different identity format."
+            );
+        }
```
bit-aloo: I don't think we need any of these extra logic once we make identity per upstream definitive.

miner-apps/jd-client/src/lib/config.rs
```diff
@@ -307,6 +307,8 @@ pub struct Upstream {
     // The network address of the JDS.
     pub jds_address: String,
     pub jds_port: u16,
+    #[serde(default)]
```
bit-aloo: This shouldn't be default, we want user to provide user_identity.

miner-apps/jd-client/src/lib/config.rs
bit-aloo: We shouldn't be needing this anymore.

miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -435,7 +437,9 @@ impl ChannelManager {
             share_batch_size: config.share_batch_size(),
             shares_per_minute: config.shares_per_minute(),
             miner_tag_string: config.jdc_signature().to_string(),
-            user_identity: config.user_identity().to_string(),
+            upstream_index: Arc::new(OnceLock::new()),
+            upstreams: Arc::new(config.upstreams().clone()),
+            global_user_identity: config.user_identity().to_string(),
```
bit-aloo: Considering we are setting the index from the start, we can do the same for `user_identity` as well. As soon as an upstream connects, we can update the user identity in `ChannelManager`, similar to how we handle the index, something like:

```rust
user_identity: Arc<OnceLock<String>>
```

This should work for both JDC and tproxy.

bit-aloo:
@xyephy let me know if you need any further clarification.

bit-aloo:
@xyephy if you can cherry pick this commit 0ec1ca75, I have added the changes requested, for some reason I cannot push to your branch.

xyephy:
> @xyephy if you can cherry pick this commit [551fe67](https://github.com/stratum-mining/sv2-apps/commit/551fe6785769e301864905ad07f101bd5e9e9a30), I have added the changes requested, for some reason I cannot push to your branch.

I was just about to push some updates but let me cherry pick first. Thanks

bit-aloo:
> > @xyephy if you can cherry pick this commit [551fe67](https://github.com/stratum-mining/sv2-apps/commit/551fe6785769e301864905ad07f101bd5e9e9a30), I have added the changes requested, for some reason I cannot push to your branch.
> 
> I was just about to push some updates but let me cherry pick first. Thanks

the commit has changed.

xyephy:
> the commit has changed.

I've seen

xyephy:
running tests after cherry pick, then I push.

xyephy:
@Shourya742 cherry-picked `0ec1ca75`, thanks for the changes. Found two small test issues in the
  commit that were breaking CI:
 - `start_jdc` sets `"IT_TEST"` but `jd_aggregated_topology_monitoring` was still asserting `"IT-test"`  and fixed the label mismatch.
  - `tproxy_sends_per_upstream_identity` and `tproxy_per_upstream_identity_switches_on_fallback` were asserting the identity sent as-is, but your code reintroduces the `.minerN` suffix for non-`sri/` identities. I've updated the test identities and assertions.

bit-aloo:
@xyephy can you squash the commits, so its easy to review for others.

integration-tests/lib/mod.rs
```diff
@@ -207,11 +207,31 @@ pub fn start_jdc(
...
+        &pool_with_ids,
+        template_provider_config,
+        supported_extensions,
+        required_extensions,
+        enable_monitoring,
+        jdc_mode,
+    )
+}
+
+pub fn start_jdc_with_identities(
```
GitGab19:
```suggestion
pub fn start_jdc_with_user_identities(
```

integration-tests/lib/mod.rs
```diff
@@ -349,19 +361,91 @@ pub async fn start_sv2_translator(
...
         required_extensions,
         job_keepalive_interval_secs,
-        "user_identity".to_string(),
-        false,
         enable_monitoring,
     )
     .await
 }
 
+pub async fn start_sv2_translator_with_identities(
```
GitGab19:
```suggestion
pub async fn start_sv2_translator_with_user_identities(
```

integration-tests/lib/mod.rs
```diff
@@ -349,19 +361,91 @@ pub async fn start_sv2_translator(
...
+    let translator_v2 = translator_sv2::TranslatorSv2::new(config);
+    let clone_translator_v2 = translator_v2.clone();
+    tokio::spawn(async move {
+        clone_translator_v2.start().await;
+    });
+    (translator_v2, listening_address, monitoring_address)
+}
+
 #[allow(clippy::too_many_arguments)]
 pub async fn start_sv2_translator_with_user_identity(
```
GitGab19, line 451: Do we still need this one?

integration-tests/lib/mod.rs
```diff
@@ -207,11 +207,31 @@ pub fn start_jdc(
...
+        template_provider_config,
+        supported_extensions,
+        required_extensions,
+        enable_monitoring,
+        jdc_mode,
+    )
+}
+
+pub fn start_jdc_with_identities(
+    pool: &[(SocketAddr, SocketAddr, String)], // (pool_address, jds_address, per_upstream_id)
```
GitGab19:
```suggestion
    pool: &[(SocketAddr, SocketAddr, String)], // (pool_address, jds_address, per_upstream_user_identity)
```

integration-tests/tests/jdc_identity.rs
```diff
@@ -0,0 +1,169 @@
...
+};
+use stratum_apps::stratum_core::{
+    common_messages_sv2::{SetupConnectionSuccess, MESSAGE_TYPE_SETUP_CONNECTION},
+    job_declaration_sv2::MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN,
+    mining_sv2::MESSAGE_TYPE_OPEN_EXTENDED_MINING_CHANNEL,
+    parsers_sv2::{AnyMessage, CommonMessages, JobDeclaration, Mining},
+};
+
+// Verifies that when a per-upstream `user_identity` is set on a JDC upstream entry,
+// that identity is sent as-is in `OpenExtendedMiningChannel` — no `.minerN` suffix appended.
```
GitGab19:
```suggestion
// that identity is sent as-is in `OpenExtendedMiningChannel`.
```

integration-tests/tests/jdc_identity.rs
```diff
@@ -0,0 +1,169 @@
...
+        "expected per-upstream identity '{PER_UPSTREAM_IDENTITY}', got '{identity_str}'"
+    );
+
+    shutdown_all!(translator, jdc);
+}
+
+// Verifies that when the primary pool disconnects, JDC falls back to the next upstream entry
+// and sends *that* upstream's `user_identity` in `AllocateMiningJobToken` — not the primary's.
+// All three JDC identity reads (OEMC, ExtendedChannel, AllocateMiningJobToken) share a single
+// `read_identity()` source, so asserting one path is sufficient.
```
GitGab19: What does this sentence mean?

And why not asserting also the `user_identity` value in the `OpenExtendedMiningChannel` messages?

integration-tests/tests/jdc_identity.rs
GitGab19, line 1: Do we really need a specific new file for these two tests?

integration-tests/tests/translator_integration.rs
```diff
@@ -2383,3 +2384,159 @@ async fn test_translator_fallback_during_abrupt_disconnection() {
...
+async fn tproxy_sends_per_upstream_identity() {
+    start_tracing();
+
+    let mock_upstream_addr = get_available_address();
+    let mock_upstream = MockUpstream::new(mock_upstream_addr, WithSetup::no());
+    let send_to_tproxy = mock_upstream.start().await;
+
+    let (sniffer, sniffer_addr) = start_sniffer("", mock_upstream_addr, false, vec![], None);
+
+    const PER_UPSTREAM_IDENTITY: &str = "sri/solo/bc1qtest.worker";
```
GitGab19:
```suggestion
    const PER_UPSTREAM_IDENTITY: &str = "bc1qtest.worker";
```

integration-tests/tests/translator_integration.rs
```diff
@@ -2383,3 +2384,159 @@ async fn test_translator_fallback_during_abrupt_disconnection() {
         .await;
     shutdown_all!(translator, pool_2);
 }
+
+#[tokio::test]
+async fn tproxy_sends_per_upstream_identity() {
```
GitGab19:
```suggestion
async fn tproxy_sends_per_upstream_user_identity() {
```

integration-tests/tests/translator_integration.rs
```diff
@@ -2383,3 +2384,159 @@ async fn test_translator_fallback_during_abrupt_disconnection() {
...
+    assert_eq!(
+        identity_str, PER_UPSTREAM_IDENTITY,
+        "expected per-upstream identity '{PER_UPSTREAM_IDENTITY}', got '{identity_str}'"
+    );
+
+    shutdown_all!(translator);
+}
+
+#[tokio::test]
+async fn tproxy_per_upstream_identity_switches_on_fallback() {
```
GitGab19:
```suggestion
async fn tproxy_per_upstream_user_identity_switches_on_fallback() {
```

integration-tests/tests/translator_integration.rs
```diff
@@ -2383,3 +2384,159 @@ async fn test_translator_fallback_during_abrupt_disconnection() {
...
+        vec![],
+        vec![],
+        None,
+        false,
+    )
+    .await;
+
+    let (_minerd, _) = start_minerd(tproxy_addr, None, None, false).await;
+
+    // Primary must fail.
```
GitGab19: Explain why it should fail.

Also, what's the purpose of running a real pool here, instead of a mock?

miner-apps/jd-client/Cargo.toml
```diff
@@ -27,6 +27,9 @@ bitcoin_core_sv2 = { version = "0.3.0", path = "../../bitcoin-core-sv2" }
 hex = "0.4.3"
 hotpath = "0.14.0"
 
+[dev-dependencies]
+toml = "0.8"
```
GitGab19: Why the 0.8 version?

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
```diff
@@ -189,6 +189,18 @@ impl ChannelManager {
...
+            .get()
+            .expect("payout mode should be set")
+    }
+
+    pub(crate) fn set_expected_payout_distribution(&self, payout_mode: Option<PayoutMode>) {
+        self.expected_payout_distribution
+            .set(payout_mode)
+            .expect("paymode should be set only once");
+    }
+
```
GitGab19, line 203: How are these changes related to the main goal of this PR?

miner-apps/translator/src/lib/mod.rs
```diff
@@ -75,6 +76,24 @@ impl TranslatorSv2 {
...
+            })?;
+        if let Some(distribution) = &expected_payout_distribution {
+            info!(
+                "Payout verification enabled for configured user_identity: {}",
+                distribution
+            );
+        }
+        Ok(expected_payout_distribution)
+    }
+
```
GitGab19, line 96: How are these changes related to the main goal of this PR?

bit-aloo: This was added by me. Since payout distribution is a function of user identity, and this PR changes user identity to be maintained on a per-upstream basis, this logic was required. Earlier, payout distribution depended on a global user_identity. During fallback, we switch the upstream, which also changes the user_identity, and accordingly updates the payout distribution.

miner-apps/jd-client/Cargo.toml
bit-aloo: The current latest is 1.1.2  on crates.io, `toml = "1.1.2"`

integration-tests/lib/mod.rs
xyephy: Yes, it exposes `verify_payout`

integration-tests/tests/translator_integration.rs
xyephy: `sniffer_1` intercepts the primary mock's `SetupConnectionSuccess` and replaces it with `SetupConnectionError`, forcing tproxy to fall over to `mock_fallback`.

Switched both upstreams to `MockUpstream`, assertions happen on the sniffers so no real pool is needed. The primary mock only needs to complete `SetupConnection` so `sniffer_1` has a message to intercept. The fallback mock completes `SetupConnection` so `sniffer_2` can capture the `OpenExtendedMiningChannel` and assert the correct `user_identity`.

miner-apps/translator/Cargo.toml
```diff
@@ -42,3 +42,4 @@ hotpath-alloc = ["hotpath", "hotpath/hotpath-alloc"]
 
 [dev-dependencies]
 sha2 = "0.10.6"
+toml = "0.8"
```
GitGab19: Please, align the versions between JDC and tProxy

integration-tests/lib/mod.rs
GitGab19: We can move the `verify_payout` into the new `start_sv2_translator_with_user_identities` and remove this older function.

integration-tests/tests/translator_integration.rs
```diff
@@ -2383,3 +2384,179 @@ async fn test_translator_fallback_during_abrupt_disconnection() {
...
+    let setup_err = ReplaceMessage::new(
+        MessageDirection::ToDownstream,
+        MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+        AnyMessage::Common(parsers_sv2::CommonMessages::SetupConnectionError(
+            SetupConnectionError {
+                flags: 0,
+                error_code: "test-identity-fallback".to_string().try_into().unwrap(),
+            },
+        )),
+    );
```
GitGab19: There's no need for replacing any message here.

Since you're using mocks, you can directly run the mock and make it send the error, with a logic similar to this:
```
let send_to_tproxy = MockUpstream::new(mock_primary_addr, WithSetup::no())
    .start()
    .await;

sniffer_1
    .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
    .await;

send_to_tproxy
    .send(AnyMessage::Common(
        parsers_sv2::CommonMessages::SetupConnectionError(SetupConnectionError {
            flags: 0,
            error_code: "test-identity-fallback".to_string().try_into().unwrap(),
        }),
    ))
    .await
    .unwrap();
```

integration-tests/tests/translator_integration.rs
```diff
@@ -2383,3 +2384,179 @@ async fn test_translator_fallback_during_abrupt_disconnection() {
...
+    .await;
+    let mock_fallback_addr = get_available_address();
+    let _mock_fallback = MockUpstream::new(
+        mock_fallback_addr,
+        WithSetup::yes_with_defaults(Protocol::MiningProtocol, 0),
+    )
+    .start()
+    .await;
+
+    const PRIMARY_IDENTITY: &str = "sri/solo/bc1qprimary";
```
GitGab19:
```suggestion
    const PRIMARY_IDENTITY: &str = "bc1qprimary.worker";
```

GitGab19:
@xyephy there are still a couple of minor things to be improved, as you can see from my new comments.

In addition to those, please squash the last commit ([`integration-tests: address review feedback`](https://github.com/stratum-mining/sv2-apps/pull/501/changes/61fcf7e14b68e3d3ff18e77286bea73a35ecfd8c)) in the previous one.

xyephy:
@GitGab19 Done.

GitGab19 (approved):
tACK

bit-aloo:
ACK

PR #521, opened 2026-05-23
fix broken `docs.rs`

### LINKED ISSUE #519: some `docs.rs` builds are failing
| docs.rs                                      | status |
|----------------------------------------------|--------|
| https://docs.rs/bitcoin_core_sv2/            | ✅      |
| https://docs.rs/pool_sv2/                    | ❌      |
| https://docs.rs/jd_server_sv2/         | ✅      |
| https://docs.rs/jd_client_sv2/         | ❌      |
| https://docs.rs/translator_sv2/        | ❌      |
| https://docs.rs/integration_tests_sv2/ | ❌      |
| https://docs.rs/stratum-apps/          | ✅      |

plebhash:
close #519

fixes the issues around sandboxed builds of `utoipa-swagger-ui` (as described in #519)

also standardizes/fixes `documentation` field of some `Cargo.toml` files`


stratum-apps/Cargo.toml
bit-aloo: Can you add monitoring here, stratum-apps doesn't embed monitoring in its doc

bit-aloo (commented):
This is ok, having a vendored variant, considering we anyway download the swagger ui zip and embed it later in non-vendored flow.

bit-aloo:
Can you squash this into two commits: one for the correct documentation pointing, and the other for changing the dependency.

bit-aloo:
We can add a CI workflow building docs.

plebhash:
> We can add a CI workflow building docs.

added `docs-rs.yaml` + `docs-rs-check.sh` via d98a7ec00418687ffa0839a954d0871cb7334465

together they do a few deliberate things to closely emulate `docs.rs` build process:
- runs with `nightly` toolchain
- builds docs via [`cargo-docs-rs`](https://crates.io/crates/cargo-docs-rs) instead of `cargo doc`
- uses `unshare --net` to emulate a no-network sandboxed environment

I did contemplate leveraging [`rustwide`](https://github.com/rust-lang/rustwide) + [`crates-build-env`](https://github.com/rust-lang/crates-build-env) to more closely emulate the `docs.rs` build process, but it would make CI much slower and complex to maintain, so I think the trade-off doesn't really justify and the current approach should be sufficient.

---

I validated it by reverting 7a8b0c95773e8cdc600e94c51cd9aa46e518174b, which led to https://github.com/stratum-mining/sv2-apps/actions/runs/26417294759/job/77764416386?pr=521, which is exactly the same error message that was happening on docs.rs

stratum-apps/Cargo.toml
plebhash: done

GitGab19 (approved):

scripts/docs-rs-check.sh
```diff
@@ -0,0 +1,38 @@
...
+manifest_path="$1"
+
+if [ ! -f "$manifest_path" ]; then
+  echo "Manifest not found: $manifest_path" >&2
+  exit 1
+fi
+
+# Prefetch before disabling networking so all needed crates are already cached.
+echo "Prefetching dependencies for ${manifest_path}"
+cargo fetch --manifest-path="$manifest_path"
```
bit-aloo:
```suggestion
cargo +nightly fetch --manifest-path="$manifest_path"
```

scripts/docs-rs-check.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+# Execute docs build inside a separate network namespace to emulate docs.rs
+# no-network constraints. Preserve Rust/cargo env so cargo remains available
+# inside the sudo context.
+echo "Building docs for ${manifest_path} in no-network sandbox"
+export MANIFEST_PATH="$manifest_path"
+export CARGO_BIN="$cargo_bin"
+sudo --preserve-env=MANIFEST_PATH,CARGO_BIN,PATH,CARGO_HOME,RUSTUP_HOME,HOME unshare --net --mount-proc bash -lc \
+  'ip link set lo up; "$CARGO_BIN" +nightly docs-rs --manifest-path="$MANIFEST_PATH" --offline'
```
bit-aloo: How slow is using rust-wide compared to this? considering we are using sudo here (sudo police alert!!!)...

scripts/docs-rs-check.sh
plebhash: done

scripts/docs-rs-check.sh
plebhash: `gpt-5.3-codex` estimated it could take up to 60+ min longer

but I get the `sudo` concern, so let's see in practice, pushing an alternative `docs-rs-rustwide.yaml`

plebhash: https://github.com/stratum-mining/sv2-apps/actions/runs/26449910856/job/77867420130?pr=521

plebhash: 9min via `rustwide` vs 4min without `rustwide`, much faster than what the agent had predicted... I guess that's acceptable

the main tradeoff now is just complexity, which is also something I'm willing to live with

I'll iterate on `docs-rs-rustwide.yaml` with extra docs and renaming stuff (`rustwide` seems to be used indirectly, so it's probably better to make terminology revolve around `docs_rs_builder`)

plebhash: hmm actually I reverted 56a679eaba19f8efe6cae6871f2cf2f3a088b164 and `docs-rs-rustwide.yaml` is still passing, which means the current implementation is not sane:

https://github.com/stratum-mining/sv2-apps/actions/runs/26455295753/job/77887412692?pr=521

tbh this is turning into a timesink

while I do see the merits of enforcing sanity of `docs.rs` via CI, I'd prefer that the scope of this PR was limited to patching the root cause (which was my original intention)

I created https://github.com/stratum-mining/sv2-apps/issues/524 so this can be addressed on the long term

removing `docs-rs-*.yaml` and `docs-rs-check.sh`

PR #530, opened 2026-05-26
Migrate translator to shared synchronous primitive

### LINKED ISSUE #601: Refactor tProxy to reduce nested locking
similar to #205 and #206

bit-aloo:
closes: #601

### DIFF
--- a/miner-apps/translator/src/lib/error.rs
+++ b/miner-apps/translator/src/lib/error.rs
@@ -371,6 +371,16 @@ impl From<stratum_apps::stratum_core::sv1_api::error::Error> for TproxyErrorKind
     }
 }
 
+impl From<stratum_apps::stratum_core::sv1_api::error::Error> for TproxyError<Sv1Server> {
+    fn from(_: stratum_apps::stratum_core::sv1_api::error::Error) -> Self {
+        Self {
+            kind: TproxyErrorKind::SV1Error,
+            action: Action::Log,
+            _owner: PhantomData,
+        }
+    }
+}
+
 impl From<stratum_apps::network_helpers::Error> for TproxyErrorKind {
     fn from(value: stratum_apps::network_helpers::Error) -> Self {
         TproxyErrorKind::NetworkHelpersError(value)
--- a/miner-apps/translator/src/lib/error.rs
+++ b/miner-apps/translator/src/lib/error.rs
@@ -22,7 +22,8 @@ use stratum_apps::{
         handlers_sv2::HandlerErrorType,
         noise_sv2,
         parsers_sv2::{self, ParserError, TlvError},
-        sv1_api::server_to_client::SetDifficulty,
+        stratum_translation,
+        sv1_api::{self, server_to_client::SetDifficulty},
     },
     utils::types::{
         CanDisconnect, CanFallback, CanShutdown, ChannelId, DownstreamId, ExtensionType,
@@ -100,6 +101,18 @@ where
             _owner: PhantomData,
         }
     }
+
+    /// Attaches a downstream to an error built without connection context, such as the sv1 errors
+    /// `From` produces for the `IsServer::Error` bound. Used on the `handle_message` call sites in
+    /// `Sv1Server`, where a protocol violation should drop that miner instead of only being logged.
+    /// Actions chosen explicitly (fallback, shutdown) are left alone.
+    pub fn with_sv1_downstream_context(mut self, downstream_id: DownstreamId) -> Self {
+        if matches!(&self.kind, TproxyErrorKind::SV1Error(_)) && matches!(self.action, Action::Log)
+        {
+            self.action = Action::Disconnect(downstream_id);
+        }
+        self
+    }
 }
 
 impl<O> TproxyError<O>
@@ -137,7 +150,7 @@ impl<Owner> From<TproxyError<Owner>> for TproxyErrorKind {
 #[derive(Debug)]
 pub enum TproxyErrorKind {
     /// Generic SV1 protocol error
-    SV1Error,
+    SV1Error(sv1_api::error::Error),
     /// Error from the network helpers library
     NetworkHelpersError(stratum_apps::network_helpers::Error),
     /// Error from roles logic parser library
@@ -264,7 +277,7 @@ impl fmt::Display for TproxyErrorKind {
                     "Server requires extensions that we don't support: {extensions:?}"
                 )
             }
-            SV1Error => write!(f, "Sv1 error"),
+            SV1Error(e) => write!(f, "Sv1 error: `{e:?}`"),
             TranslatorCore(e) => write!(f, "Translator core error: {e:?}"),
             NetworkHelpersError(e) => write!(f, "Network helpers error: {e:?}"),
             ParserError(e) => write!(f, "Roles logic parser error: {e:?}"),
@@ -365,19 +378,15 @@ impl From<SetDifficulty> for TproxyErrorKind {
     }
 }
 
-impl From<stratum_apps::stratum_core::sv1_api::error::Error> for TproxyErrorKind {
-    fn from(_: stratum_apps::stratum_core::sv1_api::error::Error) -> Self {
-        TproxyErrorKind::SV1Error
+impl From<sv1_api::error::Error> for TproxyErrorKind {
+    fn from(e: sv1_api::error::Error) -> Self {
+        TproxyErrorKind::SV1Error(e)
     }
 }
 
-impl From<stratum_apps::stratum_core::sv1_api::error::Error> for TproxyError<Sv1Server> {
-    fn from(_: stratum_apps::stratum_core::sv1_api::error::Error) -> Self {
-        Self {
-            kind: TproxyErrorKind::SV1Error,
-            action: Action::Log,
-            _owner: PhantomData,
-        }
+impl From<sv1_api::error::Error> for TproxyError<Sv1Server> {
+    fn from(e: sv1_api::error::Error) -> Self {
+        Self::log(e)
     }
 }
 
@@ -387,12 +396,8 @@ impl From<stratum_apps::network_helpers::Error> for TproxyErrorKind {
     }
 }
 
-impl From<stratum_apps::stratum_core::stratum_translation::error::StratumTranslationError>
-    for TproxyErrorKind
-{
-    fn from(
-        e: stratum_apps::stratum_core::stratum_translation::error::StratumTranslationError,
-    ) -> Self {
+impl From<stratum_translation::error::StratumTranslationError> for TproxyErrorKind {
+    fn from(e: stratum_translation::error::StratumTranslationError) -> Self {
         TproxyErrorKind::TranslatorCore(e)
     }
 }
--- a/miner-apps/translator/src/lib/sv1/monitoring.rs
+++ b/miner-apps/translator/src/lib/sv1/monitoring.rs
@@ -271,15 +271,15 @@ impl Sv1Server {
     }
 
     fn current_downstream_sv1_usernames(&self) -> Vec<(DownstreamId, String)> {
-        let mut worker_names = Vec::new();
+        let mut usernames = Vec::new();
         self.downstreams.for_each(|downstream_id, downstream| {
             if let Ok(worker_name) = downstream
                 .downstream_data
                 .with(|data| data.sv1_username.clone())
             {
-                worker_names.push((downstream_id, worker_name));
+                usernames.push((downstream_id, worker_name));
             }
         });
-        worker_names
+        usernames
     }
 }
--- a/miner-apps/translator/src/lib/sv1/monitoring.rs
+++ b/miner-apps/translator/src/lib/sv1/monitoring.rs
@@ -31,7 +31,7 @@ fn downstream_to_sv1_client_info(
 ) -> Option<Sv1ClientInfo> {
     downstream
         .downstream_data
-        .safe_lock(|dd| Sv1ClientInfo {
+        .with(|dd| Sv1ClientInfo {
             client_id: downstream.downstream_id,
             channel_id: dd.channel_id,
             connection_ip: dd.connection_ip,
@@ -59,34 +59,37 @@ fn downstream_to_sv1_client_info(
 
 impl Sv1ClientsMonitoring for Sv1Server {
     fn get_sv1_clients(&self) -> Vec<Sv1ClientInfo> {
-        self.downstreams
-            .iter()
-            .filter_map(|downstream| {
-                let miner_telemetry = self.miner_telemetry_for(*downstream.key());
-                let management_ip = self.miner_telemetry_management_ip_for(*downstream.key());
-                let miner_telemetry_status = self.miner_telemetry_status_for(*downstream.key());
-                downstream_to_sv1_client_info(
-                    downstream.value(),
-                    miner_telemetry,
-                    management_ip,
-                    miner_telemetry_status,
-                )
-            })
-            .collect()
+        let mut clients = Vec::new();
+        self.downstreams.for_each(|downstream_id, downstream| {
+            let miner_telemetry = self.miner_telemetry_for(downstream_id);
+            let management_ip = self.miner_telemetry_management_ip_for(downstream_id);
+            let miner_telemetry_status = self.miner_telemetry_status_for(downstream_id);
+            if let Some(client) = downstream_to_sv1_client_info(
+                downstream,
+                miner_telemetry,
+                management_ip,
+                miner_telemetry_status,
+            ) {
+                clients.push(client);
+            }
+        });
+        clients
     }
 
     fn get_sv1_client_by_id(&self, client_id: usize) -> Option<Sv1ClientInfo> {
         let miner_telemetry = self.miner_telemetry_for(client_id);
         let management_ip = self.miner_telemetry_management_ip_for(client_id);
         let miner_telemetry_status = self.miner_telemetry_status_for(client_id);
-        self.downstreams.get(&client_id).and_then(|downstream| {
-            downstream_to_sv1_client_info(
-                downstream.value(),
-                miner_telemetry,
-                management_ip,
-                miner_telemetry_status,
-            )
-        })
+        self.downstreams
+            .with(&client_id, |downstream| {
+                downstream_to_sv1_client_info(
+                    downstream,
+                    miner_telemetry,
+                    management_ip,
+                    miner_telemetry_status,
+                )
+            })
+            .flatten()
     }
 }
 
@@ -268,15 +271,15 @@ impl Sv1Server {
     }
 
     fn current_downstream_sv1_usernames(&self) -> Vec<(DownstreamId, String)> {
-        self.downstreams
-            .iter()
-            .filter_map(|downstream| {
-                downstream
-                    .value()
-                    .downstream_data
-                    .safe_lock(|data| (*downstream.key(), data.sv1_username.clone()))
-                    .ok()
-            })
-            .collect()
+        let mut worker_names = Vec::new();
+        self.downstreams.for_each(|downstream_id, downstream| {
+            if let Ok(worker_name) = downstream
+                .downstream_data
+                .with(|data| data.sv1_username.clone())
+            {
+                worker_names.push((downstream_id, worker_name));
+            }
+        });
+        worker_names
     }
 }
--- a/miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
@@ -319,7 +319,7 @@ impl Sv1Server {
             });
 
             total_hashrate += hashrate;
-            Ok(())
+            Ok::<(), TproxyError<error::Sv1Server>>(())
         })?;
 
         let Some(min_target) = min_target else {
@@ -608,7 +608,7 @@ impl Sv1Server {
                     Some(current) => current.min(target),
                     None => target,
                 });
-                Ok(())
+                Ok::<(), TproxyError<error::Sv1Server>>(())
             })?;
 
             let Some(min_target) = min_target else {
--- a/miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
@@ -13,7 +13,7 @@ use stratum_apps::{
     stratum_core::{
         bitcoin::Target,
         channels_sv2::{Vardiff, target::hash_rate_to_target},
-        mining_sv2::SetTargetOwned as SetTarget,
+        mining_sv2::SetTargetOwned,
         stratum_translation::sv2_to_sv1::{
             build_sv1_set_difficulty_from_sv2_target_with_integer_power_of_two_rounding,
             sv1_advertised_target_from_sv2_target,
@@ -150,7 +150,7 @@ impl Sv1Server {
                                 // Case 1: new_target >= upstream_target, send set_difficulty
                                 // immediately
                                 trace!(
-                                    "✅ Target comparison: new_target ({}) >= upstream_target ({}) for downstream {}, will send set_difficulty immediately",
+                                    "✅ Target comparison: new_target ({}) >= upstream_target ({}) for downstream {}, will send mining.set_difficulty immediately",
                                     new_target, upstream_target, downstream_id
                                 );
                                 immediate_updates.push((
@@ -162,7 +162,7 @@ impl Sv1Server {
                                 // Case 2: new_target < upstream_target, delay set_difficulty until
                                 // SetTarget
                                 trace!(
-                                    "⏳ Target comparison: new_target ({}) < upstream_target ({}) for downstream {}, will delay set_difficulty until SetTarget",
+                                    "⏳ Target comparison: new_target ({}) < upstream_target ({}) for downstream {}, will delay mining.set_difficulty until SetTarget",
                                     new_target, upstream_target, downstream_id
                                 );
                                 self.pending_target_updates
@@ -178,7 +178,7 @@ impl Sv1Server {
                         None => {
                             // No upstream target set yet, send set_difficulty immediately as fallback
                             trace!(
-                                "No upstream target set for downstream {}, will send set_difficulty immediately",
+                                "No upstream target set for downstream {}, will send mining.set_difficulty immediately",
                                 downstream_id
                             );
                             immediate_updates.push((channel_id, Some(downstream_id), new_target));
@@ -226,7 +226,7 @@ impl Sv1Server {
                     Ok(message) => message,
                     Err(e) => {
                         error!(
-                            "Failed to build immediate SetDifficulty for downstream {downstream_id}: {e:?}; skipping"
+                            "Failed to build immediate mining.set_difficulty for downstream {downstream_id}: {e:?}; skipping"
                         );
                         continue;
                     }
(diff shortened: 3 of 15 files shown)


miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
```diff
@@ -185,33 +192,37 @@ impl IsServer<'static> for Sv1Server {
...
         let downstream_id = client_id.expect("Downstream id should exist");
-        let Some(downstream) = self.downstreams.get(&downstream_id) else {
+        let Some(downstream) = self.downstreams.get_cloned(&downstream_id) else {
             return Err(Error::UnknownID(downstream_id as u64));
         };
         let is_authorized = downstream
             .downstream_data
-            .super_safe_lock(|data| data.authorized_worker_name == *name);
+            .with(|data| data.authorized_worker_name == *name)
+            .unwrap();
```
bit-aloo: Can't propagate error from Sv1Server downstream message handler, considering it doesn't support TproxyError type.

bit-aloo: Moved forward with UnknownId variant for now, we can extend error enum in Sv1 crate later to include some poison related variant.

GitGab19:
Is this aiming to close https://github.com/stratum-mining/sv2-apps/issues/601?

bit-aloo:
> Is this aiming to close #601?

yup updated the description

GitGab19:
Can you solve conflicts?

bit-aloo:
> Can you solve conflicts?

Should be good now

GitGab19:
Preliminary clanker findings:
```
## Issues

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 High | `difficulty_manager.rs`: a failed `set_difficulty` send to **one** downstream is now `TproxyError::shutdown` → cancels the global token → **whole proxy exits**. Trigger is a routine race: miner disconnects between sender clone and `send().await`. Was log-and-continue before. |
| 2 | 🔴 High | Per-downstream `hash_rate_to_target` failures (3 sites) and `try_vardiff` errors now also escalate to global shutdown; previously `continue`/ignored. One pathological miner can take down the proxy. |
| 3 | 🟡 Medium | `vardiff_future` select branch: if `handle_error_action` ever returns `Continue`, the completed future is re-polled → panic. Currently unreachable only because all escaping errors are shutdown-kind and `biased;` breaks first on cancelled tokens. Set `vardiff_enabled = false` after an error. |
| 4 | 🟢 Low | Keepalive path: `get_mut` → `with_mut_or_default` can re-create a `valid_sv1_jobs` entry for a dead channel (never cleaned up → small leak). Use `with_mut`. |
| 5 | ⚪ Nit | `Error::UnexpectedMessage("poisoned downstream data lock")` misuses a protocol error variant (×8 — extract a helper); `contains_key` + `with_mut` TOCTOU pairs in `mining_message_handler.rs` (benign today, fragile); `handle_downstream_message` now does 3–4 map lookups per SV1 message; vardiff iteration holds shard *write* guards during computation (cheap, no awaits — acceptable). |
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -57,148 +59,198 @@ impl Sv1Server {
...
-            let Some(downstream) = self.downstreams.get(downstream_id) else {
-                continue;
+            let (channel_id, hashrate, target, upstream_target) = match self
+                .with_registered_downstream(downstream_id, |downstream| {
+                    downstream
+                        .downstream_data
+                        .with(|data| {
+                            // It's safe to unwrap hashrate because we know that
+                            // the downstream has a hashrate (we are
+                            // doing vardiff)
```
GitGab19, line 75: How do you know that we are doing vardiff? Vardiff can be disabled from the config.

miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
GitGab19: Why not tackling that now, by doing the changes on `sv1_api`?

GitGab19:
Second pass of clanker review:

## Review — updated revision (b12b08fb)

The update addresses all substantive issues from my previous review. Two remaining
requests, plus nits:

### 1. Latent panic: completed vardiff/keepalive futures can be re-polled

`spawn_vardiff_loop` / `spawn_job_keepalive_loop` only exit by returning `Err`, so
when their select branch fires the future is complete. The `vardiff_enabled = false`
disarm from the previous revision is gone, so if `handle_error_action` ever returns
`LoopControl::Continue`, the next iteration re-polls the completed future →
panic ("async fn resumed after completion"), killing the task without `cleanup()`.

Today this is unreachable — `biased;` ordering covers the token-cancelled paths, and
all errors escaping these loops happen to be Break-kind — but nothing enforces that.
A future `Log`/`Disconnect`-kind error propagating out of either loop becomes a
process-killing panic.

Suggestion: on `Continue`, re-arm with `vardiff_future.set(self.clone().spawn_vardiff_loop())`
(also restores the service instead of leaving it dead). At minimum, comment the
Break-only invariant on both loops.

### 2. sv1_api error handling in `downstream_message_handler.rs` — fix at the source

`IsServer` (sv1_api crate) hardcodes its error type to `sv1_api::Error` — unlike the
SV2 handler traits, there is no associated error type — which is what forces the
`Error::UnexpectedMessage("poisoned downstream data lock")` workaround repeated ~10×
(a poisoned lock is an internal failure, not a protocol error; the variant is only
"least wrong"). `Error::UnknownID` for "downstream not registered" is correct and
stays as is.

Rather than shipping the workaround, let's fix this properly now:

- Open a PR against stratum-core giving `IsServer` an associated
  `type Error: From<sv1_api::Error>` (mirroring the SV2 handler traits). An additive
  `Error::Internal(String)` variant would also work, but since the enum is not
  `#[non_exhaustive]` either change is breaking — if we're bumping anyway, the
  associated type is the right spend and makes sv1 consistent with sv2.
- Bump the dependency here (a git dependency on the stratum-core branch is fine
  while the release is pending) and have these handlers return `TproxyError`
  directly, with `TproxyErrorKind::PoisonLock` / `DownstreamNotFound` and proper
  actions — deleting the `UnexpectedMessage` mapping entirely.
- Independently: hoist the two panics out of lock closures
  (`VersionRollingParams::new(...).expect(...)` in `handle_configure`,
  `Notify::try_from(...).expect(...)` in `downstream.rs`). A panic inside a closure
  is the realistic way these locks get poisoned in the first place.

If the sv1_api change lands, the duplicated `.map_err` closures and the
`with_registered_downstream_sv1` / `with_registered_downstream` split collapse
naturally.

### Nits (non-blocking)

- The "first group channel" lookup via `for_each` + `get_or_insert` (3× in
  `mining_message_handler.rs`) scans the whole map with no early exit and silently
  picks an arbitrary entry if the single-group-channel assumption is ever violated.
  Consider a `first_key()` helper + `warn!` when `len() > 1`.
- The nested `with_mut(...).ok_or(...)??` chains in the group-channel branches are
  correct but dense; extracting per-channel bodies into named methods would help.

### Tests

No coverage for the new error plumbing. Most valuable: force `handle_vardiff_updates`
to fail and assert clean shutdown (would also catch a re-poll regression), and pin
the "share dropped with `Ok(())` when downstream disconnects mid-submission" behavior
this revision just fixed.

---

As I asked [here](https://github.com/stratum-mining/sv2-apps/pull/530#discussion_r3704826681) I would prefer to make the changes on `sv1_api` in order to proceed in a clean way with this PR, wdyt @bit-aloo ?

miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
bit-aloo: This would require adding an error variant to `sv1_api`, which I can do. But it wouldn't really be specific to the library we would basically be adding a poison variant just because our downstream crate needs it.

Initially, I was thinking about having an associated error type, similar to what we do in handlers. But I need to think about it a bit more and consider how the APIs around this might evolve.

bit-aloo: Or maybe I can have an universal error mapping in sv1 error enum. Maybe that could do the trick.

bit-aloo: Blocked by: https://github.com/stratum-mining/sv2-apps/pull/664, I will pull this PR back to draft and wait for the assoc error type PR to get merged.

bit-aloo:
Both of the issues mentioned above by Clanker aren't applicable.

For the first one, it's saying that if the vardiff future ever returns an error variant like `Log` or `Disconnect` in the future, we would panic. But that shouldn't happen, since the only error vardiff currently propagates is `Shutdown`.

The second one is about the error type on the server trait, which is now fixed with the recent changes to the SV1 server trait and its error variant.

GitGab19:
The second point is what led us to work on the `sv1_api` crate, so it's now solved.

About the first point, it's not saying it's panicking today, but it's saying that if in the future we change anything and we end up in a situation where the `handle_error_action` returns `Continue`, we are gonna panic.

That's why it's suggesting a guard there, to be safe for any future change we're gonna have on tProxy.

Here the diff suggested:
```
                     res = &mut vardiff_future, if vardiff_enabled => {
                         if let Err(e) = res {
                             if let LoopControl::Break = self.handle_error_action(
                                 "Sv1Server::spawn_vardiff_loop",
                                 &e,
                                 &cancellation_token,
                                 &fallback_token,
                             ).await {
                                 self.cleanup();
                                 break;
                             }
                         }
+                        // The future just completed; re-arm it so the select
+                        // never polls a completed future again (which panics).
+                        vardiff_future.set(self.clone().spawn_vardiff_loop());
                     }
```
```
                     res = &mut keepalive_future, if keepalive_enabled => {
                         if let Err(e) = res {
                             if let LoopControl::Break = self.handle_error_action(
                                 "Sv1Server::spawn_job_keepalive_loop",
                                 &e,
                                 &cancellation_token,
                                 &fallback_token,
                             ).await {
                                 self.cleanup();
                                 break;
                             }
                         }
+                        // The future just completed; re-arm it so the select
+                        // never polls a completed future again (which panics).
+                        keepalive_future.set(self.clone().spawn_job_keepalive_loop());
                     }
```

Wdyt?

bit-aloo:
I do get the point, but I would prefer to continue with the invariant we have. Even in the future, if we introduce some other variant that could lead to re-polling a completed future, we would catch the panic in the integration tests themselves. Also, I don’t think we plan to change how we operate in vardiff, especially around errors. The only reason we return errors right now is because of lock poisoning, and nothing else.

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -1,17 +1,19 @@
...
+        Sv1Server,
+        sv1_server::{PendingTargetUpdate, SV1_MIN_DIFFICULTY_FOR_INTEGER_POWER_OF_TWO_ROUNDING},
+    },
 };
 
 use stratum_apps::{
     stratum_core::{
         bitcoin::Target,
         channels_sv2::{Vardiff, target::hash_rate_to_target},
+        mining_sv2::SetTargetOwned as SetTarget,
```
GitGab19: Why this change? I feel that it would be better to keep coherence across our codebase. 

If you use aliases here, I think we should use aliases everywhere else as well.

Maybe we can have a scoped PR in the future which does that, but I wouldn't introduce this subtle (even if innocent) change on this PR.

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -59,148 +59,198 @@ impl Sv1Server {
...
+                    Err(e) => {
                         error!(
-                            "Failed to send immediate SetDifficulty message to downstream {}: {:?}",
-                            downstream_id, e
-                        );
-                    } else {
-                        trace!(
-                            "Sent immediate SetDifficulty to downstream {} (new_target >= upstream_target)",
-                            downstream_id
+                            "Failed to build immediate SetDifficulty for downstream {downstream_id}: {e:?}; skipping"
```
GitGab19:
```suggestion
                            "Failed to build immediate mining.set_difficulty for downstream {downstream_id}: {e:?}; skipping"
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -59,148 +59,198 @@ impl Sv1Server {
...
+                    // All updates will be sent as UpdateChannel messages
+                    all_updates.push((downstream_id, channel_id, new_target, new_hashrate));
+                    // Determine if we should send set_difficulty immediately or wait
+                    match upstream_target {
+                        Some(upstream_target) => {
+                            if new_target >= upstream_target {
+                                // Case 1: new_target >= upstream_target, send set_difficulty
+                                // immediately
+                                trace!(
+                                    "✅ Target comparison: new_target ({}) >= upstream_target ({}) for downstream {}, will send set_difficulty immediately",
```
GitGab19:
```suggestion
                                    "✅ Target comparison: new_target ({}) >= upstream_target ({}) for downstream {}, will send mining.set_difficulty immediately",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -59,148 +59,198 @@ impl Sv1Server {
...
+                                immediate_updates.push((
+                                    channel_id,
+                                    Some(downstream_id),
+                                    new_target,
+                                ));
+                            } else {
+                                // Case 2: new_target < upstream_target, delay set_difficulty until
+                                // SetTarget
+                                trace!(
+                                    "⏳ Target comparison: new_target ({}) < upstream_target ({}) for downstream {}, will delay set_difficulty until SetTarget",
```
GitGab19:
```suggestion
                                    "⏳ Target comparison: new_target ({}) < upstream_target ({}) for downstream {}, will delay mining.set_difficulty until SetTarget",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -59,148 +59,198 @@ impl Sv1Server {
...
+                                    })
+                                    .map_err(TproxyError::shutdown)?;
+                            }
+                        }
+                        None => {
+                            // No upstream target set yet, send set_difficulty immediately as fallback
                             trace!(
-                                "⏳ Target comparison: new_target ({}) < upstream_target ({}) for downstream {}, will delay set_difficulty until SetTarget",
-                                new_target, upstream_target, downstream_id
+                                "No upstream target set for downstream {}, will send set_difficulty immediately",
```
GitGab19:
```suggestion
                                "No upstream target set for downstream {}, will send mining.set_difficulty immediately",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -59,148 +59,198 @@ impl Sv1Server {
...
                     }
+                };
+            if let Some(sender) = self
+                .sv1_server_io
+                .sv1_server_to_downstream_sender
+                .get_cloned(&downstream_id)
+            {
+                if let Err(e) = sender.send(set_difficulty_msg).await {
+                    warn!(
+                        "Failed to send immediate SetDifficulty message to downstream {}: {:?}; skipping (likely disconnected)",
```
GitGab19:
```suggestion
                        "Failed to send immediate mining.set_difficulty message to downstream {}: {:?}; skipping (likely disconnected)",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -59,148 +59,198 @@ impl Sv1Server {
...
+                        downstream_id, e
+                    );
+                    continue;
                 }
+                trace!(
+                    "Sent immediate SetDifficulty to downstream {} (new_target >= upstream_target)",
```
GitGab19:
```suggestion
                    "Sent immediate mining.set_difficulty to downstream {} (new_target >= upstream_target)",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -457,26 +519,26 @@ impl Sv1Server {
...
                     update.new_target,
                     SV1_MIN_DIFFICULTY_FOR_INTEGER_POWER_OF_TWO_ROUNDING,
                 ) {
-                    Ok(msg) => msg,
+                    Ok(message) => message,
                     Err(e) => {
                         error!(
-                            "Failed to build SetDifficulty for downstream {}: {:?}",
-                            update.downstream_id, e
+                            "Failed to build SetDifficulty for downstream {}: {e:?}; skipping",
```
GitGab19:
```suggestion
                            "Failed to build mining.set_difficulty for downstream {}: {e:?}; skipping",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -485,26 +547,29 @@ impl Sv1Server {
...
                 .sv1_server_io
                 .sv1_server_to_downstream_sender
-                .super_safe_lock(|downstream| downstream.get(&update.downstream_id).cloned())
+                .get_cloned(&update.downstream_id)
             {
                 if let Err(e) = sender.send(set_difficulty_msg).await {
-                    error!(
-                        "Failed to send SetDifficulty to downstream {}: {:?}",
+                    warn!(
+                        "Failed to send SetDifficulty to downstream {}: {:?}; skipping (likely disconnected)",
```
GitGab19:
```suggestion
                        "Failed to send mining.set_difficulty to downstream {}: {:?}; skipping (likely disconnected)",
```

miner-apps/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
```diff
@@ -485,26 +547,29 @@ impl Sv1Server {
...
                         update.downstream_id, e
                     );
-                } else {
-                    trace!("Sent SetDifficulty to downstream {}", update.downstream_id);
+                    continue;
                 }
+                trace!("Sent SetDifficulty to downstream {}", update.downstream_id);
```
GitGab19, line 557:
```suggestion
                trace!("Sent mining.set_difficulty to downstream {}", update.downstream_id);
```

miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
```diff
@@ -29,28 +48,29 @@ impl IsServer for Sv1Server {
...
-            );
+                    let params = server_to_client::VersionRollingParams::new(
+                        data.version_rolling_mask.clone().unwrap_or(HexU32Be(0)),
+                        data.version_rolling_min_bit.clone().unwrap_or(HexU32Be(0)),
+                    );
 
-            Ok((Some(params), Some(false)))
+                    (Some(params), Some(false))
+                })
+                .map_err(Into::into)
```
GitGab19: Why are we using this `Into::into`?

I feel it's way harder to know to what every error is mapped to with this.

miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs
```diff
@@ -118,55 +138,60 @@ impl IsServer for Sv1Server {
...
+                        share: request.clone(),
+                        extranonce: data.extranonce1.clone().into(),
+                        extranonce2_len: data.extranonce2_len,
+                        version_rolling_mask: data.version_rolling_mask.clone(),
+                        job_version: data.last_job_version_field,
+                    });
+
+                    Ok(true)
+                })
+                .map_err(TproxyErrorKind::from)?
```
GitGab19: Why are we using this `TproxyErrorKind::from`?

I feel it's way harder to know to what every error is mapped to with this.

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -268,15 +271,15 @@ impl Sv1Server {
...
-            .iter()
-            .filter_map(|downstream| {
-                downstream
-                    .value()
-                    .downstream_data
-                    .safe_lock(|data| (*downstream.key(), data.sv1_username.clone()))
-                    .ok()
-            })
-            .collect()
+        let mut worker_names = Vec::new();
```
GitGab19:
```suggestion
        let mut usernames = Vec::new();
```

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -268,15 +271,15 @@ impl Sv1Server {
...
+        self.downstreams.for_each(|downstream_id, downstream| {
+            if let Ok(worker_name) = downstream
```
GitGab19, line 276:
```suggestion
            if let Ok(username) = downstream
```

miner-apps/translator/src/lib/sv2/channel_manager/extensions_message_handler.rs
```diff
@@ -135,7 +135,10 @@ impl HandleExtensionsFromServerOwnedAsync for ChannelManager {
                 requested_extensions: Seq064KOwned::new(can_support).unwrap(),
             };
 
-            let sv2_frame: Sv2Frame = AnyMessageOwned::Extensions(new_require_extensions.into())
+            let sv2_frame: Sv2Frame =
+                AnyMessageOwned::Extensions(ExtensionsOwned::ExtensionsNegotiation(
+                    ExtensionsNegotiationOwned::RequestExtensions(new_require_extensions),
+                ))
```
GitGab19: Why this change?

miner-apps/translator/src/lib/sv2/upstream/mod.rs
```diff
@@ -381,7 +381,10 @@ impl Upstream {
                 require_extensions
             );
 
-            let sv2_frame: Sv2Frame = AnyMessageOwned::Extensions(require_extensions.into())
+            let sv2_frame: Sv2Frame =
+                AnyMessageOwned::Extensions(ExtensionsOwned::ExtensionsNegotiation(
+                    ExtensionsNegotiationOwned::RequestExtensions(require_extensions),
+                ))
```
GitGab19: Why this change?

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -49,15 +45,20 @@ use std::{
...
         bitcoin::Target,
         channels_sv2::{
             Vardiff, VardiffState,
             target::{hash_rate_from_target, hash_rate_to_target},
         },
+        mining_sv2::{
+            CloseChannelOwned as CloseChannel, SetNewPrevHashOwned as SetNewPrevHash,
+            SetTargetOwned as SetTarget,
+        },
+        parsers_sv2::MiningOwned as Mining,
```
GitGab19: Same point of the above one, let's not fragment coherent patters across our codebase.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -248,8 +247,18 @@ impl Sv1Server {
...
+                match self.handle_downstream_disconnect(downstream_id).await {
+                    Ok(()) => LoopControl::Continue,
+                    Err(cleanup_error) => {
+                        error!(
+                            downstream_id,
+                            error_kind = ?cleanup_error.kind,
+                            "failed to clean up disconnected downstream; cancelling global token"
+                        );
+                        cancellation_token.cancel();
+                        LoopControl::Break
```
GitGab19, line 254: Can you add a comment about why you're doing this?

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -505,18 +543,29 @@ impl Sv1Server {
...
+                                            .handle_downstream_disconnect(downstream_id)
+                                            .await
+                                        {
+                                            error!(
+                                                downstream_id,
+                                                error_kind = ?e.kind,
+                                                "failed to clean up disconnected downstream; cancelling global token"
+                                            );
+                                            disconnect_cancellation_token.cancel();
+                                        }
```
GitGab19, line 565: Same comment here

miner-apps/translator/src/lib/error.rs
```diff
@@ -371,6 +371,16 @@ impl From<stratum_apps::stratum_core::sv1_api::error::Error> for TproxyErrorKind
...
 }
 
+impl From<stratum_apps::stratum_core::sv1_api::error::Error> for TproxyError<Sv1Server> {
+    fn from(_: stratum_apps::stratum_core::sv1_api::error::Error) -> Self {
+        Self {
+            kind: TproxyErrorKind::SV1Error,
+            action: Action::Log,
+            _owner: PhantomData,
+        }
+    }
```
GitGab19: Does this mean that we're always gonna log and continue on every error coming from the Sv1 handlers?

Is that always the right thing to do?

bit-aloo: For the `IsServer` trait, we have three types of errors:

1. Errors that originate inside the trait, which we don't have control over. For these kinds of errors, this `From` impl is used.
2. Errors that originate from our trait implementation but are still specific to the SV1 crate. For those, we disconnect.
3. Errors that originate because of lock poisoning. For those, we shut down.

But now that you mention it, I think we should disconnect for the internal errors as well, instead of just logging them, considering we used to disconnect even before this set of changes.

Though I still need to keep this `From` implementation, as we can't use the disconnect variant as it requires downstream_id which is a runtime thing, for the SV1 handlers we just need to convert these into downstream errors.

bit-aloo: Added here: [4def06b](https://github.com/stratum-mining/sv2-apps/pull/530/commits/4def06bf2bc45f7c01edd8b6a18b04a0defbb802)

miner-apps/translator/src/lib/error.rs
```diff
@@ -100,6 +101,17 @@ where
...
+    /// Attaches a downstream to an error built without connection context, such as the sv1 errors
+    /// `From` produces for the `IsServer::Error` bound. Used on the `handle_message` call sites in
+    /// `Sv1Server`, where a protocol violation should drop that miner instead of only being logged.
+    /// Actions chosen explicitly (fallback, shutdown) are left alone.
+    pub fn with_downstream(mut self, downstream_id: DownstreamId) -> Self {
+        if matches!(self.action, Action::Log) {
+            self.action = Action::Disconnect(downstream_id);
+        }
+        self
+    }
```
GitGab19: I find the terminology a bit confusing here.

`Action::Log` is effectively being used as a placeholder for an error that is still missing its downstream context, while `with_downstream` actually changes the recovery action to `Disconnect`.

Can we call this something like `with_sv1_downstream_context` and limit it specifically to `SV1Error`? Ideally we shouldn't represent a pending disconnect as `Log` (the real and sound fix for this is to add the downstream_id in the `sv1_api` crate), but at least this would make the current workaround explicit.


```suggestion
    pub fn with_sv1_downstream_context(mut self, downstream_id: DownstreamId) -> Self {
    if matches!(&self.kind, TproxyErrorKind::SV1Error(_))
        && matches!(self.action, Action::Log)
    {
        self.action = Action::Disconnect(downstream_id);
    }

    self
}
```

bit-aloo: Yup, changing to this.

bit-aloo: Done

GitGab19: Why you didn't include the `matches!(&self.kind, TproxyErrorKind::SV1Error(_))` check?

GitGab19 (approved):

miner-apps/translator/src/lib/error.rs
GitGab19: @bit-aloo pushed my FULL suggestion above, resolving this

plebhash (approved):

Issue #529, opened 2026-05-26
need to fix `hotpath-rs` instrumentation


plebhash:
as noted in https://github.com/stratum-mining/sv2-apps/issues/492#issuecomment-4546168067 and https://github.com/stratum-mining/sv2-apps/issues/493#issuecomment-4546168514, the `hotpath-rs` instrumentation is suboptimal and arguably misleading

we are getting high latency readings that could be simply attributed to tasks sitting idle in runtime executor scheduling

Issue #257, opened 2026-02-11
Add primitive for batched disconneciton


Shourya742:
We need the ability to disconnect downstream connections in bulk (or all at once).

Issue #182, opened 2026-01-14
Add batch error flow


Shourya742:
In our apps, there are several scenarios where we perform batch processing. During these batches, errors may or may not occur. When an error does occur, the current behavior is to immediately propagate it, which can leave the execution state in an invalid or partially-updated state. Instead, we need a mechanism to collect and send a batch of errors to the main status loop, allowing it to process them centrally and take appropriate action. This will ensure we do not break execution flow mid-way or leave the system in an inconsistent state.

bit-aloo:
This pattern can be looked into: https://github.com/stratum-mining/sv2-apps/pull/531#discussion_r3427245919

PR #531, opened 2026-05-27
Migrate pool to new synchronous API

### LINKED ISSUE #205: Refactor Pool to Reduce Nested Locking
In our current Pool implementation, we rely heavily on nested locking. While this is functionally correct and provides the necessary atomic guarantees, it does not scale well under very high-throughput workloads. In such scenarios, contention increases significantly, along with the risk of lock poisoning. We recently migrated part of the tproxy implementation to use DashMap (a concurrent hash map), and the performance improvement was substantial.

Going forward, we should refactor the Pool to adopt more appropriate concurrent data structures and reduce our dependence on deeply nested locks.

bit-aloo:
closes: #205

### DIFF
--- a/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
+++ b/pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
@@ -56,16 +56,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
     ) -> Result<Vec<u16>, Self::Error> {
         let downstream_id =
             client_id.expect("client_id must be present for downstream_id extraction");
-        self.channel_manager_data.super_safe_lock(|data| {
-            let Some(downstream) = data.downstream.get(&downstream_id) else {
-                return Err(PoolError::disconnect(
-                    PoolErrorKind::DownstreamNotFound(downstream_id),
-                    downstream_id,
-                ));
-            };
+        self.with_live_downstream(downstream_id, |downstream| {
             downstream
-                .downstream_data
-                .super_safe_lock(|data| Ok(data.negotiated_extensions.clone()))
+                .negotiated_extensions
+                .get()
+                .map_err(PoolError::shutdown)
         })
     }
 
@@ -78,27 +73,21 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
         info!("Received Close Channel: {msg}");
         let downstream_id =
             client_id.expect("client_id must be present for downstream_id extraction");
-        self.channel_manager_data
-            .super_safe_lock(|channel_manager_data| {
-                let Some(downstream) = channel_manager_data.downstream.get_mut(&downstream_id)
-                else {
-                    return Err(PoolError::disconnect(
-                        PoolErrorKind::DownstreamNotFound(downstream_id),
-                        downstream_id,
-                    ));
-                };
-
-                downstream
-                    .downstream_data
-                    .super_safe_lock(|downstream_data| {
-                        downstream_data.standard_channels.remove(&msg.channel_id);
-                        downstream_data.extended_channels.remove(&msg.channel_id);
-                    });
-                channel_manager_data
-                    .vardiff
-                    .remove(&(downstream_id, msg.channel_id).into());
-                Ok(())
-            })
+        self.with_live_downstream(downstream_id, |downstream| {
+            downstream
+                .group_channel
+                .with(|group_channel| {
+                    if group_channel.has_channel_id(msg.channel_id) {
+                        group_channel.remove_channel_id(msg.channel_id);
+                    }
+                })
+                .map_err(PoolError::shutdown)?;
+            downstream.standard_channels.remove(&msg.channel_id);
+            downstream.extended_channels.remove(&msg.channel_id);
+            Ok(())
+        })?;
+        self.vardiff.remove(&(downstream_id, msg.channel_id).into());
+        Ok(())
     }
 
     async fn handle_open_standard_mining_channel(
@@ -114,62 +103,97 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
 
         info!("Received OpenStandardMiningChannel: {}", msg);
 
-        let messages = self.channel_manager_data.super_safe_lock(|channel_manager_data| {
-            let Some(downstream) = channel_manager_data.downstream.get_mut(&downstream_id) else {
-                return Err(PoolError::disconnect(PoolErrorKind::DownstreamIdNotFound, downstream_id));
-            };
-
-            if downstream.requires_custom_work.load(Ordering::SeqCst) {
-                error!("OpenStandardMiningChannel: Standard Channels are not supported for this connection");
-                let open_standard_mining_channel_error = OpenMiningChannelError {
-                    request_id,
-                    error_code: ERROR_CODE_OPEN_MINING_CHANNEL_STANDARD_CHANNELS_NOT_SUPPORTED_FOR_CUSTOM_WORK
-                        .to_string()
-                        .try_into()
-                        .expect("error code must be valid string"),
-                };
-                return Ok(vec![(downstream_id, Mining::OpenMiningChannelError(open_standard_mining_channel_error)).into()]);
-            }
-
-            let Some(last_future_template) = channel_manager_data.last_future_template.clone() else {
-                return Err(PoolError::disconnect(PoolErrorKind::FutureTemplateNotPresent, downstream_id));
-            };
-
-            let Some(last_set_new_prev_hash_tdp) = channel_manager_data.last_new_prev_hash.clone() else {
-                return Err(PoolError::disconnect(PoolErrorKind::LastNewPrevhashNotFound, downstream_id));
-            };
-
-            let payout_mode = match PayoutMode::try_from(user_identity.as_str()) {
-                Ok(mode) => mode,
-                Err(PayoutModeError::NoPayoutMode(_)) => PayoutMode::FullDonation,
-                Err(_) => {
-                    error!("Invalid user_identity '{}': does not match any supported identity format", user_identity);
+        let messages = self.with_live_downstream(downstream_id, |downstream| {
+                if downstream.requires_custom_work.load(Ordering::SeqCst) {
+                    error!("OpenStandardMiningChannel: Standard Channels are not supported for this connection");
                     let open_standard_mining_channel_error = OpenMiningChannelError {
                         request_id,
-                        error_code: ERROR_CODE_OPEN_MINING_CHANNEL_INVALID_USER_IDENTITY
+                        error_code: ERROR_CODE_OPEN_MINING_CHANNEL_STANDARD_CHANNELS_NOT_SUPPORTED_FOR_CUSTOM_WORK
                             .to_string()
                             .try_into()
                             .expect("error code must be valid string"),
                     };
-                    return Ok(vec![(downstream_id, Mining::OpenMiningChannelError(open_standard_mining_channel_error)).into()]);
+                    return Ok(vec![(
+                        downstream_id,
+                        Mining::OpenMiningChannelError(open_standard_mining_channel_error),
+                    )
+                        .into()]);
                 }
-            };
 
-            let coinbase_outputs = payout_mode.coinbase_outputs(
-                last_future_template.coinbase_tx_value_remaining,
-                &self.coinbase_reward_script,
-            );
+                let Some(last_future_template) = self
+                    .last_future_template
+                    .get()
+                    .map_err(PoolError::shutdown)?
+                else {
+                    return Err(PoolError::disconnect(
+                        PoolErrorKind::FutureTemplateNotPresent,
+                        downstream_id,
+                    ));
+                };
 
-            downstream.downstream_data.super_safe_lock(|downstream_data| {
-                downstream_data.payout_mode = Some(payout_mode);
+                let Some(last_set_new_prev_hash_tdp) =
+                    self.last_new_prev_hash.get().map_err(PoolError::shutdown)?
+                else {
+                    return Err(PoolError::disconnect(
+                        PoolErrorKind::LastNewPrevhashNotFound,
+                        downstream_id,
+                    ));
+                };
+
+                let payout_mode = match PayoutMode::try_from(user_identity.as_str()) {
+                    Ok(mode) => mode,
+                    Err(PayoutModeError::NoPayoutMode(_)) => PayoutMode::FullDonation,
+                    Err(_) => {
+                        error!(
+                            "Invalid user_identity '{}': does not match any supported identity format",
+                            user_identity
+                        );
+                        let open_standard_mining_channel_error = OpenMiningChannelError {
+                            request_id,
+                            error_code: ERROR_CODE_OPEN_MINING_CHANNEL_INVALID_USER_IDENTITY
+                                .to_string()
+                                .try_into()
+                                .expect("error code must be valid string"),
+                        };
+                        return Ok(vec![(
+                            downstream_id,
+                            Mining::OpenMiningChannelError(open_standard_mining_channel_error),
+                        )
+                            .into()]);
+                    }
+                };
+
+                let coinbase_outputs = payout_mode.coinbase_outputs(
+                    last_future_template.coinbase_tx_value_remaining,
+                    &self.coinbase_reward_script,
+                );
+
+                downstream
+                    .payout_mode
+                    .set(Some(payout_mode))
+                    .map_err(PoolError::shutdown)?;
 
                 let nominal_hash_rate = msg.nominal_hash_rate;
-                let requested_max_target = Target::from_le_bytes(msg.max_target.inner_as_ref().try_into().unwrap());
-                let extranonce_prefix = channel_manager_data.extranonce_allocator.allocate_standard().map_err(PoolError::shutdown)?;
+                let requested_max_target =
+                    Target::from_le_bytes(msg.max_target.inner_as_ref().try_into().unwrap());
+                let extranonce_prefix = self
+                    .extranonce_allocator
+                    .with(|allocator| allocator.allocate_standard())
+                    .map_err(PoolError::shutdown)?
+                    .map_err(PoolError::shutdown)?;
 
-                let channel_id = downstream_data.channel_id_factory.fetch_add(1, Ordering::SeqCst);
+                let channel_id = downstream.channel_id_factory.fetch_add(1, Ordering::SeqCst);
 
-                let mut standard_channel = match StandardChannel::new_for_pool(channel_id, user_identity.to_string(), extranonce_prefix, requested_max_target, nominal_hash_rate, self.share_batch_size, self.shares_per_minute, self.pool_tag_string.clone()) {
+                let mut standard_channel = match StandardChannel::new_for_pool(
+                    channel_id,
+                    user_identity.to_string(),
+                    extranonce_prefix,
+                    requested_max_target,
+                    nominal_hash_rate,
+                    self.share_batch_size,
+                    self.shares_per_minute,
+                    self.pool_tag_string.clone(),
+                ) {
                     Ok(channel) => channel,
                     Err(e) => match e {
                         StandardChannelError::OpenChannelInvalidNominalHashrate(code) => {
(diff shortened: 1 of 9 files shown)


bit-aloo:
Opening this up for review. Let me know what you think of API surface and if anything can be improved or is not clear

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -116,7 +90,24 @@ impl ChannelManagerIo {
...
+    downstream_id_factory: Arc<AtomicUsize>,
+    // Mapping of `(downstream_id, channel_id)` -> vardiff controller.
+    // Each entry manages variable difficulty for a specific downstream channel.
+    pub(crate) vardiff: SharedMap<VardiffKey, VardiffState>,
+    // Coinbase outputs.
+    pub(crate) coinbase_outputs: Vec<u8>,
+    // Last new prevhash.
+    pub(crate) last_new_prev_hash: SharedLock<Option<SetNewPrevHash<'static>>>,
+    // Last future template.
+    pub(crate) last_future_template: SharedLock<Option<NewTemplate<'static>>>,
```
GitGab19, line 105: Since now these two are independent, we should have a single wrapper behind a single `SharedLock`, containing both of them.

In this way we'll be sure that `last_future_template` and `last_new_prev_hash` are always in sync between each-other.

GitGab19:
My clanker found a real concern about the usage of `self.downstream.get_cloned()` in the channel openings handlers:
```
Possible interleaving:

1. Open-channel handler clones the `Downstream`.
2. The downstream disconnects.
3. `remove_downstream()` removes the downstream and retains away existing `vardiff` entries.
4. The open-channel handler continues with the cloned handle.
5. It inserts a channel into `downstream.standard_channels` / `extended_channels`.
6. It inserts a new `(downstream_id, channel_id)` entry into `self.vardiff`.

At that point, `self.vardiff` can contain an entry for a downstream that no longer exists in `self.downstream`. `run_vardiff()` will then skip it forever because `self.downstream.get_cloned(&downstream_id)` returns `None`.

This is not a memory-safety issue, but it does break the manager invariant that `vardiff` entries correspond to live downstream/channel state. I think the final channel insertion and `vardiff` insertion should happen while holding/rechecking the downstream map entry, or otherwise verify membership at commit time.
```

pool-apps/pool/src/lib/channel_manager/mod.rs
bit-aloo: Even if we not keep them inside a single wrapper abstraction, they still would be in sync. Like I can individually update them. I am not understanding, how wrapping them together helps? Can you elaborate more?

GitGab19: The point here is not so much about updating their values in sync, but more about reading their values.

If one of the two values gets updated while something else is reading the other, you could end up in a situation where the `last_future_template` and `last_new_prev_hash` you read are not in sync between each-other.

Is it more clear now?

bit-aloo: Hmm, I do get what you are trying to say. Let’s say the task order is this:

```rust
(Prevhash_old, future_template_old)

1. handle_new_template (Prevhash_old, future_template_new)
   // Currently in an inconsistent state, as the future template does not correspond to the prevhash.

2. downstream connects
   // We initialize it with an inconsistent pair of future template and prevhash.
   // But this is okay, because by the time we send OpenChannelSuccess,
   // it will receive a consistent snapshot of prevhash and future_template.

3. handle_new_prevhash (Prevhash_new, future_template_new)
```

We use a common `select!` block in `ChannelManager`, which processes ChannelManager tasks sequentially.

The only times these values are used are:

1. When a downstream connects (having an inconsistent state here is okay, since it will eventually get corrected).
2. When we run a ChannelManager task, which happens sequentially because of the `select!` block, ensuring we do not read things out of order.

But I do get your point, ideally, we should wait for the pair to become consistent before even performing a read operation.

Currently, and even before this refactor, that was not the case, because both of these fields (`last_new_prevhash` and `last_future_template`) are updated through completely different messages, which naturally introduces a time window where they can be inconsistent.

bit-aloo:
> This is not a memory-safety issue, but it does break the manager invariant that `vardiff` entries correspond to live downstream/channel state. I think the final channel insertion and `vardiff` insertion should happen while holding/rechecking the downstream map entry, or otherwise verify membership at commit time.
> ```

Added here: b3dcf5d, thanks.

pool-apps/pool/src/lib/channel_manager/mod.rs
GitGab19: As discussed during yesterday's PR Review Club call, we are gonna improve this later, as tracked by https://github.com/stratum-mining/sv2-apps/issues/551

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -116,7 +90,24 @@ impl ChannelManagerIo {
 /// to perform message traversal.
 #[derive(Clone)]
 pub struct ChannelManager {
-    pub(crate) channel_manager_data: Arc<Mutex<ChannelManagerData>>,
+    // Mapping of `downstream_id` -> `Downstream` object,
+    // used by the channel manager to locate and interact with downstream clients.
+    pub(crate) downstream: SharedMap<DownstreamId, Downstream>,
```
plebhash: on multiple occasions while reviewing this PR I read `self.downstream` and think it's a single object, only to find out later that's actually a collection

```suggestion
    pub(crate) downstreams: SharedMap<DownstreamId, Downstream>,
```

plebhash:
I'm a bit concerned with the usage of `get_cloned` on this PR.

I assume it's being introduced as a strategy to optimize against contention from long-held entry guards in `with` and `with_mut`

more specifically, the point raised by @GitGab19 above has not been fully addressed

there are other potential TOCTOU race windows related to similar usage of `get_cloned` in several places in this PR:
- `handle_open_standard_mining_channel` (as already stated)
- `handle_open_extended_mining_channel` (as already stated)
- `handle_update_channel`
- `handle_set_custom_mining_job`
- `handle_new_template`
- `handle_set_new_prev_hash`
- `run_vardiff`

https://github.com/stratum-mining/sv2-apps/commit/b3dcf5da277ea92f55b995f6c2f5c9676a37ebc0 partially mitigates the issue, but open-channel handlers still allow stale cloned downstream handles to commit channel/vardiff state after disconnect... while I think this could be ok to live with as a deliberate tradeoff for this specific scenario, we're still violating invariants and I also don't think we can trivially generalize this pattern to solve all potential TOCTOU race windows listed above

---

overall, we're already swapping the monolithic/nested-lock model for fine-grained `SharedLock` and `SharedMap`, and we're yet to analyze the performance gains coming from this improvement

let's not forget that the goal of #368 was to introduce safe `DashMap`-based primitives that will help us deprecate the unsafe custom `Mutex`... the fact that now we're bypassing those primitives via `get_cloned` feels like we're moving the goal post in a precipitated way

moreover, #529 still needs to be fixed, so even if de-nesting locks + `DashMap` sharding only brings mild/moderate performance improvements, we still lack empirical proof of broader contention issues

so IMHO we should do things one step at a time:

for now, avoid aggressively using `get_cloned` as an over-optimized strategy to reduce contention on this PR, since that is introducing concurrency side-effects for which mitigation is non-trivial and will likely blow up scope

for now, `get_cloned` usage should be restricted to snapshot-style usage only (and the function should be documented accordingly)

if we later see that contention remains an issue, then we should leverage the foundations established by #368 + this PR to address it with a proper strategy

bit-aloo:
The argument that the above suggestion is only partially valid is not entirely sound.

If you look closely, the concern raised above is that we are checking liveness only at the end rather than at every commit point. This is intentional, as I wanted to keep the logic simple and easier to reason about. These intermediate commit points do not materially affect the outcome. If a downstream disconnect occurs during channel opening, performing the cleanup at the end is sufficient. Otherwise, we would need to introduce cleanup logic after every commit step, which would significantly increase complexity and reduce maintainability.

Regarding the TOCTOU concern, where it is mentioned that a race could occur in-between the channel manager methods and the vardiff task, I agree that the concern is partially valid(only for vardiff and select task). However, all channel manager methods execute sequentially, so a race condition cannot occur between those methods themselves. The only concurrency involved is with the vardiff task. A side note the partial commits are ok if the object eventually gonna be dropped

It will be useful, if you can mention the TOCTOU scenarios?

Thanks for review though, would be awesome discuss more on this.

plebhash:
> so a race condition cannot occur between those methods themselves. 

that is not the claim here, it's about `handle_*` execution and `remove_downstream`

> The argument that the above suggestion is only partially valid is not entirely sound.

`is_alive` introduced via b3dcf5d is only partial mitigation because it could return `true` and allow execution to continue, only for `remove_downstream` to be executed immediately afterwards and break invariants again

end of the day, the truth is that with this extensive usage of `get_cloned`, the removal of a `Downstream` object is not atomic, and `is_alive` only gives us loose guarantees but no sound atomicity

if we decided to live with these loose guarantees, we should consciously accept the following tradeoffs:
- potential wasted CPU cycles
- potential misleading logs with concurrent contradictory signals about the application state
- temporary ghost vardiff entries from TOCTOU races on `handle_open_*`

(maybe more, these are the ones I can identify right now)

but the question is: do we want/need to make that decision now?

---

overall I'm not going to die on this hill, and I don't want to spend more time debating all the details of this rabbit hole

I'd rather focus on my broader point, which I'm happy to clarify in case it's not already sufficiently clear:

the extensive usage of `get_cloned` as a strategy for contention optimization feels precipitated (given scopes of #368 + #529), and it seems we don't have a good grasp of the second-order consequences of it, so I don't see the point in rushing things at the expense of a new category of unknown problems to be solved in the future

IMHO we should stick to the originally planned scope and measure the performance gains before we jump into new strategies for contention optimization... if we later find that we're still facing performance issues, we can evaluate `get_cloned` as a potential strategy and its inherent tradeoffs more carefully

bit-aloo:
Hmm, the idea behind introducing the clone was not because of any technical limitation or API issue . It was mainly about making the code easier to write and follow. If you look at the code now, it is just a simple flow: get the object, perform the operation, and move on, let the system follow eventual consistency. 

Even if the object becomes invalid, the operation can still happen, and the system will eventually recover and become consistent again. That was the idea behind this. The introduction of `is_alive` was also along the same lines: making the system faster by relying on eventual consistency instead of always enforcing strong consistency. The side effects of operating on an invalid object were not expected to be serious. At least, that was my thinking.

That said, nothing is stopping us from providing atomic access to objects and ensuring they cannot become invalid while they are being used. That would give us stronger consistency guarantees (like before).

I agree that we should get some numbers before deciding. I will add a commit that prevents operations on invalid objects. Cool.

bit-aloo:
Removed get_cloned for downstream, here: [ce528a0](https://github.com/stratum-mining/sv2-apps/pull/531/commits/ce528a07eeef60b9039d408812321ca263c708f1).


I will squash commit later, would make it easy to review.

plebhash:
> Even if the object becomes invalid, the operation can still happen, and the system will eventually recover and become consistent again. That was the idea behind this. The introduction of is_alive was also along the same lines: making the system faster by relying on eventual consistency instead of always enforcing strong consistency. The side effects of operating on an invalid object were not expected to be serious. At least, that was my thinking.

I think this is a reasonable rationale, and I wouldn't completely rule it out of the design space yet.

but tbh conflicting logs and ghost vardiff entries that could last up to 60s in memory worry me about the UX implications

now that `sv2-ui` is a first-class citizen on the SRI stack, we must always put ourselves in the shoes of the end user before we make key architectural/design decisions

> Hmm, the idea behind introducing the clone was not because of any technical limitation or API issue . It was mainly about making the code easier to write and follow. If you look at the code now, it is just a simple flow: get the object, perform the operation, and move on, let the system follow eventual consistency.

I empathize with the urge to make devX better. It's been painful to live with closure complexity of the soon-to-be-deprecated custom `Mutex`, and I also wish it went away.

But we made it this far, and migrating away from it is a delicate operation. So I feel it's wiser to move slow, evaluate consequences abroad the entire SRI stack, and optimize things gradually.

We laid some solid foundation on #368. Let's take some time and explore it. How will granularly locked sharded maps perform in face of curtailed loads (with granular instrumentation of the source code)? I'm genuinely curious to look at those numbers!

> Removed get_cloned for downstream, here: [ce528a0](https://github.com/stratum-mining/sv2-apps/pull/531/commits/ce528a07eeef60b9039d408812321ca263c708f1).
> 
> I will squash commit later, would make it easy to review.

awesome, thanks!

plebhash:
can we clean up commit history?

e0fe2cb7eade507026278366f1014efe4c95458e and ce528a07eeef60b9039d408812321ca263c708f1 are a bit redundant now (add `get_cloned` then remove `get_cloned`)

bit-aloo:
> I will squash commit later, would make it easy to review.

 I assume you had looked into the final commit earlier.

> can we clean up commit history?
> 
> [e0fe2cb](https://github.com/stratum-mining/sv2-apps/commit/e0fe2cb7eade507026278366f1014efe4c95458e) and [ce528a0](https://github.com/stratum-mining/sv2-apps/commit/ce528a07eeef60b9039d408812321ca263c708f1) are a bit redundant now (add `get_cloned` then remove `get_cloned`)

Squashed. Awesome, thanks.

plebhash (approved):

pool-apps/pool/src/lib/monitoring.rs
```diff
@@ -109,12 +107,8 @@ impl Sv2ClientsMonitoring for ChannelManager {
...
-        self.channel_manager_data
-            .safe_lock(|d| {
-                d.downstream
-                    .get(&client_id)
-                    .and_then(downstream_to_sv2_client_info)
-            })
-            .unwrap_or(None)
+        self.downstreams
+            .get_cloned(&client_id)
+            .and_then(|downstream| downstream_to_sv2_client_info(&downstream))
```
GitGab19: Why are you still using `get_cloned()` here?

According to [your comment](https://github.com/stratum-mining/sv2-apps/pull/531#issuecomment-4691121816), you should have removed it?

bit-aloo: Slipped, thanks updated.

stratum-apps/src/sync.rs
```diff
@@ -122,6 +123,16 @@ impl<K: Eq + Hash + Clone, V> SharedMap<K, V> {
         SharedMap(Arc::new(DashMap::new()))
     }
 
+    /// Get an owned clone of a value.
+    ///
+    /// Prefer [`Self::with`] when only part of a large value is needed.
```
plebhash: let's make docs explicitly clear about the appropriate vs inappropriate usage of this method

stratum-apps/src/sync.rs
bit-aloo, line 237: I have a question about the `for_each` APIs. In cases like this, if an error occurs, we simply return. Since we're iterating over a collection, multiple items may fail, and we generally don't want a single error to cause an early exit and prevent the remaining items from being processed.

At the same time, I would prefer not to make the API surface significantly more complex. However, I think we do need some form of batching or error aggregation here to ensure that iteration isn't aborted prematurely. 

All ears for new design.

GitGab19: My clanker gave me this example from `tokio`, which could be taken into consideration as a possible approach: https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_all

In the `join_all` docs, they explicitly say: "If any tasks on the JoinSet fail with an [JoinError](https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html), then this call to join_all will panic and all remaining tasks on the JoinSet are cancelled. To handle errors in any other way, manually call [join_next](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_next) in a loop.`

The `join_next` docs can be found here: https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_next

---

In our case here, the method corresponding to the tokio's `join_next()` is the `for_each()` that we already have, right?

So I guess we can use that when we know there's a fallible operation, and collect the eventual errors at the caller site?

Maybe I'm missing something?

jokeez:
Hi — reviewed this in context of #551/#552 (TDP cache/snapshot ordering).

I can contribute a small state-transition test matrix for:
- NewTemplate -> SetNewPrevHash ordering
- missing/late SetNewPrevHash after NewTemplate
- stale snapshot reuse during channel init

No commercial pitch — just reproducible edge cases + expected invariants.
If useful, I’ll post it here as a markdown table first, then optional PR with tests.

stratum-apps/src/sync.rs
plebhash: @Shourya742 is there some part of this PR that you feel is blocked by this?

I'm asking because we already have #182 (batch error flow) and #257 (batched disconnect)

I'm not against brainstorming here, this is a good opportunity because in a way we're still trimming the rough edges of #368

but since we already have trackers for this specific issue, it might make sense to tackle that in a separate PR?

unless there's a concrete blocker for this PR ofc

---

btw, from a relatively superficial analysis I think `next` seems to make sense... instead of looping inside the function body, do it on the call site and handle errors individually (e.g.: disconnect) before moving on to the next iteration

maybe we can take note of `next` pattern as a potential solution on #182 and #257 and proceed with this PR without addressing this specific issue?

bit-aloo: This sounds good as well, I will update this PR, and we can proceed with the merge.

stratum-apps/src/sync.rs
GitGab19: Maybe we can add something specifically about the risks of using it here @Shourya742 ?

bit-aloo: I did add a note here: [24afb06](https://github.com/stratum-mining/sv2-apps/pull/531/commits/24afb06fae3044904628de2ea629e0dd5441e533)

GitGab19: This is what my clanker would suggest as docs:
```
/// Get an owned clone of a value.
///
/// This releases the map entry guard immediately, so the entry may be removed
/// or replaced while the caller is still using the clone. Use this only when
/// stale clones are acceptable. Prefer [`Self::with`] or [`Self::with_mut`]
/// when later work depends on the entry still being present.
```

bit-aloo: done

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -699,6 +705,32 @@ impl ChannelManager {
 
         Ok(())
     }
+
+    /// Runs `f` for a downstream that must still be connected.
+    ///
+    /// Returns `DownstreamNotFound` if the downstream disappeared before the lookup completed.
```
GitGab19:
```suggestion
    /// Runs `f` while holding the downstream map entry guard.
    ///
    /// Use this when mutations must only happen if the downstream is still
    /// registered in the ChannelManager. Do not await, send messages, or re-enter
    /// `self.downstreams` inside `f`.
    ///
    /// Returns the closure result if the downstream is registered. Returns
    /// `DownstreamNotFound` with a disconnect action if the downstream is no
    /// longer registered.
```

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -699,6 +705,32 @@ impl ChannelManager {
...
+    #[allow(clippy::result_large_err)]
+    fn with_live_downstream<R, F>(
```
GitGab19:
```suggestion
    fn with_registered_downstream<R, F>(
```

pool-apps/pool/src/lib/channel_manager/mod.rs
bit-aloo: compiler wont allow using .awaits inside non async closure. I think we are kinda ok even if we not mention the await point

GitGab19: Alternative without the `async` word :laughing: :
```
/// Runs `f` while holding the downstream map entry guard.
///
/// Use this when mutations must only happen if the downstream is still
/// registered in the ChannelManager. Keep `f` short: do not perform blocking
/// work, send/forward messages, or re-enter `self.downstreams` inside it.
///
/// Returns the closure result if the downstream is registered. Returns
/// `DownstreamNotFound` with a disconnect action if the downstream is no
/// longer registered.
```

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -78,27 +73,21 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
-            })
+        self.with_live_downstream(downstream_id, |downstream| {
+            downstream
+                .group_channel
+                .with(|group_channel| {
+                    if group_channel.has_channel_id(msg.channel_id) {
+                        group_channel.remove_channel_id(msg.channel_id);
+                    }
+                })
+                .map_err(PoolError::shutdown)?;
```
GitGab19, line 84: Why did you introduce this part about group channel?

bit-aloo: because we were not cleaning it before when a close channel message is received from dowstream. I had a separate commit for this before to be explicit about this change but got squashed during rebasing.

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -579,143 +639,209 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
-                            info!(
-                                "SubmitSharesStandard: valid share | downstream_id: {}, channel_id: {}, sequence_number: {}, share_hash: {}, share_work: {} ✅",
-                                downstream_id, channel_id, msg.sequence_number, share_hash, share_work
-                            );
+                                Ok(messages)
+                            });
+                    match validation {
+                        Some(validation) => {
+                            self.vardiff.with_mut(&vardiff_key, |vardiff| {
+                                vardiff.increment_shares_since_last_update()
```
GitGab19, line 820: Why did you move the `increment_shares_since_last_update()` inside the `Some()`?

bit-aloo: because validation would not happen only in case a channel doesn't exist, so we only increment the vardiff if the channel is actually present. I would say this is mostly because of the datastructure API that we have to have such matches, otherwise this would always be some considering we are already checking for channel existence before.

GitGab19: Yeah, I just checked out the PR branch locally, and I verified that everything is still the same.

Sorry but looking at the diffs over github is complicated in this part of the codebase.

pool-apps/pool/src/lib/monitoring.rs
```diff
@@ -12,95 +12,93 @@ use crate::{channel_manager::ChannelManager, downstream::Downstream};
 /// Helper to convert a Downstream to Sv2ClientInfo.
 /// Returns None if the lock cannot be acquired (graceful degradation for monitoring).
```
GitGab19:
```suggestion
```

GitGab19 (approved):

plebhash:
I feel bad for pointing this out over and over, but I have to ask:

why do we have 0f2dad92b52622ba85a2d7b7e990f835ef4aa570 and 262b54552df86818a5902262553bf251ab3f2d0c as standalone commits?

PR #536, opened 2026-06-01
feat(test-tools): add shape-proxy SV2 share-gating proxy

gimballock:
**Summary**

Adds a new test tool that sits between mining devices and a pool, enabling controlled modification of the share
submission rate without touching mining hardware. Modulating share streams against a live pool is otherwise difficult
— miners produce shares at a rate determined by physics, not configuration.

This tool complements [the vardiff simulation framework](https://github.com/stratum-mining/stratum/pull/2154) by providing calibration testing against real pool
infrastructure with real miners.

- Token-bucket gating with supply-relative rate profiles (Track, Step, Ramp, Stall, Burst, Oscillate)
- Difficulty-weighted supply measurement via count-based rolling window (N=15), stable across vardiff changes
- HTTP API for dynamic profile control and live status monitoring
- Connection resilience — upstream reconnects with exponential backoff; miners stay connected during pool disconnects

**Architecture**

```
[Miner(s)] <--SV1--> [Translator] <--SV2--> [Shape-Proxy] <--SV2--> [Pool]
                                                 |
                                              HTTP API
```

Each arriving share is immediately ACKed to the miner (decoupling it from pool behavior), then the gate decides
whether to forward or drop based on the active profile and measured supply rate.

Using with SV1 miners

Most miners only speak Stratum V1. To use them with shape-proxy, place the existing translator_sv2 (from
roles/translator) between the miner and the proxy:
```
# translator config (translator.toml)
upstream_address = "127.0.0.1:34262"  # shape-proxy's downstream_listen
upstream_authority_pubkey = "<shape-proxy's public key>"
downstream_address = "0.0.0.0:3333"   # what the miner connects to

# shape-proxy config (shape-proxy.toml)
upstream_address = "pool-host:3335"
downstream_listen = "0.0.0.0:34262"
```
The translator is a pass-through protocol adapter — it doesn't buffer, gate, or modify shares. It adds no meaningful
latency or behavioral change to the share stream. The only operational impact is an additional process to manage.

**Test plan**

- [x] cargo test passes for all unit tests (profile factor_at, gate behavior, target_to_difficulty)
- [x] Connect an SV1 miner → translator → shape-proxy → pool — verify shares flow at Track{1.0}
- [x] Set Step profile via API — verify pool sees the rate change at the configured time
- [ ] Set Stall profile — verify shares stop forwarding during the stall window then resume
- [ ] Verify difficulty floor prevents pool SetTarget below threshold
- [ ] Verify upstream reconnection — kill pool, wait, restart — miner stays connected, shares resume


GitGab19:
While this might be something useful for your investigation related to vardiff, it's definitely not something which we need to add and maintain on this repo.

gimballock:
Yeah, this probably isn't the right place for this to live.
I think we will keep the vardiff stuff on our org in GitHub.

The vardiff experiment repo itself has a similar identity crisis. I think I will make an isolated pr just for the final vardiff change, to separate the test framework from the results.

PR #544, opened 2026-06-05
skip aggregated UpdateChannel forwarding when the channel with the upstream is not opened

### LINKED ISSUE #543: tProxy sends `UpdateChannel` upstream even if the `AGGREGATED_CHANNEL_ID` channel hasn't been opened
On https://github.com/stratum-mining/sv2-apps/issues/386, @coleFD reported some issues he's having while connecting a S21 to our tProxy.

For a still unknown reason the miner disconnects before tProxy is even able to open a channel with its upstream, and after that tProxy still sends a `UpdateChannel` message (`UpdateChannel { channel_id: 0, nominal_hash_rate: 0.0, ... }`.

The reason is that we are not currently checking for the presence of the `AGGREGATED_CHANNEL_ID` channel before sending this message to the upstream.

GitGab19:
Closes #543

### DIFF
--- a/integration-tests/tests/translator_integration.rs
+++ b/integration-tests/tests/translator_integration.rs
@@ -625,6 +625,69 @@ async fn test_translator_keepalive_job_sent_and_share_received_by_pool() {
     shutdown_all!(translator, pool);
 }
 
+// Verifies that aggregated tProxy does not send UpdateChannel upstream if an SV1
+// downstream disconnects before the aggregated upstream channel has opened (see #543).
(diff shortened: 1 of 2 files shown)


bit-aloo (commented):
Nice catch. Can we add an IT?

GitGab19:
@Shourya742 pushed an additional commit with the IT you asked.

plebhash (approved):

bit-aloo (approved):

PR #545, opened 2026-06-05
notify `sv2-ui` on JDC/tProxy config changes

### LINKED ISSUE #540: add workflow for changes on the config that affects the sv2-ui
similar to what we have for the `openapi.json` that triggers an issue on `sv2-ui` repo. see https://github.com/stratum-mining/sv2-apps/blob/main/.github/workflows/notify-monitoring-api-update.yaml

also similar to the docker drift workflow that we have here: https://github.com/stratum-mining/sv2-apps/blob/main/.github/workflows/check-docker-config.yaml

I think we should also trigger a issue on `sv2-ui` if something change in the config files of JDC/TProxy.

this necessity was triggered by the merge of https://github.com/stratum-mining/sv2-apps/pull/501 which broke the `sv2-ui` config file generator and took me sometime to realize that this was the problem. Because my initial thought was that something was wrong with the shared docker volumes.

GitGab19:
This PR introduces a new workflow which opens an issue on `sv2-ui` whenever something on the JDC/tProxy config schema changes.

Closes #540

### DIFF
--- a/.github/workflows/notify-jdc-tproxy-config-update.yaml
+++ b/.github/workflows/notify-jdc-tproxy-config-update.yaml
@@ -1,0 +1,194 @@
+name: Notify JDC/tProxy Config Update
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'miner-apps/jd-client/config-examples/**/*.toml'
+      - 'miner-apps/translator/config-examples/**/*.toml'
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number (for manual runs)'
+        required: true
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Detect config schema changes
+        id: config-schema-changes
+        shell: bash
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            echo "Manual dispatch requested; skipping config schema-change gate."
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          CHANGED_FILES="$(scripts/list-config-schema-changes.sh "${{ github.event.before }}" "${{ github.sha }}")"
+
+          if [[ -z "$CHANGED_FILES" ]]; then
+            echo "No JDC/tProxy config schema changes detected."
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          {
+            echo "changed=true"
+            echo "files<<EOF"
+            echo "$CHANGED_FILES"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Create Issue
+        if: steps.config-schema-changes.outputs.changed == 'true'
+        uses: actions/github-script@v9
+        env:
+          CHANGED_CONFIG_FILES: ${{ steps.config-schema-changes.outputs.files }}
+        with:
+          github-token: ${{ secrets.SV2_UI_TOKEN }}
+          script: |
+            const { owner, repo } = context.repo;
+            const targetOwner = 'stratum-mining';
+            const targetRepo = 'sv2-ui';
+            const labelName = 'jdc-tproxy-config-change';
+            const issueTitlePrefix = 'JDC/tProxy config updated in PR';
+            const watchedFilePatterns = [
+              /^miner-apps\/jd-client\/config-examples\/.*\.toml$/,
+              /^miner-apps\/translator\/config-examples\/.*\.toml$/
+            ];
+
+            let prNumber;
+
+            // 1. Manual trigger support
+            if (context.eventName === 'workflow_dispatch') {
+              prNumber = context.payload.inputs?.pr_number;
+            }
+
+            // 2. Try to resolve from context
+            if (!prNumber) {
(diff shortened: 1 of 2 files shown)


GitGab19:
I ran some tests on my fork, for which you can see the results here:
- workflow execution --> https://github.com/GitGab19/sv2-apps/actions/runs/27025217227/job/79763409893
- issue opened on my fork of `sv2-ui` --> https://github.com/GitGab19/sv2-ui/issues/3

.github/workflows/notify-jdc-tproxy-config-update.yaml
```diff
@@ -0,0 +1,193 @@
+name: Notify JDC/tProxy Config Update
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'miner-apps/jd-client/config-examples/**/*.toml'
+      - 'miner-apps/translator/config-examples/**/*.toml'
+      - 'docker/config/jdc-config.toml.template'
+      - 'docker/config/translator-proxy-config.toml.template'
```
lucasbalieiro: I think watching for the docker files might introduce false positives because we could make changes there that not necessarily  reflect a change on teh config file itself.

I would look just for the original toml files

GitGab19: Removed those two paths and force-pushed again the PR.

lucasbalieiro (approved):
utACK

bit-aloo (commented):
If I understand correctly, the current script deduplicates repeated sections and effectively squashes them into a single set of key paths to detect whether keys were added or removed. That works for key-level changes, but I think we should also track the shape of the toml structure.
For example, changing `[[upstreams]]` to `[upstreams]` would be a breaking schema change even if the nested key names stay the same. Similarly, changing fields from scalar values to arrays, such as moving from repeated upstream tables to one upstream section with array-valued fields, would not be detected by key-path comparison.
Since those shape/type changes can affect the sv2-ui config generator, it may be better to compare a lightweight schema signature instead of only comparing keys.

GitGab19:
Good point @Shourya742, I updated the script to make it check also types now.

.github/workflows/notify-jdc-tproxy-config-update.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+            // 2. Try to resolve from context
+            if (!prNumber) {
+              let prs = [];
+              try {
+                const response = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+                  owner,
+                  repo,
+                  commit_sha: context.sha
+                });
```
bit-aloo, line 88: Can we move this to a script file.

scripts/list-config-schema-changes.sh
```diff
@@ -0,0 +1,127 @@
...
+    *) return 1 ;;
+  esac
+}
+
+extract_schema_from_file() {
+  local file="$1"
+
+  # Use Python's standard TOML parser so the signature follows TOML semantics
+  # instead of approximating table shape with shell text parsing.
+  python3 - "$file" <<'PY' | sort -u
```
bit-aloo, line 37: Shouldn't this require a pre-requisite on python, do we have python on our ci?

bit-aloo (commented):
Few suggestions:

Clanker pointed this out:

```rust
The most significant correctness issue is probably #3 (array-of-tables detection for empty arrays), followed by #4 (empty arrays causing false schema diffs) if these config examples frequently contain placeholder empty lists
```

.github/workflows/notify-jdc-tproxy-config-update.yaml
GitGab19: What would be the benefit of doing that?

That script wouldn't be used anywhere else than on this CI job.

scripts/list-config-schema-changes.sh
GitGab19: Github runners have python installed, but I added a step to install it to be more safe.

GitGab19:
> Few suggestions:
> 
> Clanker pointed this out:
> 
> ```rust
> The most significant correctness issue is probably #3 (array-of-tables detection for empty arrays), followed by #4 (empty arrays causing false schema diffs) if these config examples frequently contain placeholder empty lists
> ```

If I understood this correctly, I think we don't have any issue here.

In case we fire the notification for a change on a placeholder empty field, we can always close the issue on `sv2-ui` for that false-positive.

It's way better than introducing unnecessary complexity and risk to miss real config changes.

.github/workflows/notify-jdc-tproxy-config-update.yaml
bit-aloo: It would be more readable, but ya ah, we won't be using it anywhere else. Not a blocker for me.

bit-aloo:
> > Few suggestions:
> > Clanker pointed this out:
> > ```rust
> > The most significant correctness issue is probably #3 (array-of-tables detection for empty arrays), followed by #4 (empty arrays causing false schema diffs) if these config examples frequently contain placeholder empty lists
> > ```
> 
> If I understood this correctly, I think we don't have any issue here.
> 
> In case we fire the notification for a change on a placeholder empty field, we can always close the issue on `sv2-ui` for that false-positive.
> 
> It's way better than introducing unnecessary complexity and risk to miss real config changes.

It took me a while to understand this. Do you think it would make sense to just compare the config byte by byte, and if anything changes, let the action open an issue? We can always close it if it's not relevant. That would make what we're doing here much simpler. Not a blocker :P.

bit-aloo (approved):

GitGab19:
> Do you think it would make sense to just compare the config byte by byte, and if anything changes, let the action open an issue?

I think this would be way less precise than the current script, triggering false positives every time we're gonna change just the value of a field.

PR #548, opened 2026-06-08
`bitcoin_core_sv2` supports multiple versions of Bitcoin Core (v30.x + v31.x)

### LINKED ISSUE #516: `bitcoin_core_sv2` must simultaneously support Bitcoin Core `v30.x` + `v31.x`
## intro

currently, each version of `bitcoin_core_sv2` crate supports one specific version of Bitcoin Core... that is tracked by this table:
https://github.com/stratum-mining/sv2-apps/blob/fe4e51d7947872c3315e67b88b8388010b8324a8/bitcoin-core-sv2/README.md?plain=1#L39-L46

(which btw is outdated on `main`, PR https://github.com/stratum-mining/sv2-apps/pull/514 is updating it)

as a consequence, Pool/JDS and JDC are always limited to specific versions of Bitcoin Core, with no flexibility

this is starting to cause PITA on `sv2-ui`, because ideally we need to support both `v30.x` and `v31.x` there, but HTTP API drifts across different `sv2-apps` release cycles result in broken compatibility with the `sv2-ui` codebase

we experimented with some strategies around backporting PRs that change the HTTP APIs into past release cycles (see #504), but the [conclusion](https://github.com/stratum-mining/sv2-apps/issues/504#issuecomment-4491142098) was that this is not a sustainable strategy

so currently, our only viable approach is to bloat the `sv2-ui` codebase with conditionals that treat the HTTP endpoints differently depending on which release cycle of `sv2-apps` is being deployed via docker containers

while that works as a temporary strategy, it is very dirty and suboptimal at best

## solution

as a more sustainable strategy, we must tackle backwards compatibility with Bitcoin Core at the `bitcoin_core_sv2` architecture level

the crate must be able to support IPC-enabled releases of Bitcoin Core (starting at `v30.x`) by providing the following modular structure:
 - `bitcoin_core_sv2::common::{job_declaration_protocol,template_distribution_protocol}` modules define version-agnostic traits (locked down to Sv2 TDP/JDP IO primitives)
 - `bitcoin_core_sv2::unix_capnp::v30x::{job_declaration_protocol,template_distribution_protocol}` modules have an IPC-flavored implementation of the traits for `v30.x`
 - `bitcoin_core_sv2::unix_capnp::v31x::{job_declaration_protocol,template_distribution_protocol}` modules have an IPC-flavored implementation of traits for `v31.x`
 - etc

then, Pool/JDS and JDC get a new `config.toml` parameter:
```
[template_provider_type.BitcoinCoreIpc]
...
# possible values currently limited to 30 or 31, anything else results in immediate shutdown
version = 31
```

this config parameter dictates which submodule of `bitcoin_core_sv2` the app loads at runtime

## version support

officially, there's active maintenance of Bitcoin Core for the past 3 releases (currently `v28.x` has already reached EOL)

for now, the only IPC-enabled releases of Bitcoin Core are `v30.x` and `v31.x`, so we don't have to worry about EOL for the near future

but whenever `v33.x` cycle starts, we will have to decide whether to drop `v30.x` support or not

## test coverage

the main Integration Test coverage of `sv2-apps` must be always locked to whatever is the latest Bitcoin Core release cycle (currently `v31.x`)

but we must have base TDP + JDP Integration Test coverage for both `v30.x` and `v31.x`

every time `sv2-apps` starts supporting a new release cycle of Bitcoin Core:
- we add the same base TDP + JDP Integration Test coverage for this new version
- the main Integration Test coverage is updated to use this new version

plebhash:
close #516

---

in summary:
- `bitcoin_core_sv2` now supports **both Bitcoin Core IPC v30.x and v31.x** via one common API with versioned backends.
- `template_provider_type.BitcoinCoreIpc.version` is now **required** (`30` or `31`) and wired through Pool/JDC/JDS runtime selection.
- Added focused integration coverage for both versions in new IPC IO tests (JDP + TDP).
- Docker config is now version-configurable via:
  - `POOL_BITCOIN_CORE_IPC_VERSION`
  - `JDC_BITCOIN_CORE_IPC_VERSION`

---

note: should be merged after https://github.com/stratum-mining/sv2-apps/pull/545


bitcoin-core-sv2/src/common/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Supported Bitcoin Core IPC schema families.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum BitcoinCoreVersion {
+    V30X,
+    V31X,
+}
+
+impl BitcoinCoreVersion {
+    pub const fn as_major(self) -> u16 {
```
bit-aloo: u8 would work?
```rust
pub const fn as_major(self) -> u8
```

bitcoin-core-sv2/src/common/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+        match self {
+            Self::V30X => 30,
+            Self::V31X => 31,
+        }
+    }
+}
+
+impl TryFrom<u16> for BitcoinCoreVersion {
+    type Error = u16;
```
bit-aloo: u8 here as well.

bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,319 @@
...
+        tracing::info!(
+            "Validating DeclareMiningJob - version: {:?}, coinbase inputs: {}, outputs: {}, locktime: {}",
+            version,
+            coinbase_tx.input.len(),
+            coinbase_tx.output.len(),
+            coinbase_tx.lock_time.to_consensus_u32()
+        );
+        tracing::debug!(
+            "Declared coinbase scriptSig: {:?}",
+            coinbase_tx.input[0].script_sig
```
bit-aloo: tracing we can import, so we dont have to write full qualified path

pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
```diff
@@ -34,6 +34,7 @@ monitoring_cache_refresh_secs = 15
 # Supported networks: mainnet, testnet4, signet, regtest
 # Default data_dir: ~/.bitcoin (Linux) or ~/Library/Application Support/Bitcoin (macOS)
 [template_provider_type.BitcoinCoreIpc]
+version = 31
```
bit-aloo, line 37: Can't we get this via bitcoin-core instead of mentioning in config?

miner-apps/jd-client/src/lib/mod.rs
```diff
@@ -57,6 +59,13 @@ pub struct JobDeclaratorClient {
 
 #[cfg_attr(not(test), hotpath::measure_all)]
 impl JobDeclaratorClient {
+    fn map_bitcoin_core_version(version: BitcoinCoreIpcVersion) -> BitcoinCoreVersion {
+        match version {
+            BitcoinCoreIpcVersion::V30 => BitcoinCoreVersion::V30X,
+            BitcoinCoreIpcVersion::V31 => BitcoinCoreVersion::V31X,
+        }
+    }
+
```
bit-aloo: lets move it at the bottom.

bit-aloo: Can't we expose this method in stratum-apps

stratum-apps/src/tp_type.rs
```diff
@@ -11,6 +11,37 @@ pub enum BitcoinNetwork {
...
+    V31,
+}
+
+impl BitcoinCoreIpcVersion {
+    pub const fn as_major(self) -> u16 {
+        match self {
+            Self::V30 => 30,
+            Self::V31 => 31,
+        }
+    }
```
bit-aloo: u8 would cover all the ranges.

pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
plebhash: unfortunately no

there's no capnp RPC method that returns Bitcoin Core version, it can only be done via HTTP JSON RPC

plebhash: hmm discussing this with @Shourya742 and @lucasbalieiro on PR review call

@lucasbalieiro pointed out that Bitcoin Core v32 is about to land a legacy-compatible RPC interface over capnp: https://github.com/bitcoin/bitcoin/blob/master/src/ipc/capnp/rpc.capnp

so after v32, we could leverage that for inferring Bitcoin Core version over IPC

---

and that made me wonder whether we could have some kind of heuristics:
- 1. try `getnetworkinfo` RPC over IPC, if it works, we know whether it's v32, v33, v34, etc
  - i.a. if `getnetworkinfo` RPC over IPC isn't available, we know it's either v31 or v30
- 2. try to load `v31` mining interface
  - ii.a. if it works, we know it's `v31`
  - ii.b. if it doesn't, we know it's `v30`

---

I'll do some deeper evaluation on whether it makes sense to start implementing step ii now, so we already have the heuristics ready for when v32 lands

plebhash: I think this has real potential to bring substantial improvements to UX/UI

but this would have cross-repo ramifications, so in the spirit of sticking to the originally planned scope and doing things one step at a time, I feel it's wiser to do this later after more careful evaluation: https://github.com/stratum-mining/sv2-apps/issues/570

stratum-apps/src/tp_type.rs
plebhash: done

bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs
plebhash: this pattern predates this PR

doing over the entire crate via bd4dc11bdc4147b79b8ad45ff4234750e70f4f9a

bitcoin-core-sv2/src/common/mod.rs
plebhash: done

bitcoin-core-sv2/src/common/mod.rs
plebhash: done

miner-apps/jd-client/src/lib/mod.rs
plebhash: > lets move it at the bottom.

done

plebhash: > Can't we expose this method in stratum-apps

`bitcoin_core_sv2` is now a dependency of `stratum-apps` via 3411b783dacfc3953b3e3ba53fc288f5863e38c1

`map_bitcoin_core_version` was completely removed

bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/mempool.rs
GitGab19, line 1: This implementation is exactly the same between v30x and v31x.

I'm not sure I would extract it into the `common` module though, but I just wanted to point this out.

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/monitors.rs
GitGab19, line 1: Similarly to the `mempool.rs` in JDP, this is the same between v30 and v31.

Though, each of this `monitor.rs` imports a specific version of `BitcoinCoreSv2TDP`, so I'm not sure it's so easy to abstract it.

Not a blocker for this PR for sure, I just wanted to point it out, especially in case we want to improve this code duplication in the future.

GitGab19 (commented):
I haven't tested it yet, but it's definitely a very solid and precise direction, great job @plebhash !

I left two minor considerations, which are not blockers for this PR.

I'll test this ASAP and report back here my results.

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/monitors.rs
plebhash: yeah this came across my mind as well, and I don't love the fact that we ended up with so much code duplication

I spent some time trying to imagine an alternative design, but with a deliberate tradeoff of more complex module hierarchy tree... here's the rationale:

over time, as more Bitcoin Core releases come out, we can expect for sure:
- many modules to remain identical across releases (resulting in code duplication)
- some modules to undergo changes across releases

so we could introduce shared modules:
```
unix_capnp
├── mod.rs
├── v30x_v31x                              # provides shared code for v30x + v31x
│   ├── mod.rs                             # exports shared modules
│   ├── job_declaration_protocol
│   │   ├── mod.rs                         # exports shared JDP modules
│   │   └── mempool.rs                     # shared implementation
│   └── template_distribution_protocol
│       ├── mod.rs                         # exports shared TDP modules
│       └── monitors.rs                    # shared implementation
├── v30x
│   ├── mod.rs
│   ├── job_declaration_protocol
│   │   ├── mod.rs
│   │   ├── error.rs
│   │   ├── handlers.rs
│   │   ├── mempool.rs                     # uses shared code from v30x_v31x/job_declaration_protocol/mempool.rs
│   │   └── monitors.rs
│   └── template_distribution_protocol
│       ├── mod.rs
│       ├── error.rs
│       ├── handlers.rs
│       ├── monitors.rs                    # uses shared code from v30x_v31x/template_distribution_protocol/monitors.rs
│       └── template_data.rs
└── v31x
    ├── mod.rs
    ├── job_declaration_protocol
    │   ├── mod.rs
    │   ├── error.rs
    │   ├── handlers.rs
    │   ├── mempool.rs                     # uses shared code from v30x_v31x/job_declaration_protocol/mempool.rs
    │   └── monitors.rs
    └── template_distribution_protocol
        ├── mod.rs
        ├── error.rs
        ├── handlers.rs
        ├── monitors.rs                    # uses shared code from v30x_v31x/template_distribution_protocol/monitors.rs
        └── template_data.rs
```

> I'm not sure I would extract it into the `common` module though

the shared `v30x_v31x` module above is specific for IPC connections, so is is being placed under `unix_capnp`, while leaving `common` for "runtime API" abstractions

`common` could be renamed into `runtime_api` or some other name to make this idea more explicitly clear and avoid confusion

---

now, let's imagine `v32x` comes out, and:
- JDP `mempool` module is different from `v30x_v31x`
- JDP `handlers` module is identical to `v31x` but different from `v30x`
- TDP `monitors` module is identical to `v30x_v31x`

(note: this is purely imaginary so we can speculate on how the module hierarchy would evolve over time... I'm not claiming `v32x` will happen exactly like this)

we would end up with a few different modules for shared code:
- `v30x_v31x`
- `v30x_v31x_v32x`
- `v31x_v32x`

```
unix_capnp
├── mod.rs
├── v30x_v31x                              # provides shared code for v30x + v31x
│   ├── mod.rs                             # exports shared modules
│   └── job_declaration_protocol
│       ├── mod.rs                         # exports shared JDP modules
│       └── mempool.rs                     # shared implementation
├── v31x_v32x                              # provides shared code for v31x + v32x
│   ├── mod.rs                             # exports shared modules
│   └── job_declaration_protocol
│       ├── mod.rs                         # exports shared JDP modules
│       └── handlers.rs                    # shared implementation
├── v30x_v31x_v32x                         # provides shared code for v30x + v31x + v32x
│   ├── mod.rs                             # exports shared modules
│   └── template_distribution_protocol
│       ├── mod.rs                         # exports shared TDP modules
│       └── monitors.rs                    # shared implementation
├── v30x
│   ├── mod.rs
│   ├── job_declaration_protocol
│   │   ├── mod.rs
│   │   ├── error.rs
│   │   ├── handlers.rs
│   │   ├── mempool.rs                     # uses shared code from v30x_v31x/job_declaration_protocol/mempool.rs
│   │   └── monitors.rs
│   └── template_distribution_protocol
│       ├── mod.rs
│       ├── error.rs
│       ├── handlers.rs
│       ├── monitors.rs                    # uses shared code from v30x_v31x_v32x/template_distribution_protocol/monitors.rs
│       └── template_data.rs
├── v31x
│   ├── mod.rs
│   ├── job_declaration_protocol
│   │   ├── mod.rs
│   │   ├── error.rs
│   │   ├── handlers.rs                    # uses shared code from v31x_v32x/job_declaration_protocol/handlers.rs
│   │   ├── mempool.rs                     # uses shared code from v30x_v31x/job_declaration_protocol/mempool.rs
│   │   └── monitors.rs
│   └── template_distribution_protocol
│       ├── mod.rs
│       ├── error.rs
│       ├── handlers.rs
│       ├── monitors.rs                    # uses shared code from v30x_v31x_v32x/template_distribution_protocol/monitors.rs
│       └── template_data.rs
└── v32x
    ├── mod.rs
    ├── job_declaration_protocol
    │   ├── mod.rs
    │   ├── error.rs
    │   ├── handlers.rs                    # uses shared code from v31x_v32x/job_declaration_protocol/handlers.rs
    │   ├── mempool.rs
    │   └── monitors.rs
    └── template_distribution_protocol
        ├── mod.rs
        ├── error.rs
        ├── handlers.rs
        ├── monitors.rs                    # uses shared code from v30x_v31x_v32x/template_distribution_protocol/monitors.rs
        └── template_data.rs
```

---

in summary, the pattern here is that we would end up with multiple permutations of modules for shared code

at some point, we expect Bitcoin Core to become stable... so I don't necessarily expect complexity to blow up exponentially over time

but in the short/mid term, I'd also expect a substantial amount of new permutations popping up

plebhash: I'm not opinionated either way

I could adapt this PR to go in this direction, and I'm also fine with living with some code duplication on the short term, while keeping the rationale above as a potential solution for this problem as time goes on

plebhash:
after we merged #553, plus with the addition of f24d1237beb60ca76b7074b55a22f60a761ddc9d, it seems we're substantially increasing the build artifact size, which is causing CI to break: https://github.com/stratum-mining/sv2-apps/actions/runs/27780182723/job/82202664566?pr=548

so I'm introducing f4a88f85bed914bf459e316c45b2117cd74844b7 to reduce disk pressure on the github runner

GitGab19:
Why didn't we get CI failures on #553 though?

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/monitors.rs
GitGab19: For now I'm fine with this duplication, let's see in the future if that can becomes a problem.

The good thing is that your pattern described above would solve it.

plebhash:
> Why didn't we get CI failures on #553 though?

I asked `gpt-5.3-codex` a similar question earlier:

> 
> Great question. Short answer: **PR #573 was close, but still below the runner disk limit; PR #548 crossed it**.
> 
> What’s different:
> 
> - PR #573 (`run 27775350660`, MSRV `job 82185981896`) used the same workflow and succeeded, including `integration-tests --no-run`.
> - In #548 (`run 27780182723`, `job 82202664566`), logs show extra compile footprint from the multi-version runtime stack:
>   - repeated `capnp v0.21.7` builds across multiple workspaces (in addition to `v0.25.x`).
>   - PR #573 log does not show `capnp v0.21.x`.
> - Your branch also enables the facade feature in integration tests (`integration-tests/Cargo.toml:17` includes `bitcoin-core-sv2`), which increases what gets compiled/linked in that job.
> - MSRV runs all workspace builds and then all `test --no-run` in one runner (`.github/workflows/msrv.yaml:23` through `.github/workflows/msrv.yaml:58`), so artifact accumulation is the killer.
> - Result: PR #573 stayed under threshold; #548 hit `/usr/bin/ld ... No space left on device`.
> 
> So this is mostly a **capacity threshold + branch footprint** issue, not a “new code error.”

IIUC this PR is already increasing the build artifact footprint due to:
- multiple `capnp` + `capnp-rpc` crate versions on the dependency list
- more source code

then #553 increased it further with the addition of `asic-rs`

and now the combination of both started pushing the github runners over the limit

plebhash:
also getting similar disk-out-of-space errors after merging #531:

- https://github.com/stratum-mining/sv2-apps/actions/runs/27786973093/job/82226027941
- https://github.com/stratum-mining/sv2-apps/actions/runs/27786973066/job/82226028010

so maybe https://github.com/stratum-mining/sv2-apps/commit/f4a88f85bed914bf459e316c45b2117cd74844b7 is not enough, and we might continue to get flaky PITA over this.... let's see

.github/workflows/msrv.yaml
GitGab19:
```suggestion
        run: cargo test --manifest-path=pool-apps/Cargo.toml --all-features --no-run
```

GitGab19:
```suggestion
        run: cargo test --manifest-path=miner-apps/Cargo.toml --all-features --no-run
```

GitGab19: JDS is part of the `pool-apps` workspace, so there's no need for this independent step anymore
```suggestion
```

GitGab19 (approved):
Besides of my last comments, which are only about the `msrv.yaml` workflow changes, I just tested with both v30.2 and v31, and everything worked fine!

tACK

.github/workflows/msrv.yaml
plebhash: done

plebhash: done

plebhash: removed

GitGab19 (approved):

.github/workflows/msrv.yaml
```diff
@@ -30,9 +30,6 @@ jobs:
       - name: Check test compilation for pool
         run: cargo test --manifest-path=pool-apps/Cargo.toml --all-features --no-run
 
-      - name: Check test compilation for jd-server
-        run: cargo test --manifest-path=pool-apps/jd-server/Cargo.toml --no-run
-
```
bit-aloo, line 50: this makes sense, can you do the same in ci,yaml and lockfile.yaml

pool-apps/pool/src/lib/mod.rs
```diff
@@ -241,6 +246,7 @@ impl PoolSv2 {
 
                 let bitcoin_core_cancellation_token = CancellationToken::new();
                 let bitcoin_core_config = BitcoinCoreSv2TDPConfig {
+                    version: version.to_bitcoin_core_version(),
```
bit-aloo: why we have two different versioning shouldn't they be the same?

stratum-apps/src/tp_type.rs
```diff
@@ -26,6 +26,12 @@ impl BitcoinCoreIpcVersion {
             Self::V31 => 31,
         }
     }
+
+    #[cfg(feature = "bitcoin-core-sv2")]
+    pub fn to_bitcoin_core_version(self) -> bitcoin_core_sv2::common::BitcoinCoreVersion {
+        bitcoin_core_sv2::common::BitcoinCoreVersion::try_from(self.as_major())
+            .expect("BitcoinCoreIpcVersion and BitcoinCoreVersion majors must stay aligned")
+    }
```
bit-aloo: Shouldn't this be a from instead of try_from, considering we already converted from config to BitcoinCoreIpcVersion enum and after  than any conversion should be lossless.

stratum-apps/src/tp_type.rs
```diff
@@ -26,6 +26,12 @@ impl BitcoinCoreIpcVersion {
...
 }
 
 impl TryFrom<u16> for BitcoinCoreIpcVersion {
```
bit-aloo: Shouldn't this be u8?

bitcoin-core-sv2/src/common/template_distribution_protocol/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+#[async_trait::async_trait(?Send)]
+pub trait Sv2TemplateDistributionProtocol {
+    async fn run(&mut self);
+}
+
+/// Version-agnostic TDP runtime handle.
+///
+/// Instances are created with [`new`], which selects the concrete backend for the requested
+/// [`BitcoinCoreVersion`].
+pub type BitcoinCoreSv2TDP = Box<dyn Sv2TemplateDistributionProtocol>;
```
bit-aloo: We can model it via enums having a common interface, something like this:

```rust
pub enum BitcoinCoreSv2TDP {
    V30X(v30x::template_distribution_protocol::BitcoinCoreSv2TDP),
    V31X(v31x::template_distribution_protocol::BitcoinCoreSv2TDP),
}

impl BitcoinCoreSv2TDP {
    pub async fn run(&mut self) {
        match self {
            Self::V30X(runtime) => runtime.run().await,
            Self::V31X(runtime) => runtime.run().await,
        }
    }
}
```

bit-aloo: Would help in removal of `async_trait` and vtable lookup

bitcoin-core-sv2/src/common/template_distribution_protocol/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        write!(
+            f,
+            "failed to initialize bitcoin_core_sv2 TDP for v{}: {}",
+            self.version.as_major(),
+            self.details
+        )
+    }
+}
+
+impl std::error::Error for BitcoinCoreSv2TDPError {}
```
bit-aloo: jd_server part also does have this very specific and similar error type, maybe we can define a common error type in common module which can be protocol specific, something like this: 

```rust
enum Protocol {
 TDP,
 JDS
}
/// Error returned when selecting and initializing a versioned Bitcoin Core IPC runtime fails.
#[derive(Debug)]
pub struct BitcoinCoreSv2Error<Protocol> {
    version: BitcoinCoreVersion,
    details: String,
}

impl BitcoinCoreSv2Error<P> {
    pub(crate) fn from_debug<E>(
        version: BitcoinCoreVersion,
        error: E,
    ) -> Self
    where
        E: fmt::Debug,
    {
        Self {
            version,
            details: format!("{error:?}"),
        }
    }
}

impl fmt::Display for BitcoinCoreSv2Error<P> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(
            f,
            "failed to initialize bitcoin_core_sv2 {} for v{}: {}",
            P,
            self.version.as_major(),
            self.details
        )
    }
}

impl std::error::Error for BitcoinCoreSv2Error<P> {}

```

bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait(?Send)]
+impl Sv2JobDeclarationProtocol for BitcoinCoreSv2JDP {
+    async fn run(&self) {
+        BitcoinCoreSv2JDP::run(self).await;
+    }
+}
```
bit-aloo: This can be removed if we have enum dispatch.

bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mod.rs
```diff
@@ -437,10 +443,17 @@ impl BitcoinCoreSv2JDP {
...
         Ok(())
     }
 }
+
+#[async_trait::async_trait(?Send)]
+impl Sv2JobDeclarationProtocol for BitcoinCoreSv2JDP {
+    async fn run(&self) {
+        BitcoinCoreSv2JDP::run(self).await;
+    }
+}
```
bit-aloo: We can remove this with the enum dispatch,

bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs
```diff
@@ -814,3 +821,10 @@ mod tests {
         assert_eq!(coinbase_tx.output, vec![required_output]);
     }
 }
+
+#[async_trait::async_trait(?Send)]
+impl Sv2TemplateDistributionProtocol for BitcoinCoreSv2TDP {
+    async fn run(&mut self) {
+        BitcoinCoreSv2TDP::run(self).await;
+    }
+}
```
bit-aloo: This can removed with enum dispatch.

bitcoin-core-sv2/src/common/job_declaration_protocol/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+#[async_trait::async_trait(?Send)]
+pub trait Sv2JobDeclarationProtocol {
+    async fn run(&self);
+}
+
+/// Version-agnostic JDP runtime handle.
+///
+/// Instances are created with [`new`], which selects the concrete backend for the requested
+/// [`BitcoinCoreVersion`].
+pub type BitcoinCoreSv2JDP = Box<dyn Sv2JobDeclarationProtocol>;
```
bit-aloo: The enum variant for this can be very similar?

```rust
pub enum BitcoinCoreSv2JDP {
    V30X(v30x::job_declaration_protocol::BitcoinCoreSv2JDP),
    V31X(v31x::job_declaration_protocol::BitcoinCoreSv2JDP),
}

impl BitcoinCoreSv2JDP {
    pub async fn run(&self) {
        match self {
            Self::V30X(runtime) => runtime.run().await,
            Self::V31X(runtime) => runtime.run().await,
        }
    }
}

```

bitcoin-core-sv2/src/common/job_declaration_protocol/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+        write!(
+            f,
+            "failed to initialize bitcoin_core_sv2 JDP for v{}: {}",
+            self.version.as_major(),
+            self.details
+        )
+    }
+}
+
+impl std::error::Error for BitcoinCoreSv2JDPError {}
```
bit-aloo: We can abstract it out with common Error from common module

bitcoin-core-sv2/src/common/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+impl BitcoinCoreVersion {
+    pub const fn as_major(self) -> u8 {
+        match self {
+            Self::V30X => 30,
+            Self::V31X => 31,
+        }
+    }
+}
```
bit-aloo, line 22: Maybe the new methods can become part of the version fields itself, instead of us having to pass version in the new constructor:

```rust
    pub async fn new_jdp<P>(
        self,
        bitcoin_core_unix_socket_path: P,
        incoming_requests: async_channel::Receiver<job_declaration_protocol::io::JdRequest>,
        cancellation_token: tokio_util::sync::CancellationToken,
        ready_tx: tokio::sync::oneshot::Sender<()>,
    ) -> Result<job_declaration_protocol::BitcoinCoreSv2JDP, BitcoinCoreSv2Error>
    where
        P: AsRef<std::path::Path>,
    {
        ...
    }

    /// Builds a TDP runtime for this Bitcoin Core IPC schema family.
    #[allow(clippy::too_many_arguments)]
    pub async fn new_tdp<P>(
        self,
        bitcoin_core_unix_socket_path: P,
        fee_threshold: u64,
        min_interval: u8,
        incoming_messages: async_channel::Receiver<
            stratum_core::parsers_sv2::TemplateDistribution<'static>,
        >,
        outgoing_messages: async_channel::Sender<
            stratum_core::parsers_sv2::TemplateDistribution<'static>,
        >,
        global_cancellation_token: tokio_util::sync::CancellationToken,
    ) -> Result<template_distribution_protocol::BitcoinCoreSv2TDP, BitcoinCoreSv2Error>
    where
        P: AsRef<std::path::Path>,
    {
       ...
    }
```

specific version association attached to version itself.

bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mempool.rs
```diff
@@ -0,0 +1,126 @@
+//! Local mempool mirror for Bitcoin Core v31.x Sv2 Job Declaration Protocol via capnp over UNIX
```
bit-aloo, line 1: Can we move the common module at the same level as versioning module?

bit-aloo: I think its the same file as v30x?

bit-aloo (commented):
Some suggestions and questions

bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mempool.rs
plebhash: that's similar to what's already been discussed here: https://github.com/stratum-mining/sv2-apps/pull/548#discussion_r3435297594

> Can we move the common module at the same level as versioning module?

tbh if we were to do this, we should apply the same pattern across the entire crate, which is the scope of https://github.com/stratum-mining/sv2-apps/issues/574

bit-aloo: I just read the issue, with the current structuring it won't be easy for us to remove duplication. I am now thinking whether it will be viable to have version based distinct implementation, like in a method, based on version we can have different implementation (via match). What you think of this?

plebhash: when I discussed this with @GitGab19 the consensus was to leave code de-duplication as something to be solved later.

I also don't fully understand what you're trying to say on the comment above.

> with the current structuring it won't be easy for us to remove duplication. 

by "current structuring" do you mean what is being done on this PR or what's proposed on #574?

> I am now thinking whether it will be viable to have version based distinct implementation, like in a method, based on version we can have different implementation (via match). What you think of this?

assuming this suggestion is meant to be taken as an alternative approach to #574, I first need to understand what you understand as a problem with what was proposed there

and also, do you think this is something that needs to be addressed now?

if not, we should discuss it under #574 and not here

pool-apps/pool/src/lib/mod.rs
plebhash: I also don't love the duplication of  `stratum-apps::tp_type::BitcoinCoreIpcVersion` vs `bitcoin_core_sv2::common::BitcoinCoreVersion`

we can get rid of `stratum-apps::tp_type::BitcoinCoreIpcVersion` in favor of `bitcoin_core_sv2::common::BitcoinCoreVersion`, but with the tradeoff of having `stratum-apps::tp_type::TemplateProviderType::BitcoinCoreIpc` variant behind `bitcoin-core-sv2` feature gate

tbh, that feels like a reasonable tradeoff, implementing it

bitcoin-core-sv2/src/common/template_distribution_protocol/mod.rs
plebhash: when I originally wrote the scope of #516 I explicitly mentioned the introduction of traits on the `common` module, which would serve the purpose of establishing clear interfaces for which all different version implementations would follow

but the implementation converged towards a very thin trait interface (limited to a single `run` method), which honestly makes it not so much valuable anymore

I think this suggestion makes sense, since we'd still be following the same `run` pattern, while enforcing it inside the `match`

so yeah thanks for the suggestion @Shourya742 implementing it

.github/workflows/msrv.yaml
plebhash: done

bitcoin-core-sv2/src/common/mod.rs
plebhash: I experimented with this but I didn't like the result... adding a `new_jdp`/`new_tdp` constructors to `BitcoinCoreVersion` feels like we're mixing abstractions

keeping as-is

stratum-apps/src/tp_type.rs
plebhash: based on the adaptations from https://github.com/stratum-mining/sv2-apps/pull/548#discussion_r3455144654, this is no longer relevant

stratum-apps/src/tp_type.rs
plebhash: based on the adaptations from https://github.com/stratum-mining/sv2-apps/pull/548#discussion_r3455144654, this is no longer relevant

bitcoin-core-sv2/src/common/template_distribution_protocol/mod.rs
plebhash: done

bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/mod.rs
plebhash: done

bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mod.rs
plebhash: done

bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs
plebhash: done

bitcoin-core-sv2/src/common/job_declaration_protocol/mod.rs
plebhash: done

bitcoin-core-sv2/src/common/job_declaration_protocol/mod.rs
plebhash: done

bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mempool.rs
bit-aloo: I think this is relevant to this PR because it introduces the module structure that #574 is proposing to build on.

My concern is that once we start organizing code around version-specific modules and shared-version modules, removing duplication later may become harder. As mentioned in #574, this could naturally lead to modules such as v30x_v31x, v31x_v32x, v30x_v31x_v32x, etc., where compatibility between versions becomes part of the directory structure itself.

I am not sure this will scale well in the long run. As we add support for more versions, we may end up with more combinations of shared modules, making it harder to understand where code should live and how different versions are related.

I would rather keep the module hierarchy aligned with the protocol concepts (such as TDP and JDP) and handle version-specific behavior closer to the implementation itself, for example using version-aware methods or match statements where the behavior actually differs.

Because of that, I see this as more than just a future cleanup or deduplication concern. The structure being introduced here will influence how we extend and maintain this code later, so I think it's worth discussing whether this is the direction we want to take before we expand it further.

bitcoin-core-sv2/Cargo.toml
```diff
@@ -13,15 +13,21 @@ keywords = ["stratum", "mining", "bitcoin", "bitcoin-core"]
...
 tracing = "0.1.41"
 tracing-subscriber = "0.3.19"
-capnp = "0.25.0"
-capnp-rpc = "0.25.0"
 tokio = { version = "1.44.1", features = ["full"] }
 tokio-util = { version = "0.7.10", features = ["codec", "compat"] }
 async-channel = "1.5.1"
 
+# Bitcoin Core v30.x IPC dependencies
+capnp_v30 = { package = "capnp", version = "0.21.4" }
```
bit-aloo: Damn... We should backport capnp version 0.25.0 for bitcoin_capnp_type_v30 variant. Pulling same lib with different version just bulks us up.... We can do a minor release for v0.1.1 bitcoin_capnp_type_v30, and then we can remove this.

bitcoin-core-sv2/src/common/mod.rs
bit-aloo: I thought it would look good. No hard opinions on this.

bitcoin-core-sv2/Cargo.toml
bit-aloo: Released v0.1.1 for bitcoin-capnp-type: https://crates.io/crates/bitcoin-capnp-types/0.1.1, we can remove capnp and capnp_rpc variants now.

bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mempool.rs
bit-aloo: I discussed this with Plebhash, and we spent some time exploring the solution space. Through that discussion, I gained a better understanding of the rationale behind the current structure and how my suggestion could either break or introduce additional complexity if we need to change the object structure in the future, potentially making the crate harder to reason about.

Given that, I think we can move forward with the current implementation. While it does involve some duplication, it appears to be the simpler and more maintainable approach from a reasoning and evolution standpoint.

bitcoin-core-sv2/Cargo.toml
plebhash: new dependency layout:

`Cargo.toml`:
```toml
...
# Bitcoin Core v30.x IPC dependencies
bitcoin_capnp_types_v30 = { package = "bitcoin-capnp-types", version = "0.1.2" }

# Bitcoin Core v31.x IPC dependencies
bitcoin_capnp_types_v31 = { package = "bitcoin-capnp-types", version = "0.2.1" }
...
```

(both use the same underlying `0.25.0`)

`*.rs`:
```rust
use bitcoin_capnp_types::{
    capnp,
    capnp_rpc::{RpcSystem, rpc_twoparty_capnp, twoparty},
    ...
}
```

bit-aloo (approved):
ACK

Issue #539, opened 2026-06-03
version bumps for `v0.6.0`


GitGab19:
Similar to https://github.com/stratum-mining/sv2-apps/issues/487

GitGab19:
https://github.com/stratum-mining/sv2-apps/pull/553 is bumping:
- `stratum-apps` from `0.5.0` to `0.6.0`
- `jd_client_sv2` from `0.3.0` to `0.4.0`
- `translator_sv2` from `0.3.0` to `0.4.0`

plebhash:
#575 is bumping:
- `pool_sv2` from `0.4.0` to `0.5.0`

plebhash:
#548 is bumping:
- `bitcoin_core_sv2` from `0.3.0` to `0.4.0`
- `jd_server_sv2` from `0.3.0` to `0.4.0`

plebhash:
#553 + #575 + #548 should have bumped `integration_tests_sv2`, but they didn't

doing it via #605

PR #553, opened 2026-06-11
add `asic-rs` telemetry data in tProxy and JDC monitoring

### LINKED ISSUE #443: use `asic-rs` crate to enrich monitoring with miners' data
In https://github.com/stratum-mining/sv2-ui/issues/70 we started discussing some approaches to leverage `asic-rs` (https://github.com/256foundation/asic-rs) to show more detailed data of miners connected to our stack (initially mainly Sv1-fw machines connected to our Translator Proxy).

We already have a `monitoring` module inside our `stratum-apps` crate, enabled by its feature flag, which spins up a HTTP server exposing some monitoring APIs and a Prometheus endpoint.

The idea here is to integrate `asic-rs` functionalities exactly in that part of our stack, adding more data to our current API endpoints, or creating new ones.

We can do many things with `asic-rs` (big shoutout to @b-rowan for that), though, I'd proceed with a gradual approach.

So I would open specific sub-issues to track individual features we want to leverage from it.

### LINKED ISSUE #445: Add miners' data to `/api/v1/sv1/clients` endpoint of tProxy
We should enrich the `/api/v1/sv1/clients` endpoint of tProxy with some miners' data.

Below a screenshot of current data we're getting from that endpoint:
Image

I think we should be able to show individual miners':
- reported hashrate
- reported power consumption
- reported temps (maybe?)

### LINKED ISSUE #567: add `asic-rs` telemetry data to JDC (and Pool?)
As noted in [this comment](https://github.com/stratum-mining/sv2-apps/pull/553#issuecomment-4727894535), there's probably value in adding `asic-rs` telemetry data to JDC (and Pool?) as well.

GitGab19:
This PR adds `asic-rs` as a dependency and uses its APIs to enrich tProxy and JDC monitoring data with telemetry data of every miner supported by `asic-rs`.

I had to bump MSRV from `1.85` to `1.88`, as we already discussed in https://github.com/stratum-mining/sv2-apps/issues/443.

Here the results with my Bitaxe Gamma connected to tProxy:

Screenshot From 2026-06-11 11-50-33

cc @b-rowan @pavlenex 

Closes #445 
Closes #567 
Closes #443


miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,196 @@
...
+                    .downstream_data
+                    .safe_lock(|data| (*downstream.key(), data.connection_ip))
+                    .ok()
+            })
+            .collect()
+    }
+}
+
+async fn fetch_miner_telemetry(factory: &MinerFactory, ip: IpAddr) -> Option<MinerTelemetry> {
+    let miner = match factory.scan_miner(ip).await {
```
b-rowan: Should use `get_miner` here, `scan_miner` is for if you don't know if a miner exists at that IP.

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,196 @@
...
+            debug!("No miner management interface discovered at {ip}");
+            return None;
+        }
+        Err(error) => {
+            debug!("Failed to discover miner management interface at {ip}: {error}");
+            return None;
+        }
+    };
+
+    Some(miner_data_to_telemetry(miner.get_data().await))
```
b-rowan: Could use `get_data_filtered` here to pass an `exclude` list, that way you don't gather more data than needed and it prevents extra load on the network.  Better for large networks.

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,196 @@
...
+            .map(|hashrate| hashrate.as_unit(HashRateUnit::Hash).value),
+        power_consumption_w: data.wattage.map(|power| power.as_watts()),
+        efficiency_j_per_th: data.efficiency,
+        average_temperature_c: data
+            .average_temperature
+            .map(|temperature| temperature.as_celsius()),
+        uptime_secs: data.uptime.map(|uptime| uptime.as_secs()),
+        is_mining: Some(data.is_mining),
+    }
+}
```
b-rowan: Is it worth moving this to a `From` impl on the `MinerTelemetry` struct?  Would need to be alongside the struct and feature gated...

 ```
impl From<MinerData> for MinerTelemetry ...
```

File: miner-apps/translator/src/lib/sv1/monitoring.rs (hunk 2 above)
Comment (b-rowan) [reply miner-apps/translator/src/lib/sv1/monitoring.rs]: https://docs.rs/asic-rs-core/latest/asic_rs_core/traits/miner/trait.GetMinerData.html#tymethod.get_data_filtered

File: miner-apps/translator/src/lib/sv1/monitoring.rs (hunk 1 above)
Comment (b-rowan) [reply miner-apps/translator/src/lib/sv1/monitoring.rs]: https://docs.rs/asic-rs/latest/asic_rs/factory/struct.MinerFactory.html#method.get_miner

File: miner-apps/translator/src/lib/sv1/monitoring.rs (hunk 3 above)
Comment (GitGab19) [reply miner-apps/translator/src/lib/sv1/monitoring.rs]: This is a very good point, but I see a couple of problems:
- `MinerTelemetry` lives in `stratum-apps` while `MinerData` comes from `asic-rs-core`
- implementing `From<MinerData>` next to the schema would require adding an `asic-rs-core` dependency to `stratum-apps`. 

Since only tProxy currently sources telemetry from `asic-rs`, I kept the conversion there to avoid coupling the monitoring schema in `stratum-apps` to `asic-rs-core`.

File: miner-apps/translator/src/lib/sv1/monitoring.rs (hunk 1 above)
Comment (GitGab19) [reply miner-apps/translator/src/lib/sv1/monitoring.rs]: Thanks a lot @b-rowan, changed it.

File: miner-apps/translator/src/lib/sv1/monitoring.rs (hunk 2 above)
Comment (GitGab19) [reply miner-apps/translator/src/lib/sv1/monitoring.rs]: Thanks a lot @b-rowan, changed it.

File: miner-apps/translator/src/lib/sv1/monitoring.rs (hunk 3 above)
Comment (b-rowan) [reply miner-apps/translator/src/lib/sv1/monitoring.rs]: Understandable, but that's why I thought feature gating it behind the monitoring flag made sense.  No worries in either case.

**Review by b-rowan — APPROVED** [review]

File: miner-apps/translator/src/lib/sv1/downstream.rs [hunk 4]
```diff
@@ -68,6 +69,7 @@ pub struct DownstreamData {
     pub extranonce2_len: usize,
     pub target: Target,
     pub hashrate: Option<Hashrate>,
+    pub connection_ip: IpAddr,
```
bit-aloo, line 74: We should have these feature gated with monitoring.

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -89,7 +91,7 @@ pub struct DownstreamData {
 }
 
 impl DownstreamData {
-    pub fn new(hashrate: Option<Hashrate>, target: Target) -> Self {
+    pub fn new(hashrate: Option<Hashrate>, target: Target, connection_ip: IpAddr) -> Self {
```
bit-aloo: Same here, feature gated.

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -98,6 +100,7 @@ impl DownstreamData {
             extranonce2_len: 4,
             target,
             hashrate,
+            connection_ip,
```
bit-aloo, line 110: Same here.

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -240,9 +243,14 @@ impl Downstream {
...
         hashrate: Option<Hashrate>,
+        connection_ip: IpAddr,
         downstream_cancellation_token: CancellationToken,
     ) -> Self {
-        let downstream_data = Arc::new(Mutex::new(DownstreamData::new(hashrate, target)));
+        let downstream_data = Arc::new(Mutex::new(DownstreamData::new(
+            hashrate,
+            target,
+            connection_ip,
+        )));
```
bit-aloo: Same for these, we can feature gate them

stratum-apps/src/monitoring/sv1.rs
```diff
@@ -11,6 +11,7 @@ use utoipa::ToSchema;
 pub struct Sv1ClientInfo {
     pub client_id: usize,
     pub channel_id: Option<u32>,
+    pub connection_ip: Option<String>,
```
bit-aloo: Should we have tp for this as IpAddr?

bit-aloo: Like `connection_ip: Option<IpAddr>`

stratum-apps/src/monitoring/sv1.rs
```diff
@@ -20,6 +21,32 @@ pub struct Sv1ClientInfo {
...
     pub version_rolling_mask: Option<String>,
     pub version_rolling_min_bit: Option<String>,
+    pub miner_telemetry: Option<MinerTelemetry>,
+}
+
+/// Telemetry reported by the miner's management interface.
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct MinerTelemetry {
+    /// IP address queried for miner management telemetry.
+    pub ip: String,
```
bit-aloo: Can be made `ip: IpAddr`

.github/workflows/binary-release.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: dtolnay/rust-toolchain@1.85
+      - uses: dtolnay/rust-toolchain@1.88
```
bit-aloo, line 37: We can point it to current stable version, considering we changing it anyway

bit-aloo: I think current is stable is 1.96

.github/workflows/ci.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
         run: sudo apt-get install -y capnproto libcapnp-dev
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@1.85
+        uses: dtolnay/rust-toolchain@1.88
```
bit-aloo, line 26: We change this to 1.96 as well.

bit-aloo: And the same for rest.

miner-apps/translator/Cargo.toml
```diff
@@ -32,10 +32,11 @@ clap = { version = "4.5.39", features = ["derive"] }
 hex = "0.4.3"
 hotpath = "0.14.0"
 dashmap = "6.1.0"
+asic-rs = { version = "0.6.0", optional = true }
```
bit-aloo: why not making it part of monitoring and exposing from there?

bit-aloo: Under a feature flag

miner-apps/translator/src/lib/mod.rs
```diff
@@ -94,6 +94,86 @@ impl TranslatorSv2 {
...
+        cancellation_token: CancellationToken,
+        fallback_coordinator: FallbackCoordinator,
+        task_manager: Arc<TaskManager>,
+        action: &str,
+    ) {
+        let Some(monitoring_addr) = self.config.monitoring_address() else {
+            return;
+        };
+
+        info!("{action} monitoring server on http://{}", monitoring_addr);
```
bit-aloo:
```rust
info!("{action} monitoring server on http://{monitoring_addr}");
```

miner-apps/translator/src/lib/mod.rs
```diff
@@ -94,6 +94,86 @@ impl TranslatorSv2 {
...
+    #[cfg(feature = "monitoring")]
+    fn start_monitoring_tasks(
+        &self,
+        channel_manager: Arc<ChannelManager>,
+        sv1_server: Arc<Sv1Server>,
+        cancellation_token: CancellationToken,
+        fallback_coordinator: FallbackCoordinator,
+        task_manager: Arc<TaskManager>,
+        action: &str,
+    ) {
```
bit-aloo: Can we return an error from this?

miner-apps/translator/src/lib/sv1_monitoring.rs
```diff
@@ -37,13 +68,150 @@ impl Sv1ClientsMonitoring for Sv1Server {
...
+                        }
+                        _ = fallback_token.cancelled() => {
+                            info!("SV1 miner telemetry loop received fallback signal");
+                            break 'telemetry;
+                        }
+                        _ = self.refresh_miner_telemetry(&factory) => {}
+                    }
+                }
+            }
+        }
```
bit-aloo: We can remove 'telemetry in the loop

miner-apps/translator/src/lib/sv1_monitoring.rs
```diff
@@ -37,13 +68,150 @@ impl Sv1ClientsMonitoring for Sv1Server {
...
+            .hashrate
+            .map(|hashrate| hashrate.as_unit(HashRateUnit::Hash).value),
+        power_consumption_w: data.wattage.map(|power| power.as_watts()),
+        efficiency_j_per_th: data.efficiency,
+        average_temperature_c: data
+            .average_temperature
+            .map(|temperature| temperature.as_celsius()),
+        uptime_secs: data.uptime.map(|uptime| uptime.as_secs()),
+        is_mining: Some(data.is_mining),
     }
```
bit-aloo: From impl can be used?

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -37,13 +68,150 @@ impl Sv1ClientsMonitoring for Sv1Server {
...
+
+    async fn refresh_miner_telemetry(&self, factory: &MinerFactory) {
+        let downstreams = self.current_downstream_connection_ips();
+        let active_downstream_ids = downstreams
+            .iter()
+            .map(|(downstream_id, _)| *downstream_id)
+            .collect::<HashSet<_>>();
+
+        self.miner_telemetry
+            .retain(|downstream_id, _| active_downstream_ids.contains(downstream_id));
```
bit-aloo, line 126: We can remove miner_telemetry from sv1_server from downstream shutdown callback, then we wont need this check

miner-apps/translator/src/lib/mod.rs
```diff
@@ -94,6 +94,86 @@ impl TranslatorSv2 {
...
 
+    #[cfg(feature = "monitoring")]
+    fn start_monitoring_tasks(
+        &self,
+        channel_manager: Arc<ChannelManager>,
+        sv1_server: Arc<Sv1Server>,
+        cancellation_token: CancellationToken,
+        fallback_coordinator: FallbackCoordinator,
+        task_manager: Arc<TaskManager>,
+        action: &str,
```
bit-aloo: This action can be removed, the associated logs can put at the callsite.

bit-aloo (commented):
For some reason I am not able to add suggestions. weird. Did a first pass of code review.

.github/workflows/binary-release.yaml
GitGab19: I'm bumping MSRV to 1.88 because of the new `asic-rs` dependency, but bumping it to the latest stable (1.96) is a different decision.

Tbh I would not do it.

miner-apps/translator/src/lib/sv1_monitoring.rs
GitGab19: Yeah this is the same point brought up by @b-rowan here: https://github.com/stratum-mining/sv2-apps/pull/553#discussion_r3396879154

I'll re-evaluate the design soon.

GitGab19: My main push-back is about having `stratum-apps` depending (even though behind a feature-flag) on `asic-rs-core`, just for this From implementation.

Up to now, `stratum-apps` only defines the schemas and traits to enable monitoring at the apps level, which is a clear separation of concerns.

.github/workflows/binary-release.yaml
bit-aloo: On sv2-apps, if I remember correctly, the reason we pinned the MSRV was that new stable compiler releases would sometimes introduce new clippy lints, causing the clippy workflow to fail. As a result, our PRs would occasionally start failing in CI whenever a new compiler version was released.

We wanted to avoid that, so we pinned the compiler version.

Updating it to the latest stable version should be ok?

miner-apps/translator/src/lib/sv1_monitoring.rs
bit-aloo: Let me think about it...

.github/workflows/binary-release.yaml
GitGab19: Why do we need to bump it to the latest stable when we can support earlier versions?

miner-apps/translator/src/lib/sv1/monitoring.rs
GitGab19: I added the `From` impl behind a new `asic-rs-telemetry` feature flag on `stratum-apps`.

miner-apps/translator/src/lib/sv1_monitoring.rs
GitGab19: I added the `From` impl behind a new `asic-rs-telemetry` feature flag on `stratum-apps`.

miner-apps/translator/Cargo.toml
GitGab19: I added the `asic-rs-telemetry` feature flag on `stratum-apps`,  which is called here now.

That feature flag enables only the `asic-rs-core` dependency on `stratum-apps` (needed for the `From` impl on the schema/struct), but tProxy still needs the full `asic-rs` crate for its implementation.

.github/workflows/binary-release.yaml
bit-aloo: Hmmm, I guess Mujina now uses stratum-apps and integration-test-sv2 as dependencies, so keeping a lower MSRV would be ideal for them.

On our side, this means we won't be able to use any newer compiler features, but I don't think that will affect us much. Our implementation is fairly stable now and doesn't require many changes. The only thing I was looking forward to was using the alloc APIs once they become stable in Rust.

Since you're hesitant about bumping the MSRV, I won't push for it. It doesn't really benefit us at the moment unless we eventually need a newer compiler feature.

miner-apps/translator/src/lib/mod.rs
GitGab19: Done

miner-apps/translator/src/lib/mod.rs
GitGab19: Done

miner-apps/translator/src/lib/sv1/monitoring.rs
GitGab19: I think we still need this guard.

The refresh loop snapshots `(downstream_id, ip)` before awaiting the ASIC request. If the downstream disconnects while that request is in flight, the disconnect callback removes the telemetry, but the fetch could then return and reinsert stale data.

So the callback handles cleanup on disconnect, while this check prevents stale reinsertion after an async fetch.

bit-aloo: hmm... I get your point, very similar to discussion we had on Pool PR. This will make the miner_telemetry eventually consistent because of preflight request metioned.

plebhash:
we should bump MAJOR for `stratum-apps` + `translator_sv2` (and take note on #539)

plebhash:
tACK

a few observations about `asic-rs` cc @b-rowan :

- Braiins BMM100 shows with `"make": "Antminer"`
- FutureBit Apollo II doesn't get recognized

```
curl -s "http://127.0.0.1:9092/api/v1/sv1/clients" | jq

{
  "offset": 0,
  "limit": 100,
  "total": 3,
  "items": [
    {
      "client_id": 8,
      "channel_id": 2,
      "connection_ip": "192.168.15.31",
      "authorized_worker_name": "bmm100",
      "user_identity": "bmm100",
      "target_hex": "00000000001c25c26a38a078f9fb2209d291fa59962595bdb8f43c38a611f2d6",
      "hashrate": 1000000000000.0,
      "stable_hashrate": true,
      "extranonce1_hex": "0100fbaf000000000000000000000001",
      "extranonce2_len": 4,
      "version_rolling_mask": "1fffe000",
      "version_rolling_min_bit": "00000010",
      "miner_telemetry": {
        "ip": "192.168.15.31",
        "make": "Antminer",
        "model": "BRAIINS MINI MINER BMM 100",
        "firmware_version": "2026-04-14-0-912d084c-26.04-plus",
        "reported_hashrate_hs": 1012813471661.3801,
        "power_consumption_w": 41.0,
        "efficiency_j_per_th": 40.48129408542047,
        "average_temperature_c": null,
        "uptime_secs": 944,
        "is_mining": true
      }
    },
    {
      "client_id": 1,
      "channel_id": 1,
      "connection_ip": "192.168.15.27",
      "authorized_worker_name": "x",
      "user_identity": "x",
      "target_hex": "0000000000285b574ecf1422f0a9b9585ec2e509ae1a49b2a9bb43e12822b9f7",
      "hashrate": 697466000000.0,
      "stable_hashrate": true,
      "extranonce1_hex": "0100fbaf000000000000000000000000",
      "extranonce2_len": 4,
      "version_rolling_mask": "1fffe000",
      "version_rolling_min_bit": "00000000",
      "miner_telemetry": {
        "ip": "192.168.15.27",
        "make": "Bitaxe",
        "model": "Gamma",
        "firmware_version": "v2.14.0b3",
        "reported_hashrate_hs": 877891296400.0,
        "power_consumption_w": 14.4696045,
        "efficiency_j_per_th": 16.482227992618245,
        "average_temperature_c": 61.0,
        "uptime_secs": 1173,
        "is_mining": true
      }
    },
    {
      "client_id": 17,
      "channel_id": 3,
      "connection_ip": "192.168.15.7",
      "authorized_worker_name": "apollo",
      "user_identity": "apollo",
      "target_hex": "00000000001c25c26a38a078f9fb2209d291fa59962595bdb8f43c38a611f2d6",
      "hashrate": 1000000000000.0,
      "stable_hashrate": true,
      "extranonce1_hex": "0100fbaf000000000000000000000003",
      "extranonce2_len": 4,
      "version_rolling_mask": "1fffe000",
      "version_rolling_min_bit": "00000010",
      "miner_telemetry": null
    }
  ]
}
```

b-rowan:
> Braiins BMM100 shows with "make": "Antminer"

That's not supposed to happen...

> FutureBit Apollo II doesn't get recognized
> 

Not on the supported miner list, happy to add it though...

Could you open issues for both of these on asic-rs and we can go into more detail there?

stratum-apps/src/monitoring/sv1.rs
```diff
@@ -3,14 +3,21 @@
...
+#[cfg(feature = "asic-rs-telemetry")]
+use asic_rs_core::data::{hashrate::HashRateUnit, miner::MinerData};
+
 /// Information about a single SV1 client connection
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct Sv1ClientInfo {
     pub client_id: usize,
     pub channel_id: Option<u32>,
+    #[schema(value_type = Option<String>)]
+    pub connection_ip: Option<IpAddr>,
```
bit-aloo: We can remove this, or the one in `MinerTelemetry`. At the same time, if we keep connection ip here, it should not be an option as we always know which IP we are connected to.

stratum-apps/src/monitoring/sv1.rs
bit-aloo, line 23: We have type names defined for these fields in stratum-apps, we can use them instead of vanilla datatypes.

miner-apps/translator/Cargo.toml
bit-aloo: Maybe we can get this reimported from stratum apps? instead of defining it here?

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,195 @@
...
+                .version_rolling_mask
+                .as_ref()
+                .map(|mask| format!("{:08x}", mask.0)),
+            version_rolling_min_bit: dd
+                .version_rolling_min_bit
+                .as_ref()
+                .map(|bit| format!("{:08x}", bit.0)),
+        })
+        .ok()
+}
```
bit-aloo, line 45: This can be moved to downstream impl

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,195 @@
...
+    DataField::Chips,
+    DataField::ExpectedHashrate,
+    DataField::Fans,
+    DataField::PsuFans,
+    DataField::FluidTemperature,
+    DataField::TuningTarget,
+    DataField::LightFlashing,
+    DataField::Messages,
+    DataField::Pools,
+];
```
bit-aloo: I would move this to stratum-apps now, considering we have a feature flag now for asic-rs.

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,195 @@
...
+            return None;
+        }
+    };
+
+    Some(MinerTelemetry::from(
+        miner
+            .get_data_filtered(MINER_TELEMETRY_EXCLUDED_FIELDS.to_vec())
+            .await,
+    ))
+}
```
bit-aloo: I was thinking of something like this:

```

pub struct MinerFactorySv2(MinerFactory);

impl Deref for MinerFactorySv2 {
    type Target = MinerFactory;
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl MinerFactorySv2 {
    fn new() -> MinerFactorySv2 {
        MinerFactorySv2(MinerFactory::new())
    }

    async fn fetch_miner_telemetry(&self, ip: IpAddr) -> Option<MinerTelemetry> {
        let miner = match self.get_miner(ip).await {
            Ok(Some(miner)) => miner,
            Ok(None) => {
                debug!("No miner management interface found at {ip}");
                return None;
            }
            Err(error) => {
                debug!("Failed to get miner management interface at {ip}: {error}");
                return None;
            }
        };

        Some(MinerTelemetry::from(
            miner
                .get_data_filtered(MINER_TELEMETRY_EXCLUDED_FIELDS.to_vec())
                .await,
        ))
    }
}

```
Like we can have MinerFactorySv2 in the stratum-apps under the same asic-rs feature flag and then export this.

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -0,0 +1,195 @@
...
+        &self,
+        downstream_id: DownstreamId,
+    ) -> Option<MinerTelemetry> {
+        self.miner_telemetry
+            .get(&downstream_id)
+            .map(|telemetry| telemetry.clone())
+    }
+
+    pub(crate) async fn run_miner_telemetry_loop(
+        self: Arc<Self>,
```
bit-aloo:
```suggestion
        &self,
```

stratum-apps/Cargo.toml
```diff
@@ -62,6 +63,7 @@ fallback-coordinator = ["tokio-util"]
 config = []
 payout = ["config", "core"]
 monitoring = ["serde_json", "axum", "prometheus", "utoipa", "utoipa-swagger-ui"]
+asic-rs-telemetry = ["monitoring", "dep:asic-rs-core"]
```
bit-aloo: Maybe?
```suggestion
monitoring-with-asic-rs-tele = ["monitoring", "dep:asic-rs-core"]
```

stratum-apps/src/monitoring/sv1.rs
GitGab19: All the other monitoring schemas are using vanilla Rust types, so I'm gonna keep this as it for coherence.

miner-apps/translator/src/lib/sv1/monitoring.rs
GitGab19: I’d prefer to keep this as a private helper in the monitoring module for consistency with JDC and Pool.

bit-aloo: We can still do a `impl Downstream` here, like how we do for Sv1Server?

miner-apps/translator/src/lib/sv1/monitoring.rs
GitGab19: I would not change things TBH, since `stratum-apps` just contains the schemas, traits, and conversions.

And then every application implements the traits, and the logic to enrich those monitoring schemas.

In this case tProxy is responsible for fetching data through the `asic-rs` APIs, and use that to enrich the schema defined in `stratum-apps`.

All the applications are working in this way, since we introduced the monitoring module in `stratum-apps`.

I think keeping the separation of concerns is better.

miner-apps/translator/src/lib/sv1/monitoring.rs
GitGab19: See https://github.com/stratum-mining/sv2-apps/pull/553#discussion_r3426731080

stratum-apps/src/monitoring/sv1.rs
bit-aloo: Yup, not a blocker for this PR, but we do use a lot of types in monitoring which can be taken from stratum-apps, as we do for sv2-apps. Maybe we can open an issue for this.

miner-apps/translator/Cargo.toml
GitGab19: `stratum-apps` only needs `asic-rs-core` (not the full `asic-rs`) for the MinerData -> MinerTelemetry conversion.

bit-aloo:
@GitGab19 whats the reason why we don't have this for JDC, as sv2 native miners can connect to it directly as well, and we might want this for those connections?

miner-apps/translator/src/lib/sv1/monitoring.rs
GitGab19: I'd prefer to keep the helper to match what is being done by Pool and JDC.

stratum-apps/src/monitoring/sv1.rs
GitGab19: https://github.com/stratum-mining/sv2-apps/issues/566

miner-apps/translator/Cargo.toml
bit-aloo: hmm... Now as you mentioned I saw... these are different.

GitGab19:
> @GitGab19 whats the reason why we don't have this for JDC, as sv2 native miners can connect to it directly as well, and we might want this for those connections?

The scope of this PR is adding `asic-rs` telemetry data to tProxy, as described in https://github.com/stratum-mining/sv2-apps/issues/445

I see the value in leveraging `asic-rs` also on JDC and Pool, but I'd do it in a separate PR.

In that PR we might move some shared logic in `stratum-apps`, as you were suggesting on this PR.

plebhash (approved):

miner-apps/translator/src/lib/sv1/monitoring.rs
plebhash: whenever we implement this on JDC, wouldn't we end up with a duplicated version of `fetch_miner_telemetry`?

GitGab19: I just pushed the new commits introducing the telemetry data also on JDC, you can take a look.

GitGab19:
I added `asic-rs` telemetry data also on JDC, as discussed during today's PR Review Club call.

Most of the common logic between tProxy and JDC have been moved to `stratum-apps`.

This PR is now ready for review again.

stratum-apps/src/monitoring/sv1.rs
```diff
@@ -20,6 +25,7 @@ pub struct Sv1ClientInfo {
     pub extranonce2_len: usize,
     pub version_rolling_mask: Option<String>,
     pub version_rolling_min_bit: Option<String>,
+    pub miner_telemetry: Option<MinerTelemetry>,
```
bit-aloo, line 30: I think we can feature gate, this?

bit-aloo: This will also help us to feature gate the miner_telemetry module as a whole instead of parts of it.

GitGab19: > This will also help us to feature gate the miner_telemetry module as a whole instead of parts of it.

Can you elaborate more on this?

I kept MinerTelemetry ungated because it's a schema type.

bit-aloo: Can't we gate schema types? like clients should be able to parse payload with or without telemetry type?

GitGab19: Sure, but we would end up in a situation where the same API endpoint can return that field, or not.

More specifically, in this case the Pool's API would not return it, while JDC/tProxy would.

GitGab19: I just pushed some changes where I feature-gated the `miner_telemetry` module and the schema field as well.

bit-aloo (approved):
ACK. Looks good to me now. Thanks.

plebhash:
tACK JDC:

```
curl -s "http://127.0.0.1:9091/api/v1/clients" | jq
{
  "offset": 0,
  "limit": 25,
  "total": 2,
  "items": [
    {
      "client_id": 7,
      "extended_channels_count": 0,
      "standard_channels_count": 1,
      "total_hashrate": 1000000000.0,
      "miner_telemetry": {
        "make": "Antminer",
        "model": "BRAIINS MINI MINER BMM 100",
        "firmware_version": "2026-04-14-0-912d084c-26.04-plus",
        "reported_hashrate_hs": 911813115727.8176,
        "power_consumption_w": 36.0,
        "efficiency_j_per_th": 39.48177469597426,
        "average_temperature_c": null,
        "uptime_secs": 510,
        "is_mining": true
      }
    },
    {
      "client_id": 3,
      "extended_channels_count": 0,
      "standard_channels_count": 1,
      "total_hashrate": 735294100000.0,
      "miner_telemetry": {
        "make": "Bitaxe",
        "model": "Gamma",
        "firmware_version": "v2.14.0b3",
        "reported_hashrate_hs": 883045410200.0,
        "power_consumption_w": 13.6532593,
        "efficiency_j_per_th": 15.461559668723817,
        "average_temperature_c": 53.0,
        "uptime_secs": 581,
        "is_mining": true
      }
    }
  ]
}
```

PR #554, opened 2026-06-12
deprecate `channels_sv2::server::jobs::job_store::JobStore`

plebhash:
companion https://github.com/stratum-mining/stratum/pull/2190


bit-aloo (approved):
ACK. Looks much nicer now.

GitGab19 (approved):

PR #560, opened 2026-06-16
update pool healthcheck endpoint to accurately reflect availability

based64god:
Fixes an issue where the API healthiness endpoint would incorrectly state that the pool is healthy. A readiness check would probe /api/v1/health and get a 200 even if the attached bitcoin node is still syncing.


pool-apps/pool/src/lib/monitoring.rs
```diff
@@ -1,14 +1,27 @@
...
 use crate::{channel_manager::ChannelManager, downstream::Downstream};
 
+/// How long the pool may go without a fresh template or prev-hash from the
+/// bitcoin node / Template Provider before it is treated as unavailable.
+///
+/// During normal operation the Template Provider pushes a new template on every
+/// chain tip and whenever the mempool changes, so a healthy node updates far
+/// more frequently than this. A gap this long indicates the node is unreachable
+/// or stalled (e.g. a hung Template Provider whose TCP connection stays open).
+const NODE_TEMPLATE_STALENESS_TIMEOUT: Duration = Duration::from_secs(120);
```
based64god, line 23: Open to making this configurable, but seemed like a sensible default. Let me know.

GitGab19:
Can you please describe with more details the issue you ran into?

What were you trying to do?

It's not clear at all, after reading the PR description.

based64god:
Sure. Sorry. This fixes an issue where the API healthiness endpoint would incorrectly state that the pool is healthy. A readiness check would probe /api/v1/health and get a 200 even if the attached bitcoin node is still syncing.

My assumption here is that since the bitcoin node and pool are necessarily moving lockstep due to IPC constraints, that this state change should be reflected poolside as well.

plebhash:
I'm afraid we don't have full clarity on what we're trying to label as "stale" here:
- are we waiting for IBD/sync to finish?
- is mempool stale or empty?

the current approach doesn't distinguish between these two scenarios, and IMHO labeling the application as "unavailable" because of an empty/stale mempool is not correct

also, I'm not a fan of how the `NodeHealth`+`HealthMonitoring` abstractions are being introduced, because they heavily imply that all possible application edge cases revolve around node health, which is not necessarily true

---

but overall, I think this PR raises a good point:

currently, `/api/v1/health` naively returns `ok` regardless of application state, which can be a bit misleading UX... so sooner or later we should start thinking about:
- which abstractions should be introduced to represent edge-case application states
- how to encode edge-case application states as HTTP responses

this also has some conceptual overlap with a potential "diagnostics" API endpoint that we've discussed in the past (for which I couldn't find a tracker issue)

---

@based64god would you be willing to give us more context on what you're working on?

that would help maintainers evaluate how to prioritize the work around the issues being raised here

based64god:
> I'm afraid we don't have full clarity on what we're trying to label as "stale" here:
> 
> * are we waiting for IBD/sync to finish?
> * is mempool stale or empty?
> 
> the current approach doesn't distinguish between these two scenarios, and IMHO labeling the application as "unavailable" because of an empty/stale mempool is not correct
> 
> also, I'm not a fan of how the `NodeHealth`+`HealthMonitoring` abstractions are being introduced, because they heavily imply that all possible application edge cases revolve around node health, which is not necessarily true
> 
> but overall, I think this PR raises a good point:
> 
> currently, `/api/v1/health` naively returns `ok` regardless of application state, which can be a bit misleading UX... so sooner or later we should start thinking about:
> 
> * which abstractions should be introduced to represent edge-case application states
> * how to encode edge-case application states as HTTP responses
> 
> this also has some conceptual overlap with a potential "diagnostics" API endpoint that we've discussed in the past (for which I couldn't find a tracker issue)
> 
> @based64god would you be willing to give us more context on what you're working on?
> 
> that would help maintainers evaluate how to prioritize the work around the issues being raised here

Sure. The primary issue is one around deployment automation. If I have a large number of pools running on a cloud provider or in a large datacenter/site, then I need to have some way to communicate Pool/Node state to a load balancer so I can guarantee that miners are going to get the dance they expect. There's a small handful of ways a pool could be sufficiently "Not ready" for traffic. The main one is a syncing node, but you raise a good point about mempool staleness as well.

plebhash:
closing in favor of #577

thanks @based64god!

PR #569, opened 2026-06-17
use integers powers of 2 in Sv1 `mining.set_difficulty` messages 

### LINKED ISSUE #561: tProxy should send integer power of 2 values in the Sv1 `mining.set_difficulty` messages
One adopter reported us some issues around Sv1 `mining.set_difficulty` messages crafted by our tProxy.

The main problem is about having float values (and not integers) which are not powers of 2.

Apparently there are some firmware versions which misbehave when receiving float values.

I manually tested some of the major pools endpoints, and they all send integer values which are powers of 2, so our tProxy should probably do the same to be more sure about Sv1 compatibility in general.

GitGab19:
Closes #561 

Companion https://github.com/stratum-mining/stratum/pull/2210

### DIFF
--- a/miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
+++ b/miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
@@ -43,7 +43,10 @@ use stratum_apps::{
                 build_sv2_open_extended_mining_channel,
                 build_sv2_submit_shares_extended_from_sv1_submit,
             },
-            sv2_to_sv1::{build_sv1_notify_from_sv2, build_sv1_set_difficulty_from_sv2_target},
+            sv2_to_sv1::{
+                build_sv1_notify_from_sv2,
+                build_sv1_set_difficulty_from_sv2_target_with_integer_power_of_two_rounding,
+            },
         },
         sv1_api::{json_rpc, server_to_client, utils::HexU32Be, IsServer},
     },
@@ -53,6 +56,8 @@ use stratum_apps::{
 use tokio::net::TcpListener;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, trace, warn};
+
+const SV1_MIN_DIFFICULTY_FOR_INTEGER_POWER_OF_TWO_ROUNDING: f64 = 1.0;
 
 #[derive(Clone)]
 struct Sv1ServerIo {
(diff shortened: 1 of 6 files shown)


GitGab19:
I just saw that this change was also recently made by DMND guys on their `dmnd-client`: https://github.com/dmnd-pool/dmnd-client/commit/0eda6aee1433ad98fb60d98ec8978ccda32f06d4

miner-apps/translator/src/lib/sv1/sv1_server/sv1_difficulty.rs
```diff
@@ -0,0 +1,119 @@
...
+        return Err(TproxyErrorKind::InvalidSv1Difficulty(difficulty));
+    }
+
+    // Preserve fractional sub-1 difficulties so low-hashrate dev/test miners can still submit
+    // shares at the configured rate.
+    if difficulty < 1.0 {
+        let value = serde_json::Number::from_f64(difficulty)
+            .ok_or_else(|| TproxyErrorKind::InvalidSv1Difficulty(difficulty))?;
+        return Ok(serde_json::Value::Number(value));
+    }
```
plebhash: and this won't cause the same kind of problems?

GitGab19: Yes, but this rule applies only for diffs < 1, which means we're in a testing environment (ITF) or we're pointing cpu-miners at tProxy.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -43,7 +45,7 @@ use stratum_apps::{
                 build_sv2_open_extended_mining_channel,
                 build_sv2_submit_shares_extended_from_sv1_submit,
             },
-            sv2_to_sv1::{build_sv1_notify_from_sv2, build_sv1_set_difficulty_from_sv2_target},
```
plebhash, line 46: `build_sv1_set_difficulty_from_sv2_target` originally came from `stratum-translation` crate

and now we are providing an alternative implementation (with the same name) inside `translator_sv2`, which feels a bit redundant and confusing

wouldn't it be better to tweak `stratum-translation` so that it provides two flavors of `build_sv1_set_difficulty_from_sv2_target`?

GitGab19: I'll do that, and open a PR on `stratum` soon.

GitGab19: Done, opened https://github.com/stratum-mining/stratum/pull/2210 on `stratum` and adapted the code on this PR accordingly.

plebhash (approved):

GitGab19:
Cherry-picked the commit from https://github.com/stratum-mining/sv2-apps/pull/581

PR #573, opened 2026-06-18
bump `asic-rs` to `v0.6.2`

plebhash:
follow-up to #553

`v0.6.2` incorporates fixes to the issues I reported:
- https://github.com/256foundation/asic-rs/issues/276
- https://github.com/256foundation/asic-rs/issues/275

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -164,9 +164,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "asic-rs"
-version = "0.6.0"
+version = "0.6.2"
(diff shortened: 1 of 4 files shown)


bit-aloo (approved):

GitGab19:
Shouldn't we automatically get the `0.6.2` without having to specifying it in the toml?

We already have `0.6.0`.

plebhash:
> Shouldn't we automatically get the `0.6.2` without having to specifying it in the toml?
> 
> We already have `0.6.0`.

that would be the case only if these were strictly lib crates, but we commit lockfiles

so yeah, maybe this PR doesn't need to touch the `.toml` files, but if we don't bump the `Cargo.lock` files we're going to continue pinned to `0.6.0`

GitGab19:
Right, you're correct!

GitGab19 (approved):

PR #576, opened 2026-06-19
binary sv2 cleanup on sv2-apps

bit-aloo:
companion: https://github.com/stratum-mining/stratum/pull/2200


bit-aloo:
This is rebased now. We can remove the first commit when the stratum PR gets merged.

plebhash (approved):

PR #582, opened 2026-06-21
refactor(pool): replace monolithic start loop with state machine PoolRuntime

### LINKED ISSUE #528: need to refactor `PoolSv2::start`
as described by @Shourya742 below

nulllpc:
Refactors the monolithic `start` function inside the Pool (`pool-apps/pool`) into a structured, typestate-based state machine (`PoolRuntime`) to manage bootstrap, component lifecycles, and graceful shutdown.

- Introduce PoolRuntime state machine
- Properly use and propagate PoolError
- Remove the old start function
- Fix linting

Fixes https://github.com/stratum-mining/sv2-apps/issues/528

### DIFF
--- a/pool-apps/pool/src/lib/mod.rs
+++ b/pool-apps/pool/src/lib/mod.rs
@@ -1,35 +1,14 @@
-use std::{
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
-    thread::JoinHandle,
+use error::PoolErrorKind;
+use pool_runtime::{Init, PoolRuntime};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
 };
-
-use async_channel::unbounded;
-
-use stratum_apps::{
-    bitcoin_core_sv2::common::template_distribution_protocol::CancellationToken,
-    stratum_core::bitcoin::consensus::Encodable, task_manager::TaskManager,
-    tp_type::TemplateProviderType, utils::types::GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS,
-};
+use stratum_apps::bitcoin_core_sv2::common::template_distribution_protocol::CancellationToken;
 use tokio::sync::Notify;
-use tracing::{debug, error, info, warn};
+use tracing::info;
 
-use jd_server_sv2::job_declarator::{
-    job_validation::{bitcoin_core_ipc::BitcoinCoreIPCEngine, JobValidationEngine},
-    JobDeclarator,
-};
-
-use crate::{
-    channel_manager::ChannelManager,
-    config::PoolConfig,
-    error::PoolErrorKind,
-    template_receiver::{
-        bitcoin_core::{connect_to_bitcoin_core, BitcoinCoreSv2TDPConfig},
-        sv2_tp::Sv2Tp,
-    },
-};
+use crate::config::PoolConfig;
 
 pub mod channel_manager;
 pub mod config;
@@ -38,6 +17,7 @@ pub mod error;
 mod io_task;
 #[cfg(feature = "monitoring")]
 mod monitoring;
+mod pool_runtime;
 pub mod template_receiver;
 pub mod utils;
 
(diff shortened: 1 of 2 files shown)


pool-apps/pool/Cargo.toml
```diff
@@ -18,6 +18,7 @@ path = "src/lib/mod.rs"
 
 [dependencies]
 stratum-apps = { version = "0.6.0", path = "../../stratum-apps", features = ["pool"] }
+async-trait = { version = "0.1.89" }
```
nulllpc: I added this crate so the `Shutdown` trait from `stratum-apps` can be async

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,861 @@
...
+pub struct PoolRuntime<State> {
+    pool: PoolSv2,
+    encoded_outputs: Vec<u8>,
+    coinbase_outputs: Vec<TxOut>,
+    monitoring_server: Option<MonitoringServer>,
+    task_manager: Arc<TaskManager>,
+    state: State,
+}
+
+#[allow(clippy::result_large_err)]
```
nulllpc, line 148: Note: I ran into an issue with clippy 

```
error: the `Err`-variant returned from this function is very large
```

I think the root cause is because the `PoolError` enum can be very large, before we just return `PoolErrorKind` so we might have sidestepped this issue. I temporarily skipped the clippy complaint because this can be a separate issue

miner-apps/translator/src/lib/sv1/monitoring.rs
```diff
@@ -4,8 +4,10 @@
 use std::{collections::HashSet, net::IpAddr, time::Duration};
 
 use stratum_apps::{
-    monitoring::sv1::{Sv1ClientInfo, Sv1ClientsMonitoring},
-    monitoring::{MinerTelemetry, MinerTelemetryCollector},
+    monitoring::{
+        sv1::{Sv1ClientInfo, Sv1ClientsMonitoring},
+        MinerTelemetry, MinerTelemetryCollector,
+    },
```
bit-aloo, line 10: Need to feature flag them now, otherwise pool wont compile without monitoring

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,860 @@
...
+                    self.task_manager.clone(),
+                    cancellation_token.clone(),
+                    jds_config.supported_extensions().to_vec(),
+                    jds_config.required_extensions().to_vec(),
+                )
+                .await
+            {
+                Ok(_) => (),
+                Err(err) => {
+                    return Err((PoolError::shutdown(PoolErrorKind::Jds(err.kind)), self));
```
bit-aloo: If an error happens here, we lose the jd, as its initialized before

stratum-apps/src/utils/types.rs
```diff
@@ -37,4 +37,13 @@ impl From<(DownstreamId, ChannelId)> for VardiffKey {
...
+#[async_trait::async_trait]
+pub trait CanShutdown {
+    /// Performs a graceful shutdown. The default implementation does nothing.
+    async fn shutdown(self)
+    where
+        Self: Sized,
+    {
+        // Default implementation is empty
+    }
+}
```
bit-aloo: Kindly revert this, as these are marker traits and we dont want lifecycle semantics associated to them

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,860 @@
...
+            encoded_outputs: self.encoded_outputs,
+            monitoring_server: self.monitoring_server,
+            state: new_state,
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl CanShutdown for PoolRuntime<Init> {
+    async fn shutdown(self)
```
bit-aloo: We can have a single shutdown abstracted over type parameters and only implemented by Ready state and partial return state. That should do the job.

pool-apps/pool/Cargo.toml
nulllpc: Reverted as requested here https://github.com/stratum-mining/sv2-apps/pull/582#discussion_r3451847160

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+
+pub struct ChannelReady {
+    channel_data: ChannelData,
+}
+
+pub struct JdsReady {
+    channel_data: ChannelData,
+
+    jd: Option<JobDeclarator>,
+}
```
plebhash, line 65:
```suggestion
pub struct JdsReady {
    channel_data: ChannelData,
    jd: Option<JobDeclarator>,
}
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+
+pub struct TemplateProviderReady {
+    channel_data: ChannelData,
+
+    jd: Option<JobDeclarator>,
+
+    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,
+}
```
plebhash, line 69:
```suggestion
pub struct TemplateProviderReady {
    channel_data: ChannelData,
    jd: Option<JobDeclarator>,
    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,
}
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+
+pub struct ChannelManagerReady {
+    channel_data: ChannelData,
+
+    jd: Option<JobDeclarator>,
+
+    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,
+
+    channel_manager: ChannelManager,
+}
```
plebhash, line 74:
```suggestion
pub struct ChannelManagerReady {
    channel_data: ChannelData,
    jd: Option<JobDeclarator>,
    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,
    channel_manager: ChannelManager,
}
```

pool-apps/pool/src/lib/error.rs
```diff
@@ -38,6 +38,9 @@ pub struct TemplateProvider;
 #[derive(Debug)]
 pub struct Downstream;
 
+#[derive(Debug)]
+pub struct PoolClient;
```
plebhash: what is the rationale for the introduction of `PoolClient`?

I think it is very confusing and counter-intuitive

IMO we should either:
- A. rename it to something like `Orchestrator`
- B. remove it in favor of `PoolErrorKind`

personally I'd lean towards B, since it feels more aligned with the current error propagation system and the signatures proposed on #528 

but I'll defer to @Shourya742 for the final decision

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+impl IntoShutdownParts for Running {
+    fn into_shutdown_parts(self) -> (Option<JobDeclarator>, Option<BitcoinCoreSv2Handle>) {
+        (self.jd, self.bitcoin_core_sv2)
+    }
+}
+
+impl<State: IntoShutdownParts> PoolRuntime<State> {
+    pub async fn shutdown(self) {
+        let (jd, bitcoin_core_sv2) = self.state.into_shutdown_parts();
```
plebhash:
```suggestion
        self.pool.cancellation_token.cancel();
        
        let (jd, bitcoin_core_sv2) = self.state.into_shutdown_parts();
```

otherwise tasks that only observe `self.pool.cancellation_token` may keep running

so `self.task_manager.join_all()` could potentially time out and force abort during bootstrap-failure shutdown

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+}
+
+pub trait IntoShutdownParts {
+    fn into_shutdown_parts(self) -> (Option<JobDeclarator>, Option<BitcoinCoreSv2Handle>)
+    where
+        Self: Sized,
+    {
+        (None, None)
+    }
+}
```
plebhash: do we really need `IntoShutdownParts`?

it was not originally proposed on #528 and I feel it's introducing unnecessary cognitive load

cc @Shourya742

pool-apps/pool/src/lib/pool_runtime.rs
plebhash: the suggestion above assumes `IntoShutdownParts` would remain on this PR

but after more careful consideration, I wonder if `IntoShutdownParts` should really stay

pool-apps/pool/src/lib/pool_runtime.rs
bit-aloo: Agree, I don't think we need any such machinery.

nulllpc: The idea behind this is I was trying to consolidate the shutdown logic, all states ultimately transition into shutdown parts and the shutdown logic takes it from there. Originally, I was implementing shutdown for each state which @Shourya742 said it was an overkill so this is what I came up with. I would really appreciate if you guys could give me some guidance on structuring the shutdown

pool-apps/pool/src/lib/error.rs
nulllpc: This comes back to my original PR about error handling in JDC. TLDR: if we strip the `PoolError` to only `PoolErrorKind`, we might lose context on the error that could help us fail fast or retry, so propagating `PoolError` would give us the full context all the way through. The `PoolClient` here is just a marker for the top level logic so I don't think it's too much to add. I'm in favor of renaming it

Here's a related discussion in my old PR
https://github.com/stratum-mining/sv2-apps/pull/456#discussion_r3224976497

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+use crate::{
+    channel_manager::ChannelManager,
+    error::{PoolClient, PoolError, PoolErrorKind},
+    template_receiver::{
+        bitcoin_core::{connect_to_bitcoin_core, BitcoinCoreSv2TDPConfig},
+        sv2_tp::Sv2Tp,
+    },
+};
+
+pub struct ChannelData {
```
plebhash:
```suggestion
pub struct Io {
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+pub struct BitcoinCoreSv2Handle {
+    join_handle: JoinHandle<()>,
+    cancellation_token: CancellationToken,
+}
+
+pub struct Init;
+
+pub struct ChannelReady {
+    channel_data: ChannelData,
+}
```
plebhash:
```suggestion
pub struct IoReady {
    io: Io,
}
```

pool-apps/pool/src/lib/pool_runtime.rs
plebhash: here's how I'd do it:

```rust
...

// pure marker, no need to carry data
// nothing has been initialized yet
pub struct Init;

pub struct JdsReady {
    io: Io,
    // no jd
}

pub struct TemplateProviderReady {
    io: Io,
    // no jd
    // no bitcoin_core_sv2
}

pub struct ChannelManagerReady {
    io: Io,
    // no jd
    // no bitcoin_core_sv2
    channel_manager: ChannelManager,
}

// pure marker, no need to carry data
// everything is initialized, and jd + bitcoin_core_sv2 exist in PoolRuntime
pub struct Running;

pub struct PoolRuntime<State> {
    pool: PoolSv2,
    task_manager: Arc<TaskManager>,
    state: State,

    // Shutdown-owned handles shared across states
    jd: Option<JobDeclarator>,
    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,

    // other fields...
    encoded_outputs: Vec<u8>,
    coinbase_outputs: Vec<TxOut>,
    #[cfg(feature = "monitoring")]
    monitoring_server: Option<MonitoringServer>,
}

impl<State> PoolRuntime<State> {
    pub async fn shutdown(mut self) {
        // IMPORTANT: Always signal global cancellation first
        self.pool.cancellation_token.cancel();

        if let Some(jd) = self.jd.take() {
            jd.shutdown();
        }

        if let Some(bitcoin_core_sv2) = self.bitcoin_core_sv2.take() {
            bitcoin_core_sv2.cancellation_token.cancel();

            info!("Waiting for BitcoinCoreSv2TDP dedicated thread to shutdown...");
            match bitcoin_core_sv2.join_handle.join() {
                Ok(_) => info!("BitcoinCoreSv2TDP dedicated thread shutdown complete."),
                Err(e) => error!("BitcoinCoreSv2TDP dedicated thread error: {e:?}"),
            }
        }

        match tokio::time::timeout(
            std::time::Duration::from_secs(GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS),
            self.task_manager.join_all(),
        )
        .await
        {
            Ok(_) => {
                info!("All tasks joined cleanly");
            }
            Err(_) => {
                warn!(
                    "Tasks did not finish within {} seconds, aborting",
                    GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS
                );
                self.task_manager.abort_all().await;
                info!("Joining aborted tasks...");
                self.task_manager.join_all().await;
                warn!("Forced shutdown complete");
            }
        }

        self.pool.shutdown_notify.notify_waiters();
        self.pool.is_alive.store(false, Ordering::Relaxed);
        info!("Pool shutdown complete.");
    }
}
```

pool-apps/pool/src/lib/error.rs
plebhash: @Shourya742 ?

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+    pub async fn bootstrap(self) -> Result<PoolRuntime<Running>, PoolError<PoolClient>> {
+        let runtime = match self.init_channels() {
+            Ok(rt) => rt,
+            Err((e, rt)) => {
+                rt.shutdown().await;
+                return Err(e);
+            }
+        };
+
+        let runtime = match runtime.bootstrap_jds().await {
```
plebhash: not a blocker, but would help make the semantics of the state types more explicitly clear


```suggestion
        let runtime: PoolRuntime<JdsReady> = match runtime.bootstrap_jds().await {
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+            monitoring_server: None,
+            task_manager: Arc::new(TaskManager::new()),
+            state: Init,
+            coinbase_outputs,
+            encoded_outputs,
+        })
+    }
+
+    pub async fn bootstrap(self) -> Result<PoolRuntime<Running>, PoolError<PoolClient>> {
+        let runtime = match self.init_channels() {
```
plebhash: not a blocker, but would help make the semantics of the state types more explicitly clear

```suggestion
        let runtime: PoolRuntime<IoReady> = match self.init_channels() {
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+
+        let runtime = match runtime.bootstrap_jds().await {
+            Ok(rt) => rt,
+            Err((e, rt)) => {
+                rt.shutdown().await;
+                return Err(e);
+            }
+        };
+
+        let runtime = match runtime.bootstrap_template_provider().await {
```
plebhash: not a blocker, but would help make the semantics of the state types more explicitly clear

```suggestion
        let runtime: PoolRuntime<TemplateProviderReady> = match runtime.bootstrap_template_provider().await {
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+            Ok(rt) => rt,
+            Err((e, rt)) => {
+                rt.shutdown().await;
+                return Err(e);
+            }
+        };
+
+        let runtime = match runtime.bootstrap_channel_manager().await {
```
plebhash: not a blocker, but would help make the semantics of the state types more explicitly clear

```suggestion
        let runtime: PoolRuntime<ChannelManagerReady> = match runtime.bootstrap_channel_manager().await {
```

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,698 @@
...
+            Ok(rt) => rt,
+            Err((e, rt)) => {
+                rt.shutdown().await;
+                return Err(e);
+            }
+        };
+
+        match runtime.start_services().await {
```
plebhash: not a blocker, but would help make the semantics of the state types more explicitly clear

```suggestion
        let _runtime: PoolRuntime<Running> = match runtime.start_services().await {
```

pool-apps/pool/src/lib/pool_runtime.rs
plebhash: as noted here: https://github.com/stratum-mining/sv2-apps/issues/528#issuecomment-4848600282

pool-apps/pool/src/lib/pool_runtime.rs
plebhash: as noted here: https://github.com/stratum-mining/sv2-apps/issues/528#issuecomment-4848600282

pool-apps/pool/src/lib/pool_runtime.rs
bit-aloo: Thats pretty cool, thats what I had in my mind.

pool-apps/pool/src/lib/error.rs
bit-aloo: I would go with the suggestion by @plebhash

nulllpc: Do you mean renaming or reverting back to PoolErrorKind? And I suppose this would apply to the JDC and other components as well? @Shourya742

plebhash: on my comment above I say I'd lean towards option B, which means removing `PoolClient` in favor of `PoolErrorKind`

> And I suppose this would apply to the JDC and other components as well?

yes, although the scope of this PR should remain limited to Pool only

nulllpc: Thanks, I misread your comment

nulllpc:
All comments are addressed, please take a look when you can. Thanks again for the suggestions @plebhash @Shourya742

plebhash:
thanks @nulllpc, we're almost there

as a last iteration, I'd populate the new `pool-apps/pool/src/lib/pool_runtime.rs` with some concise Rust Docs comments, so we keep [`docs.rs/pool_sv2`](https://docs.rs/pool_sv2/latest/pool_sv2/) nicely documented with this new module

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+    state: State,
+
+    jd: Option<JobDeclarator>,
+    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,
+
+    encoded_outputs: Vec<u8>,
+    coinbase_outputs: Vec<TxOut>,
+    #[cfg(feature = "monitoring")]
+    monitoring_server: Option<MonitoringServer>,
+}
```
plebhash, line 92:
```suggestion
pub struct PoolRuntime<State> {
    pool: PoolSv2,
    task_manager: Arc<TaskManager>,
    state: State,
    jd: Option<JobDeclarator>,
    bitcoin_core_sv2: Option<BitcoinCoreSv2Handle>,
    encoded_outputs: Vec<u8>,
    coinbase_outputs: Vec<TxOut>,
    #[cfg(feature = "monitoring")]
    monitoring_server: Option<MonitoringServer>,
}
```

pool-apps/pool/src/lib/mod.rs
```diff
@@ -60,288 +40,19 @@ impl PoolSv2 {
         }
     }
 
-    /// Starts the Pool main loop.
     pub async fn start(&self) -> Result<(), PoolErrorKind> {
```
plebhash, line 56: add some Rust Docs comment here explaining how this is starting and blocking on `PoolRuntime`

nulllpc:
I think we're ready for the final review @plebhash

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,614 @@
...
+        let (tp_to_channel_manager_sender, tp_to_channel_manager_receiver) = unbounded();
+
+        let io = Io {
+            downstream_to_channel_manager_sender,
+            downstream_to_channel_manager_receiver,
+            channel_manager_to_tp_sender,
+            channel_manager_to_tp_receiver,
+            tp_to_channel_manager_sender,
+            tp_to_channel_manager_receiver,
+        };
```
plebhash: I just noticed that there's no `IoReady`

upon investigation, I realized that `bootstrap_jds` is being executed unconditionally, and `Io` gets created inside of it

that's not very coherent

we should have a `bootstrap_io` (which does get executed unconditionally), and `bootstrap_jds` should only be executed when `Some(jds_config)`

nulllpc: Yeah, I thought we could squeeze it in `bootstrap_jd` so we would have 1 fewer state. You're right, coherence is more important here. I'll fix it

plebhash (approved):

pool-apps/pool/src/lib/pool_runtime.rs
```diff
@@ -0,0 +1,638 @@
...
+            };
+
+        let runtime: PoolRuntime<ChannelManagerReady> =
+            match runtime.bootstrap_channel_manager().await {
+                Ok(rt) => rt,
+                Err((e, rt)) => {
+                    rt.shutdown().await;
+                    return Err(e);
+                }
+            };
```
bit-aloo, line 233: I would add a Failed state instead of handling each path aborts and then handle abort at start method, only once.

plebhash: @Shourya742 can you send a follow-up PR?

bit-aloo: hmm.. that would be better. I will raise a follow-up for this.

Issue #446, opened 2026-04-21
Support loading authority secret key from environment variable


average-gary:
## Problem

The authority secret key is currently stored as plaintext in the TOML config file (e.g., `pool-config.toml`). This is a security risk because:
- The private key sits on disk during operation
- Anyone with file access can read the key
- Key may persist in backups, log files, or crash dumps

## Related

- Related to https://github.com/stratum-mining/stratum/issues/1958 (key-utils leak secrets)

## Affected Components

- Pool (pool-apps/pool)
- Job Declaration Server (pool-apps/jd-server)
- Job Declaration Client (miner-apps/jd-client)

Note: Translator only requires the authority public key (no secret key needed).

## Proposal

Add support for loading the authority secret key from an environment variable instead of the TOML config file:

- `POOL_AUTHORITY_SECRET_KEY`
- `JDS_AUTHORITY_SECRET_KEY`
- `JDC_AUTHORITY_SECRET_KEY`

### Implementation approach:

1. Make `authority_secret_key` optional in the TOML deserialization struct
2. At config loading time, check for the corresponding environment variable
3. If env var is set, use it; otherwise fall back to TOML value (for backward compatibility)
4. Validate that either env var or TOML has the key (fail startup if neither)

This follows the existing pattern where `RUST_LOG` is already loaded from environment.

## Backward Compatibility

TOML config files should continue to work. The env var takes precedence if both are set.

plebhash:
I'd say this also falls under the scope of https://github.com/stratum-mining/sv2-apps/issues/280?

PR #585, opened 2026-06-23
expose all config options as environment variables

### LINKED ISSUE #280: consider leveraging `env` variables to overwrite `.toml` configs
we already introduced `shellexpand` dependency via #184 

it seems this crate also has functionality for expanding shell environment variables

this should help us get rid of the templates under https://github.com/stratum-mining/sv2-apps/tree/main/docker/config

based64god:
This is an initial stab at making all config file values exposed to a user via an environment variable. This deprecates the need to have `envsubst` install in containerized deployments, and changes the associated examples/templates.

This is a small API change for consumers already using environment variables in their deployments, but that can be easily changed if needed.

There's a few pain points with this approach, so I'm looking for some feedback. Embedding arrays, enums, in text is pretty hairy. This is how I've seen it done elsewhere, and it seems to work.

It does introduce somewhat ugly edgecases around the priority of upstreams, given that now they're effectively prioritized by how BTree sorts <NAME> slugs under the hood. Without further changes to the config API, however, I don't know that it can get better than this. 

close #280 #446


stratum-apps/src/config_helpers/loader.rs
```diff
@@ -0,0 +1,330 @@
...
+///
+/// `<NAME>` only groups the fields of one upstream together; entries are ordered
+/// alphabetically by `<NAME>`, so use names like `A`/`B` or `1`/`2` to control the
+/// fallback priority. If any such variable is present, the resulting list fully
+/// replaces the `upstreams` array from the config file.
+pub fn load_config<T: DeserializeOwned>(
+    config_path: &str,
+    env_prefix: &str,
+    list_keys: &[&str],
+) -> Result<T, ConfigError> {
```
based64god, line 37: This maybe deserves to be broken out into a macro? I didn't want to make this any more complex than absolutely necessary, but adding/removing structured config keys will need to be reflected here.

lucasbalieiro (commented):
First round of review. I like the direction this is taking. However, it still needs a deeper review of the implementation itself. From my initial read, I think there are some edge cases worth looking into, but I'll need to test them first. I'm currently finishing a review on another repository and will come back to this afterward.

One thing I noticed is that this PR states it closes issue #280, but it diverges from the solution proposed in that issue, which was to use shell expansion.

That said, I think `shellexpand` and this PR are solving different problems at different stages.

`shellexpand` is very useful for cases where we want to keep a configuration file and expand environment variables within it, for example (see also the discussion in #446):

```bash
listen_address = "${POOL_LISTEN_ADDRESS}"
cert_validity_sec = ${POOL_CERT_VALIDITY_SEC}
shares_per_minute = ${POOL_SHARES_PER_MINUTE}
supported_extensions = [${POOL_SUPPORTED_EXTENSIONS}]
```

This PR, on the other hand, is moving in the direction of loading configuration directly from environment variables and giving them priority over values defined in the TOML file. Whether we want that behavior or not probably requires broader consensus.

Another positive aspect is how it handles more complex structures such as the `upstreams` and `bitcoin-core` configurations. When I first thought about this problem a long time ago, my approach would have been to store all upstream information in a single long string separated by `;` and then parse it, which is much less elegant.

I'll test the edge cases I think I found while reading the code and post my findings later.

One thing that could be improved immediately is the error reporting. For example:

```bash
thread 'main' panicked at pool/src/args.rs:42:6:
Failed to load or deserialize config: missing field `listen_address`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

In this case, I simply started the application without providing either a configuration file or any environment variables. The current error message doesn't make it obvious what the user is expected to do next. A more explicit message could improve the onboarding and debugging experience.

Maybe something along the lines of: `No configuration found. Supply a TOML file (-c) or set POOL__* environment variables.`

Also, the commit history could be better. One commit is doing too many things, if you have the time, split it please.

And the comments over the codebase seems bloated. I think they could be trimmed a bit.

plebhash:
> loading configuration directly from environment variables and giving them priority over values defined in the TOML file. Whether we want that behavior or not probably requires broader consensus.

yeah this is perfectly aligned with what I'd envision as the ideal UX for config parameters:
- env variables take precendence
- `.toml` files are fallback/default
- **all** mandatory parameters must be filled (either via env vars or `.toml`)... if not, app should halt with an error

---

> One thing I noticed is that this PR states it closes issue #280, but it diverges from the solution proposed in that issue, which was to use shell expansion.
> 
> That said, I think `shellexpand` and this PR are solving different problems at different stages.
> 
> `shellexpand` is very useful for cases where we want to keep a configuration file and expand environment variables within it, for example (see also the discussion in #446):
> 
> ```shell
> listen_address = "${POOL_LISTEN_ADDRESS}"
> cert_validity_sec = ${POOL_CERT_VALIDITY_SEC}
> shares_per_minute = ${POOL_SHARES_PER_MINUTE}
> supported_extensions = [${POOL_SUPPORTED_EXTENSIONS}]
> ```

thanks for the analysis @lucasbalieiro, I agree that @based64god made the right call with not using `shellexpand` here

for the ideal UX described above, if we used `shellexpand` for this functionality, we would be forced to:
- have `$VAR`-style syntax on all config parameters of all `.toml` files
- **all** config parameters would have to be populated as env vars every time we launched the app

both aspects of this approach would be suboptimal

looking back in retrospective, I guess I mixed up the functionality of `shellexpand` with what [`config::Environment`](https://docs.rs/config/latest/config/struct.Environment.html) delivers

the scope of functionality of `shellexpand` is actually much narrower, as @based64god had already mentioned over Discord

so apologies for the noise I introduced here!

plebhash:
a few high level comments on what @based64god wrote on the PR description:

> There's a few pain points with this approach, so I'm looking for some feedback. Embedding arrays, enums, in text is pretty hairy. This is how I've seen it done elsewhere, and it seems to work.

> It does introduce somewhat ugly edgecases around the priority of upstreams, given that now they're effectively prioritized by how BTree sorts slugs under the hood. Without further changes to the config API, however, I don't know that it can get better than this.

as @lucasbalieiro mentioned above, I think the approach taken here is ok... it's not perfect, but it's hard to imagine alternatives that wouldn't be much uglier

IMO we just need to document it properly

for example, we should warn the user against footguns such as:
- `10` sorts before `2`
- `PRIMARY` sorts before `BACKUP`
- double-underscores `__` have special separator meaning, different from single underscores `_`

as long as the rules are clear and the potential edge-cases are documented, I think we're ok

additionally, examples go a long way... for example, we could document/nudge people towards 0-padded enumeration:
- `UPSTREAM_01`
- `UPSTREAM_02`
- ...
- `UPSTREAM_10`

---

we should make sure all documentations of the new functionality is also available on `{pool,jd-client,translator}/README.md`

currently, they're limited to Docker-related files, but we must make sure that non-Docker users are also informed

docker/docker_env.example
```diff
@@ -1,34 +1,93 @@
...
+# config files. Variables use the `<PREFIX>__<FIELD>` naming, with `__` separating
+# nested keys. The prefixes are POOL, JDC and TPROXY.
+#
+# Conventions for non-scalar fields:
+#   * Tagged enums (template_provider_type) take the variant as a path segment,
+#     e.g. POOL__TEMPLATE_PROVIDER_TYPE__BITCOINCOREIPC__NETWORK. Case-insensitive.
+#   * Upstreams are arrays, so they use `<PREFIX>__UPSTREAM_<NAME>__<FIELD>`; the
+#     <NAME> groups one upstream's fields together and orders entries
+#     alphabetically (use A/B or 1/2 to control fallback priority).
+# =============================================================================
```
plebhash, line 18: @Shourya742 if I understood your question during dev call today, this is the answer

as @lucasbalieiro mentioned on his first review comment, I think alphabetical order and double underscore separators are a good solution

the only small adjustment I would make is to also document this on `{pool, jd-client, translator}/README.md` instead of keeping it limited under Docker scope

bit-aloo: I will have a look, questions was mostly around complex toml fields and how are we presenting them via CLI

based64god:
Appreciate the feedback. Will have an updated commit history for the meeting tomorrow ^^

based64god:
@lucasbalieiro @plebhash I addressed your guys' feedback, thanks for the help. The history should be much cleaner, there should be better documentation for addressing the lexicographic footgun, and there should be helpful error messages for missing config on startup. Should be good for a final review. Thanks :)

lucasbalieiro (commented):
tested the docker changes at `7476d119db92d026ec8fb48b758ebcf107469e3c`.

Unsure if is related to this PR, but it looks like we've reintroduced the `SIGINT` signal handling issue mentioned here: https://github.com/stratum-mining/sv2-apps/pull/299#issuecomment-4056476305

Every time I run the pool with JDS enabled and stop the Docker container with `Ctrl+C`, it crashes my Node v30.2. I haven't been able to reproduce it on v31 yet, but that node isn't fully synced, so I haven't verified it properly.

I'll open an issue if the problem persists on main branch

docker/docker-compose.yml
```diff
@@ -3,16 +3,12 @@ services:
...
     env_file:
       - docker_env
     volumes:
-      - ./config/pool-jds-config.toml.template:/app/pool-jds-config.toml.template:ro
       - "${BITCOIN_SOCKET_PATH?}:/root/.bitcoin/node.sock"
-    entrypoint: >
-      sh -c '
-        envsubst < /app/pool-jds-config.toml.template> /app/pool-config.toml &&
-        exec /app/pool_sv2
-      '
```
lucasbalieiro, line 15: about https://github.com/stratum-mining/sv2-apps/pull/585#pullrequestreview-4627071457

```diff
entrypoint: >
      sh -c '
        exec /app/pool_sv2
      '
```

let's keep the entrypoint to avoid the signal handler issue.

lucasbalieiro: let's apply this to all other services

plebhash:
> Every time I run the pool with JDS enabled and stop the Docker container with `Ctrl+C`, it crashes my Node v30.2. I haven't been able to reproduce it on v31 yet, but that node isn't fully synced, so I haven't verified it properly.

this was a known issue with v30.x, which we bypassed with a safeguard in the container entrypoint

plebhash:
some CI jobs are failing because it's out of date with `main` (after merging of https://github.com/stratum-mining/sv2-apps/pull/576)

@based64god can you rebase against `main`? (not merge plz)

docker/docker-compose.yml
based64god: Would you be opposed to baking this into the image? Seems like something we might want to make default if users build their own compose or are running in kubernetes.

lucasbalieiro: > Would you be opposed to baking this into the image?

no, sounds good

based64god: should be fixed now.

GitGab19:
@lucasbalieiro can you check if we can get rid of `shellexpand` in this PR?

lucasbalieiro:
> @lucasbalieiro can you check if we can get rid of `shellexpand` in this PR?

we could remove, but not because of this PR. They are solving different things. 
see: https://github.com/stratum-mining/sv2-apps/issues/280#issuecomment-4905027992

GitGab19:
Thanks @lucasbalieiro, I completely missed that comment, sorry!

So let's wait for the final review of @Shourya742, as we said during dev call.

pool-apps/pool/src/args.rs
```diff
@@ -31,11 +31,17 @@ pub struct Args {
 pub fn process_cli_args() -> PoolConfig {
```
bit-aloo: can you make the return type Result here, a result and propagate the error

stratum-apps/src/config_helpers/loader.rs
```diff
@@ -0,0 +1,323 @@
...
+///
+/// Environment variables join `env_prefix` and the field path with `__`, e.g.
+/// `POOL__JDS__LISTEN_ADDRESS` for the field `jds.listen_address`. Tagged enum
+/// variants are just another path segment, matched case-insensitively, e.g.
+/// `POOL__TEMPLATE_PROVIDER_TYPE__BITCOINCOREIPC__NETWORK=mainnet`.
+///
+/// Fields named in `list_keys` (lowercase `snake_case`) are parsed as
+/// comma-separated lists, e.g. `POOL__SUPPORTED_EXTENSIONS=1,2,3`. A single
+/// numeric value is read as a scalar, not a 1-element list, so list at least
+/// two values.
```
bit-aloo: Its fairly possible to have a single value for list_key case?

stratum-apps/src/config_helpers/loader.rs
```diff
@@ -0,0 +1,323 @@
...
+///
+/// Upstream arrays cannot be expressed with `__` paths, so they use the
+/// dedicated form `<PREFIX>__UPSTREAM_<NAME>__<FIELD>`. `<NAME>` groups one
+/// upstream's fields together and orders entries alphabetically; if any such
+/// variable is set, the resulting list replaces the file's `upstreams` array.
+pub fn load_config<T: DeserializeOwned>(
+    config_path: &str,
```
bit-aloo:
```suggestion
    config_path: impl AsRef<Path>,
```

pool-apps/pool/src/args.rs
```diff
@@ -31,11 +31,17 @@ pub struct Args {
...
     let args = Args::parse();
     let config_path = args.config_path.to_str().expect("Invalid config path");
```
bit-aloo: We can remove this, considering `load_config` accept `AsRef<Path>`

miner-apps/jd-client/src/args.rs
```diff
@@ -31,11 +31,12 @@ pub fn process_cli_args() -> Result<JobDeclaratorClientConfig, JDCErrorKind> {
         JDCErrorKind::BadCliArgs
     })?;
```
bit-aloo: We can remove this, considering `load_config` accept `AsRef<Path>`

miner-apps/translator/src/args.rs
```diff
@@ -33,18 +33,17 @@ pub fn process_cli_args() -> Result<TranslatorConfig, TproxyErrorKind> {
     // Parse CLI arguments
     let args = Args::parse();
 
-    // Build configuration from the provided file path
     let config_path = args.config_path.to_str().ok_or_else(|| {
         error!("Invalid configuration path.");
         TproxyErrorKind::BadCliArgs
     })?;
```
bit-aloo: We can remove this, considering `load_config` accept `AsRef<Path>`

bit-aloo (commented):
Some suggestions.

stratum-apps/src/config_helpers/loader.rs
based64god: Fixed. Includes a regression test.

scripts/validate-docker-config.sh
plebhash, line 1: given that this file is being deleted, now we should update this:

https://github.com/stratum-mining/sv2-apps/blob/6a197abaa2e3a1f09ceda58bd82db839773154f7/scripts/list-config-schema-changes.sh#L13-L14

plebhash:
I think I found another edge case worth thinking about:

```rust
// stratum-apps/src/tp_type.rs:80
pub enum TemplateProviderType {
    Sv2Tp { address: String, public_key: Option<Secp256k1PublicKey> },
    BitcoinCoreIpc { version, network, data_dir, fee_threshold, min_interval },
}
```

we're aiming for (and documenting) a UX where env vars always override whatever is set on the `.toml` files

but for TP type, if we choose `Sv2Tp` in env, and `BitcoinCoreIpc` in `.toml` (or vice-versa), the app fails at startup with a deserialize error

we can either mirror the logic for `upstream` detection (thus increasing complexity to the loader), or document this as a known caveat

---

overall it's a loud startup error, not a silent misconfiguration... so it's not critical and I'm fine with either approach

based64god:
> I think I found another edge case worth thinking about:
> 
> ```rust
> // stratum-apps/src/tp_type.rs:80
> pub enum TemplateProviderType {
>     Sv2Tp { address: String, public_key: Option<Secp256k1PublicKey> },
>     BitcoinCoreIpc { version, network, data_dir, fee_threshold, min_interval },
> }
> ```
> 
> we're aiming for (and documenting) a UX where env vars always override whatever is set on the `.toml` files
> 
> but for TP type, if we choose `Sv2Tp` in env, and `BitcoinCoreIpc` in `.toml` (or vice-versa), the app fails at startup with a deserialize error
> 
> we can either mirror the logic for `upstream` detection (thus increasing complexity to the loader), or document this as a known caveat
> 
> overall it's a loud startup error, not a silent misconfiguration... so it's not critical and I'm fine with either approach

I fixed this in the most recent push by making the loader more robust. This should now prefer the "env always wins" UX, which seems ideal to me. I'm still not super happy about `load_config`'s source of truth being reliant on the caller to know what the config <-> env mapping should be. It seems preferrable to me that this would happen with reflection or something, but that's for a follow up I think.

plebhash (approved):

pool-apps/pool/src/args.rs
```diff
@@ -28,16 +28,19 @@ pub struct Args {
...
-        .add_source(File::new(config_path, FileFormat::Toml))
-        .build()
-        .and_then(|settings| settings.try_deserialize::<PoolConfig>())
-        .expect("Failed to load or deserialize config");
+
+    // Env vars prefixed `POOL__` override values from the optional TOML file.
+    let mut config: PoolConfig = load_config(
+        &args.config_path,
+        "POOL",
+        &["supported_extensions", "required_extensions"],
```
GitGab19: JDS can also theoretically deal with some extensions, so we need to cover those as well:
```suggestion
        &[
+            "supported_extensions",
+            "required_extensions",
+            "jds.supported_extensions",
+            "jds.required_extensions",
+        ],
```

It would be good to update accordingly the Pool README and the env example with these JDS extension variables.

based64god: Fixed in the most recent commit.

GitGab19: Could we add an app-level env-only config test? The generic loader tests pass even when an app forgets to register a structured field (as happened with the JDS extension lists). A representative `PoolConfig` test would catch that wiring regression.

based64god: Added the test you're looking for

GitGab19 (approved):

bit-aloo (approved):
LGTM now, I have some points but not a blocker for this PR. Thanks @based64god .

PR #589, opened 2026-07-01
relax legacy solo payout verification in tProxy

GitGab19:
This PR relaxes the payout verification (optionally) done in tProxy when using legacy solo `user_identities`, in the shape of `<payout_address>[.worker]`.

In that case, the check is about making sure jobs pay at least 90% of spendable coinbase outputs to the miner address (allowing up to 10% service fee in legacy solo mode).

This is needed to unblock https://github.com/stratum-mining/sv2-ui/pull/203 and to handle other similar cases which might pop up in the future.


GitGab19:
Tested against PyBlock Pool endpoint (https://pyblock.xyz:8443/sv2.php) and it's working fine now.

plebhash (approved):

stratum-apps/src/payout.rs
```diff
@@ -118,6 +129,20 @@ impl PayoutMode {
             .filter(|output| output.script_pubkey.as_bytes() == script_pubkey.as_bytes())
             .map(|output| output.value.to_sat())
             .sum();
+        if matches!(self, Self::LegacySolo { .. }) {
+            if actual_miner_sats == 0 {
```
plebhash: I'm not going to block the PR over this, but I feel I must raise a concern here:

by relying on `actual_miner_sats == 0`, we're making `verify_payout` flag give an arguably false sense of security to the user

a malicious solo pool could bypass the `verify_payout` flag by allocating 1 single sat for the user, while taking the entire payout for itself

with that said, I would not propose that we add yet another `expected_solo_fee` config parameter only for this purpose, since it would make UX more confusing (especially when compared to `donation` mode), which is probably why @GitGab19 avoided this approach in the first place

---

maybe a middle ground could come from hardcoding some `MAX_LEGACY_SOLO_FEE = 10`, based on the assumption that no reasonable legacy solo pool would charge more than 10% fee

that avoids extra UX complexity, while still making it hard for a malicious solo pool to cheat their users

worst case scenario, they would advertise e.g. 1% while maliciously taking 10%, which is arguably cheating this system as well, just not as bad as only giving 1 sat to the user

so end of the day, I guess it's hard to find a perfect solution, and this would only be a partial mitigation on the vector described above

---

again, not really a blocker... just flagging because it's a real attack vector for malicious solo pools

I'm also fine with proceeding with the PR as-is and only addressing this vector in case we get concrete evidence that it is happening in the wild

PyBLOCK-Bitcoin: > so end of the day, I guess it's hard to find a perfect solution

We have a Python3-based Coinbase decoder derived from Mining.Notify that works perfectly.

If you'd like, we can send you the raw script; based on how it works, you could incorporate it or build a similar checker.

This way, you can constantly check both fields: the fee the pool takes and the total amount your address receives.

@GaltRanch.

GitGab19: @plebhash I think the middle ground solution you proposed is better, so I pushed some changes with that suggestion, thank you!

Now we have an hardocded `MIN_LEGACY_SOLO_PAYOUT_PERCENTAGE = 90` to respect in case of legacy solo path.

plebhash (approved):

PR #592, opened 2026-07-02
JDS isolates `DownstreamState`

### LINKED ISSUE #590: JDS's `BitcoinCoreIPCEngine` is sharing state across different downstream clients
## problem

`BitcoinCoreIPCEngine` stores `declared_custom_jobs` as `DashMap<RequestId, DeclaredCustomJob>`, where `request_id` is client-controlled and only unique per connection.

Because the map is shared across downstream clients, two clients can use the same `request_id` and clobber each other’s declared job state.

This can cause wrong job retrieval/removal during `SetCustomMiningJob` validation and lead to cross-client griefing/DoS behavior.

## solution

introduce this abstraction to `bitcoin_core_ipc.rs`:

```rust
struct DownstreamState {
 declared_custom_jobs: HashMap<RequestId, DeclaredCustomJob>,
 allocated_token_entries: HashMap<JdToken, AllocatedTokenEntry>
}
```
and change `BitcoinCoreIPCEngine`:
```diff
pub struct BitcoinCoreIPCEngine {
 request_sender: async_channel::Sender<JdRequest>,
- allocated_token_entries: Arc<DashMap<JdToken, AllocatedTokenEntry>>,
- declared_custom_jobs: Arc<DashMap<RequestId, DeclaredCustomJob>>,
+ downstream_state: SharedMap<DownstreamId, DownstreamState>>,
 cancellation_token: CancellationToken,
 jdp_thread_handle: Arc<Mutex<Option<JoinHandle<()>>>>,
}
```

as a side-effect, we'll also need to change `JobValidationEngine::{handle_declare_mining_job,handle_set_custom_mining_job}` method signatures, because they will need a `DownstreamId` parameter

plebhash:
close #590

### DIFF
--- a/pool-apps/jd-server/README.md
+++ b/pool-apps/jd-server/README.md
@@ -40,4 +40,4 @@ Active tokens are single-use. During `SetCustomMiningJob` handling, the `JobDecl
 
 Malformed, unknown, expired, wrong-owner, or already consumed tokens are rejected with `invalid-mining-job-token`.
 
-Allocated and active tokens are periodically removed by a janitor task after their configured timeouts. When a downstream disconnects, its allocated tokens are removed immediately because they only authorize future `DeclareMiningJob` messages from that JDS connection. Active tokens are retained until they are consumed or expire, because they authorize a later `SetCustomMiningJob` sent through the Pool's Mining Protocol path, not through the JDS connection.
+Allocated and active tokens are periodically removed by a janitor task after their configured timeouts. When a downstream disconnects, both allocated and active tokens owned by that downstream are removed immediately.
--- a/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
+++ b/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
@@ -55,7 +55,7 @@ use stratum_apps::{
         },
     },
     tp_type::BitcoinNetwork,
-    utils::types::{JdToken, RequestId},
+    utils::types::{DownstreamId, JdToken, RequestId},
 };
 
 /// Snapshot of a previously declared mining job, stored after a `DeclareMiningJob` is
@@ -432,6 +432,7 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
     ///    `DeclaredCustomJob` for later `SetCustomMiningJob` validation.
     async fn handle_declare_mining_job(
         &self,
+        _downstream_id: DownstreamId,
         declare_mining_job: DeclareMiningJob<'_>,
         provide_missing_transactions_success: Option<ProvideMissingTransactionsSuccess<'_>>,
     ) -> DeclareMiningJobResult {
@@ -636,7 +637,11 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
         }
     }
 
-    async fn handle_push_solution(&self, push_solution: PushSolution<'_>) {
+    async fn handle_push_solution(
+        &self,
+        downstream_id: DownstreamId,
+        push_solution: PushSolution<'_>,
+    ) {
         // Convert to static lifetime for channel transfer
         let push_solution_static = push_solution.into_static();
 
@@ -646,9 +651,9 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
         };
 
         if let Err(e) = self.request_sender.send(request).await {
-            tracing::error!("Failed to send PushSolution request: {}", e);
+            tracing::error!(downstream_id, "Failed to send PushSolution request: {}", e);
         } else {
-            tracing::debug!("PushSolution request sent successfully");
+            tracing::debug!(downstream_id, "PushSolution request sent successfully");
         }
     }
 
@@ -663,6 +668,7 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
     // DeclareMiningJob token.
     async fn handle_set_custom_mining_job(
         &self,
+        _downstream_id: DownstreamId,
         set_custom_mining_job: SetCustomMiningJob<'_>,
         allocated_token: JdToken, // Note: This is the corresponding DeclareMiningJob token
     ) -> SetCustomMiningJobResult {
--- a/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
+++ b/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
@@ -432,30 +432,14 @@ impl BitcoinCoreIPCEngine {
                     _ = tokio::time::sleep(janitor_interval) => {
                         let now = Instant::now();
                         janitor_downstream_states.for_each_mut(|downstream_id, state| {
-                            let expired_tokens: Vec<JdToken> = state
-                                .allocated_token_entries
-                                .iter()
-                                .filter_map(|(token, entry)| {
-                                    if now.saturating_duration_since(entry.inserted_at)
-                                        > token_timeout
-                                    {
-                                        Some(*token)
-                                    } else {
-                                        None
-                                    }
-                                })
-                                .collect();
-
-                            for token in expired_tokens {
-                                if let Some(entry) = state.allocated_token_entries.remove(&token) {
-                                    state.declared_custom_jobs.remove(&entry.request_id);
-                                    tracing::debug!(
-                                        downstream_id,
-                                        token,
-                                        request_id = entry.request_id,
-                                        "Removed expired declared custom job state"
-                                    );
-                                }
+                            let expired = state.prune_expired_allocations(now, token_timeout);
+                            for (token, request_id) in expired {
+                                tracing::debug!(
+                                    downstream_id,
+                                    token,
+                                    request_id,
+                                    "Removed expired declared custom job state"
+                                );
                             }
                         });
                     }
@@ -493,6 +477,10 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
                 }
             }
         }
+    }
+
+    fn cleanup_downstream(&self, downstream_id: DownstreamId) {
+        self.downstream_states.remove(&downstream_id);
     }
 
     /// Validates a `DeclareMiningJob` by forwarding it to Bitcoin Core over IPC.
@@ -768,34 +756,10 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
         set_custom_mining_job: SetCustomMiningJob<'_>,
         allocated_token: JdToken, // Note: This is the corresponding DeclareMiningJob token
     ) -> SetCustomMiningJobResult {
-        // Look up request_id using the allocated token
-        let request_id = match self
-            .downstream_states
-            .with(&downstream_id, |state| {
-                state
-                    .allocated_token_entries
-                    .get(&allocated_token)
-                    .map(|entry| entry.request_id)
-            })
-            .flatten()
-        {
-            Some(request_id) => request_id,
-            None => {
-                tracing::debug!(
-                    downstream_id,
-                    allocated_token,
-                    "Provided token is not associated with any DeclareMiningJob request"
-                );
-                return SetCustomMiningJobResult::Error(
-                    ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN,
-                );
-            }
-        };
-
         let declared_custom_job = match self
             .downstream_states
             .with_mut(&downstream_id, |state| {
-                state.take_declared_custom_job(request_id, allocated_token)
+                state.take_declared_custom_job(allocated_token)
             })
             .flatten()
         {
@@ -804,8 +768,7 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
                 tracing::debug!(
                     downstream_id,
                     allocated_token,
-                    request_id,
-                    "DeclaredCustomJob associated with allocated token and request id not found"
+                    "Provided token is not associated with any DeclareMiningJob request"
                 );
                 return SetCustomMiningJobResult::Error(
                     ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN,
--- a/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
+++ b/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
@@ -7,8 +7,8 @@ use crate::{
         ALLOCATED_TOKEN_TIMEOUT_SECS, JANITOR_INTERVAL_SECS,
     },
 };
-use dashmap::DashMap;
 use std::{
+    collections::HashMap,
     path::PathBuf,
     sync::{Arc, Mutex},
     thread::JoinHandle,
@@ -54,6 +54,7 @@ use stratum_apps::{
             ERROR_CODE_SET_CUSTOM_MINING_JOB_STALE_CHAIN_TIP,
         },
     },
+    sync::SharedMap,
     tp_type::BitcoinNetwork,
     utils::types::{DownstreamId, JdToken, RequestId},
 };
@@ -75,6 +76,73 @@ struct DeclaredCustomJob {
 struct AllocatedTokenEntry {
     request_id: RequestId,
     inserted_at: Instant,
+}
+
+/// Per-downstream client state for declared custom jobs and their token entries.
+#[derive(Default)]
+struct DownstreamState {
+    declared_custom_jobs: HashMap<RequestId, DeclaredCustomJob>,
+    allocated_token_entries: HashMap<JdToken, AllocatedTokenEntry>,
+}
+
+impl DownstreamState {
+    /// Stores a DeclaredCustomJob and its corresponding allocated token.
+    fn insert_declared_custom_job(
+        &mut self,
+        request_id: RequestId,
+        allocated_token: JdToken,
+        declared_custom_job: DeclaredCustomJob,
+    ) {
+        self.declared_custom_jobs
+            .insert(request_id, declared_custom_job);
+        self.allocated_token_entries.insert(
+            allocated_token,
+            AllocatedTokenEntry {
+                request_id,
+                inserted_at: Instant::now(),
+            },
+        );
+    }
+
+    /// Removes both the DeclaredCustomJob and its corresponding allocated token.
+    fn remove_declared_custom_job(&mut self, request_id: RequestId, allocated_token: JdToken) {
+        self.declared_custom_jobs.remove(&request_id);
+        self.allocated_token_entries.remove(&allocated_token);
+    }
+
+    /// Atomically removes and returns a declared custom job by its token.
+    fn take_declared_custom_job(&mut self, allocated_token: JdToken) -> Option<DeclaredCustomJob> {
+        let entry = self.allocated_token_entries.remove(&allocated_token)?;
+        let job = self.declared_custom_jobs.remove(&entry.request_id)?;
+        Some(job)
+    }
+
+    /// Removes expired allocated tokens and their corresponding DeclaredCustomJob.
+    /// Returns a list of expired `(token, request_id)` pairs for logging.
+    fn prune_expired_allocations(
+        &mut self,
+        now: Instant,
+        token_timeout: Duration,
+    ) -> Vec<(JdToken, RequestId)> {
+        let mut expired = Vec::new();
+        self.allocated_token_entries.retain(|token, entry| {
+            let keep = now.saturating_duration_since(entry.inserted_at) <= token_timeout;
+            if !keep {
+                self.declared_custom_jobs.remove(&entry.request_id);
+                expired.push((*token, entry.request_id));
+            }
+            keep
+        });
+
+        expired
+    }
+
+    /// Looks up the ValidationContext for a given RequestId.
+    fn validation_context_for_request(&self, request_id: RequestId) -> Option<ValidationContext> {
+        self.declared_custom_jobs
+            .get(&request_id)
+            .map(|job| job.validation_context)
+    }
 }
 
 #[cfg_attr(not(test), hotpath::measure_all)]
@@ -214,8 +282,7 @@ impl DeclaredCustomJob {
 #[derive(Clone)]
 pub struct BitcoinCoreIPCEngine {
     request_sender: async_channel::Sender<JdRequest>,
-    allocated_token_entries: Arc<DashMap<JdToken, AllocatedTokenEntry>>,
-    declared_custom_jobs: Arc<DashMap<RequestId, DeclaredCustomJob>>,
+    downstream_states: SharedMap<DownstreamId, DownstreamState>,
     cancellation_token: CancellationToken,
     jdp_thread_handle: Arc<Mutex<Option<JoinHandle<()>>>>,
 }
@@ -350,14 +417,11 @@ impl BitcoinCoreIPCEngine {
             }
         }
 
-        let allocated_token_to_request_id =
-            Arc::new(DashMap::<JdToken, AllocatedTokenEntry>::new());
-        let declared_custom_jobs = Arc::new(DashMap::<RequestId, DeclaredCustomJob>::new());
+        let downstream_states = SharedMap::<DownstreamId, DownstreamState>::new();
 
         // Spawn janitor task to clean up stale declared jobs that were never
         // consumed by SetCustomMiningJob.
-        let janitor_allocated_token_to_request_id = Arc::clone(&allocated_token_to_request_id);
-        let janitor_declared_custom_jobs = Arc::clone(&declared_custom_jobs);
+        let janitor_downstream_states = downstream_states.clone();
         let janitor_cancellation = cancellation_token.clone();
         tokio::spawn(async move {
             let janitor_interval = Duration::from_secs(JANITOR_INTERVAL_SECS);
@@ -367,23 +431,33 @@ impl BitcoinCoreIPCEngine {
                     _ = janitor_cancellation.cancelled() => break,
                     _ = tokio::time::sleep(janitor_interval) => {
                         let now = Instant::now();
-                        let expired_tokens: Vec<JdToken> = janitor_allocated_token_to_request_id
-                            .iter()
-                            .filter_map(|entry| {
-                                let inserted_at = entry.value().inserted_at;
-                                if now.saturating_duration_since(inserted_at) > token_timeout {
-                                    Some(*entry.key())
-                                } else {
-                                    None
-                                }
-                            })
-                            .collect();
+                        janitor_downstream_states.for_each_mut(|downstream_id, state| {
+                            let expired_tokens: Vec<JdToken> = state
+                                .allocated_token_entries
+                                .iter()
+                                .filter_map(|(token, entry)| {
+                                    if now.saturating_duration_since(entry.inserted_at)
+                                        > token_timeout
+                                    {
+                                        Some(*token)
+                                    } else {
+                                        None
+                                    }
+                                })
+                                .collect();
 
-                        for token in expired_tokens {
-                            if let Some((_, entry)) = janitor_allocated_token_to_request_id.remove(&token) {
-                                janitor_declared_custom_jobs.remove(&entry.request_id);
+                            for token in expired_tokens {
+                                if let Some(entry) = state.allocated_token_entries.remove(&token) {
+                                    state.declared_custom_jobs.remove(&entry.request_id);
+                                    tracing::debug!(
+                                        downstream_id,
+                                        token,
+                                        request_id = entry.request_id,
+                                        "Removed expired declared custom job state"
+                                    );
+                                }
                             }
-                        }
+                        });
                     }
                 }
             }
@@ -391,8 +465,7 @@ impl BitcoinCoreIPCEngine {
 
         Ok(Self {
             request_sender,
-            allocated_token_entries: allocated_token_to_request_id,
-            declared_custom_jobs,
+            downstream_states,
             cancellation_token,
             jdp_thread_handle: Arc::new(Mutex::new(Some(jdp_thread_handle))),
         })
@@ -432,7 +505,7 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
     ///    `DeclaredCustomJob` for later `SetCustomMiningJob` validation.
     async fn handle_declare_mining_job(
         &self,
-        _downstream_id: DownstreamId,
+        downstream_id: DownstreamId,
         declare_mining_job: DeclareMiningJob<'_>,
         provide_missing_transactions_success: Option<ProvideMissingTransactionsSuccess<'_>>,
     ) -> DeclareMiningJobResult {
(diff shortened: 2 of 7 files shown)


plebhash:
ready for review

pool-apps/jd-server/README.md
```diff
@@ -40,4 +40,4 @@ Active tokens are single-use. During `SetCustomMiningJob` handling, the `JobDecl
 
 Malformed, unknown, expired, wrong-owner, or already consumed tokens are rejected with `invalid-mining-job-token`.
 
-Allocated and active tokens are periodically removed by a janitor task after their configured timeouts. When a downstream disconnects, its allocated tokens are removed immediately because they only authorize future `DeclareMiningJob` messages from that JDS connection. Active tokens are retained until they are consumed or expire, because they authorize a later `SetCustomMiningJob` sent through the Pool's Mining Protocol path, not through the JDS connection.
+Allocated and active tokens are periodically removed by a janitor task after their configured timeouts. When a downstream disconnects, both allocated and active tokens owned by that downstream are removed immediately.
```
GitGab19, line 43: Is there a specific reason for changing this behavior?

plebhash: this issue predated this PR

previously, when a JDS downstream client disconnected, `TokenManager::remove_downstream` was incomplete because it only cleaned up `self.allocated_tokens` but not `self.active_tokens`... `self.active_tokens` eventually got cleaned up by the janitor task (or in the unlikely case where the client is still connected to pool and sends `SetCustomMiningJob`, consuming the token), but that's not ideal

all this context is self-contained in the commit where this discussion is happening: 6b87c93a6ce439cbc0d2919d89176cb2b0e48341

GitGab19: I see, so the idea is to cleanup also the eventual activated tokens, right?

Because from the previous description on this README, it seems the previous idea was to keep active tokens alive until their usage/expiration?

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -564,24 +586,31 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
...
-                    .insert(declare_mining_job.request_id, declared_custom_job);
-                self.allocated_token_entries.insert(
-                    allocated_token,
-                    AllocatedTokenEntry {
-                        request_id: declare_mining_job.request_id,
-                        inserted_at: Instant::now(),
-                    },
-                );
+                self.downstream_state
+                    .with_mut_or_default(downstream_id, |state| {
```
GitGab19: Why do we have `with_mut_or_default` here?

This is the race condition which could happens because of that:
```
T1: handle_declare_mining_job starts for downstream_id = 7
T1: parses token, builds request, sends validation to Bitcoin Core IPC
T1: awaits response

T2: downstream 7 disconnects
T2: cleanup_downstream(7)
T2: job_validator.cleanup_downstream(7)
T2: downstream_state.remove(&7)
T2: token_manager.remove_downstream(7)

T1: Bitcoin Core IPC returns Success
T1: with_mut_or_default(7, ...)
T1: key 7 is missing, so it creates a fresh DownstreamState
T1: inserts declared_custom_job and allocated_token_entries
```

pool-apps/jd-server/README.md
plebhash: > I see, so the idea is to cleanup also the eventual activated tokens, right?

correct

> Because from the previous description on this README, it seems the previous idea was to keep active tokens alive until their usage/expiration?

yeah, we changed the approach from a more conservative to a more pragmatical stance on this

the core point here is that JDC has two connections:
- one to Pool
- one to JDS

if JDC disconnects from JDS, but not from Pool, what do we do?

previously, we left active tokens to be either consumed or expire (since connection with Pool remains active, so a `SetCustomMiningJob` could still arrive)

now we remove them immediately

tbh I think both behaviors are correct (mainly because I don't see why a JDC would remain connected to Pool but not JDS aside from transient states), while the previous behavior is more conservative

perhaps the angle of a standalone JDS (which remains theoretically impossible on a spec level) could make the case for the more conservative behavior stronger

anyways, I'm not hardly opinionated on this... I can revert to the previous behavior too if you prefer that we remain with the more conservative approach

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: > Why do we have with_mut_or_default here?

because if this is the first `DeclareMiningJob` a client has ever sent on this connection, the new `DownstreamState` needs to be populated into `SharedMap`

but the race condition you pointed out is real

btw this also happens on the `JdResponse::MissingTransactions` case

rebasing commit history with this fix:
```diff
                     .flatten()
             });

+        // Ensure downstream state exists before awaiting IPC response.
+        // Later writes must use `with_mut` only, so a disconnect cleanup that removes this
+        // state cannot be undone by recreating it after the response arrives.
+        self.downstream_state
+            .with_mut_or_default(downstream_id, |_| {});
+
         // Create oneshot channel for response
         let (response_tx, response_rx) = tokio::sync::oneshot::channel();

@@ -586,19 +592,24 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
                     txid_list: Some(txid_list),
                     validated: true,
                 };
-                self.downstream_state
-                    .with_mut_or_default(downstream_id, |state| {
-                        state
-                            .declared_custom_jobs
-                            .insert(declare_mining_job.request_id, declared_custom_job);
-                        state.allocated_token_entries.insert(
-                            allocated_token,
-                            AllocatedTokenEntry {
-                                request_id: declare_mining_job.request_id,
-                                inserted_at: Instant::now(),
-                            },
-                        );
-                    });
+                let updated = self.downstream_state.with_mut(&downstream_id, |state| {
+                    state
+                        .declared_custom_jobs
+                        .insert(declare_mining_job.request_id, declared_custom_job);
+                    state.allocated_token_entries.insert(
+                        allocated_token,
+                        AllocatedTokenEntry {
+                            request_id: declare_mining_job.request_id,
+                            inserted_at: Instant::now(),
+                        },
+                    );
+                });
+                if updated.is_none() {
+                    tracing::error!(downstream_id, "downstream state missing after IPC response");
+                    return DeclareMiningJobResult::Error(
+                        ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR,
+                    );
+                }
                 DeclareMiningJobResult::Success
             }
             JdResponse::Error {
@@ -652,19 +663,27 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
                         txid_list: None,
                         validated: false, // this is only set to true on JdResponse::Success
                     };
-                    self.downstream_state
-                        .with_mut_or_default(downstream_id, |state| {
-                            state
-                                .declared_custom_jobs
-                                .insert(declare_mining_job.request_id, declared_custom_job);
-                            state.allocated_token_entries.insert(
-                                allocated_token,
-                                AllocatedTokenEntry {
-                                    request_id: declare_mining_job.request_id,
-                                    inserted_at: Instant::now(),
-                                },
-                            );
-                        });
+                    let updated = self.downstream_state.with_mut(&downstream_id, |state| {
+                        state
+                            .declared_custom_jobs
+                            .insert(declare_mining_job.request_id, declared_custom_job);
+                        state.allocated_token_entries.insert(
+                            allocated_token,
+                            AllocatedTokenEntry {
+                                request_id: declare_mining_job.request_id,
+                                inserted_at: Instant::now(),
+                            },
+                        );
+                    });
+                    if updated.is_none() {
+                        tracing::error!(
+                            downstream_id,
+                            "downstream state missing after IPC response"
+                        );
+                        return DeclareMiningJobResult::Error(
+                            ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR,
+                        );
+                    }

                     DeclareMiningJobResult::MissingTransactions(missing_wtxids)
                 }
```

now if this is the first `DeclareMiningJob` of a client, we populate it before starting the IPC calls

the next mutations (on `Success` or `MissingTransactions`) paths call `with_mut` and return `DeclareMiningJobResult::Error` in case it's missing

one could point out that it doesn't make sense to try to send a `DeclareMiningJob.Error` for a client that doesn't exist, which is true... but we must abide to the `JobValidationEngine` API, and trying to send a message to a non-existent client eventually becomes a `JDSError::disconnect`, which is safe/idempotent if the client was already disconnected

plebhash: also renaming `BitcoinCoreIPCEngine::downstream_state` to `BitcoinCoreIPCEngine::downstream_states` (plural) for consistency, since this is a collection (`SharedMap<DownstreamId, DownstreamState>`), not a single entity

pool-apps/jd-server/README.md
plebhash: oh wait no

now `DeclaredCustomJob` is removed from JDS memory when JDC disconnects, so even if JDC remains connected to Pool and sends a corresponding `SetCustomMiningJob`, it's impossible to verify it

so forget everything I said above, the past "conservative" behavior is literally impossible now

this should have been my answer to your original question

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -350,14 +356,11 @@ impl BitcoinCoreIPCEngine {
             }
         }
 
-        let allocated_token_to_request_id =
-            Arc::new(DashMap::<JdToken, AllocatedTokenEntry>::new());
-        let declared_custom_jobs = Arc::new(DashMap::<RequestId, DeclaredCustomJob>::new());
+        let downstream_states = SharedMap::<DownstreamId, DownstreamState>::new();
```
bit-aloo, line 420:
```suggestion
        let downstream_states = SharedMap::new();
```

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -564,24 +592,36 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
...
                 validation_context,
             } => {
-                self.declared_custom_jobs
-                    .remove(&declare_mining_job.request_id);
-                self.allocated_token_entries.remove(&allocated_token);
+                self.downstream_states.with_mut(&downstream_id, |state| {
+                    state
+                        .declared_custom_jobs
+                        .remove(&declare_mining_job.request_id);
+                    state.allocated_token_entries.remove(&allocated_token);
```
bit-aloo: lets add remove method for downstream_state?

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -564,24 +592,36 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
...
+                        .declared_custom_jobs
+                        .insert(declare_mining_job.request_id, declared_custom_job);
+                    state.allocated_token_entries.insert(
+                        allocated_token,
+                        AllocatedTokenEntry {
+                            request_id: declare_mining_job.request_id,
+                            inserted_at: Instant::now(),
+                        },
+                    );
+                });
```
bit-aloo: Lets add insert method for downstream_state?

pool-apps/jd-server/src/lib/job_declarator/job_validation/mod.rs
```diff
@@ -30,21 +30,32 @@ pub trait JobValidationEngine: Send + Sync {
...
         &self,
+        downstream_id: DownstreamId,
         set_custom_mining_job: SetCustomMiningJob<'_>,
         allocated_token: JdToken,
     ) -> SetCustomMiningJobResult;
 
+    /// Removes validation state associated with a downstream connection.
+    ///
+    /// Called by [`crate::job_declarator::JobDeclarator`] when a downstream disconnects.
+    fn cleanup_downstream(&self, _downstream_id: DownstreamId) {}
```
bit-aloo:
```suggestion
    fn cleanup_downstream(&self, downstream_id: DownstreamId);
```

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -513,11 +530,22 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
...
+                    })
+                    .flatten()
             });
 
+        // Ensure downstream state exists before awaiting IPC response.
+        // Later writes must use `with_mut` only, so a disconnect cleanup that removes this
+        // state cannot be undone by recreating it after the response arrives.
+        self.downstream_states
+            .with_mut_or_default(downstream_id, |_| {});
+
```
bit-aloo, line 590: Can you explain this a bit more? I am not understanding how this would prevent against stale downstream object creation?

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: doesn't compile without it

```
error[E0282]: type annotations needed for `SharedMap<_, _>`
   --> jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs:359:13
    |
359 |           let downstream_states = SharedMap::new();
    |               ^^^^^^^^^^^^^^^^^
...
374 |                               let expired_tokens: Vec<JdToken> = state
    |  ________________________________________________________________-
375 | |                                 .allocated_token_entries
    | |________________________________________________________- type must be known at this point
    |
help: consider giving `downstream_states` an explicit type, where the type for type parameter `V` is specified
    |
359 |         let downstream_states: SharedMap<K, V> = SharedMap::new();
    |                              +++++++++++++++++

For more information about this error, try `rustc --explain E0282`.
```

pool-apps/jd-server/src/lib/job_declarator/job_validation/mod.rs
plebhash: done in aff630de0  made cleanup_downstream(&self, downstream_id: DownstreamId) required (no default body) so future JobValidationEngine implementations must handle disconnect cleanup explicitly.

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: this change comes from this review comment: https://github.com/stratum-mining/sv2-apps/pull/592#discussion_r3560620888, which should provide extra context

> how this would prevent against stale downstream object creation?

this specific line where you're asking this question is not the whole fix by itself; it works together with the post-response changes.

the stale-state race happened because `with_mut_or_default` was used after awaiting IPC (`JdResponse::Success`/`JdResponse::MissingTransactions`), which could recreate state after disconnect cleanup.

now we:
1) create `DownstreamState` once before awaiting IPC (in case it doesn't exist yet), and
2) use `with_mut` only after IPC response

so if disconnect cleanup runs during the await and removes the entry, post-response code does not recreate it.

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
GitGab19: My clanker found another possible way for which a new token gets created even if the downstream disconnects during the `handle_declare_mining_job` execution:
```
T1: handle_declare_mining_job checks token is allocated
T1: sends validation to Bitcoin Core IPC
T1: awaits

T2: downstream disconnects
T2: cleanup_downstream(downstream_id)
T2: job_validator.cleanup_downstream(downstream_id)
T2: token_manager.remove_downstream(downstream_id)
    allocated token is removed
    active tokens are removed

T1: IPC returns Success
T1: token_manager.activate(token, downstream_id)
T1: allocated token is no longer present, but activate still creates a new active token

That contradicts the new lifecycle. Cleanup said “this downstream is gone,” but then the old in-flight DeclareMiningJob mints a fresh active token afterward.
```

pool-apps/jd-server/src/lib/job_declarator/job_validation/mod.rs
plebhash: good catch

fixed, thanks

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: done

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: done

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -663,35 +749,52 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
...
+            None => {
+                tracing::debug!(
+                    downstream_id,
+                    allocated_token,
+                    request_id,
+                    "DeclaredCustomJob associated with allocated token and request id not found"
+                );
+                return SetCustomMiningJobResult::Error(
+                    ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN,
+                );
```
GitGab19: Another point found by my clanker on this:
```
Small suggestion on the `SetCustomMiningJob` consume path: right now we first read `allocated_token -> request_id` with `with`, then later reacquire the state with `with_mut` to remove the token/job.

That leaves a small TOCTOU window where janitor cleanup, downstream cleanup, or duplicate token use can interleave between lookup and consume. It likely fails closed today, but we can make the operation atomic by deriving `request_id` inside `take_declared_custom_job` while holding the mutable state guard:

fn take_declared_custom_job(&mut self, allocated_token: JdToken) -> Option<DeclaredCustomJob> {
    let entry = self.allocated_token_entries.remove(&allocated_token)?;
    self.declared_custom_jobs.remove(&entry.request_id)
}

Then `handle_set_custom_mining_job` only needs one `with_mut`, and token/job consumption becomes a single guarded operation.
```

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: great finding, thanks for reporting!

rebasing commit history with this fix:
```diff
diff --git a/pool-apps/jd-server/src/lib/job_declarator/job_declaration_message_handler.rs b/pool-apps/jd-server/src/lib/job_declarator/job_declaration_message_handler.rs
index 2d35b1963..f09f565fd 100644
--- a/pool-apps/jd-server/src/lib/job_declarator/job_declaration_message_handler.rs
+++ b/pool-apps/jd-server/src/lib/job_declarator/job_declaration_message_handler.rs
@@ -173,17 +173,33 @@ impl HandleJobDeclarationMessagesFromClientAsync for JobDeclarator {
         {
             // if job is valid, activate token and return DeclareMiningJobSuccess
             DeclareMiningJobResult::Success => {
-                let activated_token = self.token_manager.activate(token, client_id);
-
-                let declare_mining_job_success = DeclareMiningJobSuccess {
-                    request_id: msg.request_id,
-                    new_mining_job_token: activated_token
-                        .to_le_bytes()
-                        .to_vec()
-                        .try_into()
-                        .expect("must always be valid B0_255"),
-                };
-                JobDeclaration::DeclareMiningJobSuccess(declare_mining_job_success)
+                match self.token_manager.activate(token, client_id) {
+                    Some(activated_token) => {
+                        let declare_mining_job_success = DeclareMiningJobSuccess {
+                            request_id: msg.request_id,
+                            new_mining_job_token: activated_token
+                                .to_le_bytes()
+                                .to_vec()
+                                .try_into()
+                                .expect("must always be valid B0_255"),
+                        };
+                        JobDeclaration::DeclareMiningJobSuccess(declare_mining_job_success)
+                    }
+                    None => {
+                        let declare_mining_job_error = DeclareMiningJobError {
+                            request_id: msg.request_id,
+                            error_code: ERROR_CODE_DECLARE_MINING_JOB_INVALID_MINING_JOB_TOKEN
+                                .as_bytes()
+                                .to_vec()
+                                .try_into()
+                                .expect("error code string must be valid B0_255"),
+                            error_details: Vec::new()
+                                .try_into()
+                                .expect("empty array must be valid B0_64K"),
+                        };
+                        JobDeclaration::DeclareMiningJobError(declare_mining_job_error)
+                    }
+                }
             }
             // if job is invalid, return DeclareMiningJobError
             DeclareMiningJobResult::Error(error) => {
@@ -308,19 +324,36 @@ impl HandleJobDeclarationMessagesFromClientAsync for JobDeclarator {
         {
             // if job is valid, activate token and return DeclareMiningJobSuccess
             DeclareMiningJobResult::Success => {
-                let activated_token = self
+                match self
                     .token_manager
-                    .activate(pending_declare_mining_job_token, client_id);
-
-                let declare_mining_job_success = DeclareMiningJobSuccess {
-                    request_id: msg.request_id,
-                    new_mining_job_token: activated_token
-                        .to_le_bytes()
-                        .to_vec()
-                        .try_into()
-                        .expect("must always be valid B0_255"),
-                };
-                JobDeclaration::DeclareMiningJobSuccess(declare_mining_job_success)
+                    .activate(pending_declare_mining_job_token, client_id)
+                {
+                    Some(activated_token) => {
+                        let declare_mining_job_success = DeclareMiningJobSuccess {
+                            request_id: msg.request_id,
+                            new_mining_job_token: activated_token
+                                .to_le_bytes()
+                                .to_vec()
+                                .try_into()
+                                .expect("must always be valid B0_255"),
+                        };
+                        JobDeclaration::DeclareMiningJobSuccess(declare_mining_job_success)
+                    }
+                    None => {
+                        let declare_mining_job_error = DeclareMiningJobError {
+                            request_id: msg.request_id,
+                            error_code: ERROR_CODE_DECLARE_MINING_JOB_INVALID_MINING_JOB_TOKEN
+                                .as_bytes()
+                                .to_vec()
+                                .try_into()
+                                .expect("error code string must be valid B0_255"),
+                            error_details: Vec::new()
+                                .try_into()
+                                .expect("empty array must be valid B0_64K"),
+                        };
+                        JobDeclaration::DeclareMiningJobError(declare_mining_job_error)
+                    }
+                }
             }
             // if job is invalid, return DeclareMiningJobError
             DeclareMiningJobResult::Error(error_code) => {
diff --git a/pool-apps/jd-server/src/lib/job_declarator/token_management/mod.rs b/pool-apps/jd-server/src/lib/job_declarator/token_management/mod.rs
index 756a18089..7ac69ea2f 100644
--- a/pool-apps/jd-server/src/lib/job_declarator/token_management/mod.rs
+++ b/pool-apps/jd-server/src/lib/job_declarator/token_management/mod.rs
@@ -96,8 +96,20 @@ impl TokenManager {

     /// Takes an allocated token and removes it from the internal set.
     /// Creates a corresponding active token and adds it to the internal set.
-    pub fn activate(&self, allocated_token: JdToken, downstream_id: DownstreamId) -> JdToken {
-        let removed_allocated = self.allocated_tokens.remove(&allocated_token).is_some();
+    ///
+    /// Returns `Some` with the new active token if the allocated token was present
+    /// and owned by `downstream_id`. Returns `None` if the allocated token has
+    /// already been removed (e.g. by disconnect cleanup).
+    pub fn activate(
+        &self,
+        allocated_token: JdToken,
+        downstream_id: DownstreamId,
+    ) -> Option<JdToken> {
+        let (_, (_, owner)) = self.allocated_tokens.remove(&allocated_token)?;
+
+        if owner != downstream_id {
+            return None;
+        }

         let activated_token = self.token_factory.fetch_add(1, Ordering::Relaxed);
         self.active_tokens.insert(
@@ -110,13 +122,12 @@ impl TokenManager {
             allocated_token,
             activated_token,
             downstream_id,
-            allocated_token_was_present = removed_allocated,
             allocated_tokens_len = self.allocated_tokens.len(),
             active_tokens_len = self.active_tokens.len(),
             "TokenManager: activated token"
         );

-        activated_token
+        Some(activated_token)
     }

     /// Removes an active token from the internal set.
```

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: good catch, thanks!

rebasing commit history with:
```diff
--- a/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
+++ b/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
@@ -107,13 +107,10 @@ impl DownstreamState {
         self.allocated_token_entries.remove(&allocated_token);
     }

-    fn take_declared_custom_job(
-        &mut self,
-        request_id: RequestId,
-        allocated_token: JdToken,
-    ) -> Option<DeclaredCustomJob> {
-        self.allocated_token_entries.remove(&allocated_token);
-        self.declared_custom_jobs.remove(&request_id)
+    fn take_declared_custom_job(&mut self, allocated_token: JdToken) -> Option<DeclaredCustomJob> {
+        let entry = self.allocated_token_entries.remove(&allocated_token)?;
+        let job = self.declared_custom_jobs.remove(&entry.request_id)?;
+        Some(job)
     }
 }

@@ -747,34 +744,10 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
         set_custom_mining_job: SetCustomMiningJob<'_>,
         allocated_token: JdToken, // Note: This is the corresponding DeclareMiningJob token
     ) -> SetCustomMiningJobResult {
-        // Look up request_id using the allocated token
-        let request_id = match self
-            .downstream_states
-            .with(&downstream_id, |state| {
-                state
-                    .allocated_token_entries
-                    .get(&allocated_token)
-                    .map(|entry| entry.request_id)
-            })
-            .flatten()
-        {
-            Some(request_id) => request_id,
-            None => {
-                tracing::debug!(
-                    downstream_id,
-                    allocated_token,
-                    "Provided token is not associated with any DeclareMiningJob request"
-                );
-                return SetCustomMiningJobResult::Error(
-                    ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN,
-                );
-            }
-        };
-
         let declared_custom_job = match self
             .downstream_states
             .with_mut(&downstream_id, |state| {
-                state.take_declared_custom_job(request_id, allocated_token)
+                state.take_declared_custom_job(allocated_token)
             })
             .flatten()
         {
@@ -783,8 +756,7 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
                 tracing::debug!(
                     downstream_id,
                     allocated_token,
-                    request_id,
-                    "DeclaredCustomJob associated with allocated token and request id not found"
+                    "Provided token is not associated with any DeclareMiningJob request"
                 );
                 return SetCustomMiningJobResult::Error(
                     ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN,
```

GitGab19 (approved):

pool-apps/jd-server/src/lib/job_declarator/token_management/mod.rs
```diff
@@ -96,8 +96,20 @@ impl TokenManager {
...
+    pub fn activate(
+        &self,
+        allocated_token: JdToken,
+        downstream_id: DownstreamId,
+    ) -> Option<JdToken> {
+        let (_, (_, owner)) = self.allocated_tokens.remove(&allocated_token)?;
+
+        if owner != downstream_id {
+            return None;
+        }
```
bit-aloo: this removes the allocated token before checking ownership. If called with the wrong downstream_id, it returns None but has already burned the rightful downstream’s token.

integration-tests/tests/pool_solo_mining.rs
```diff
@@ -94,7 +94,7 @@ fn assert_payout_percentage(payout_info: &PayoutInfo, expected_percentages: &[(S
             .addresses
             .iter()
             .position(|a| a == addr)
-            .expect(&format!("Address {} not found in coinbase", addr));
+            .unwrap_or_else(|| panic!("Address {} not found in coinbase", addr));
```
bit-aloo, line 97: is this because of clippy?

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -367,32 +400,41 @@ impl BitcoinCoreIPCEngine {
...
+
+                            for token in expired_tokens {
+                                if let Some(entry) = state.allocated_token_entries.remove(&token) {
+                                    state.declared_custom_jobs.remove(&entry.request_id);
+                                    tracing::debug!(
+                                        downstream_id,
+                                        token,
+                                        request_id = entry.request_id,
+                                        "Removed expired declared custom job state"
+                                    );
```
bit-aloo: Maybe we can move these cleanup to Downstream_state? wdyt?

We can club the loops and use retain instead

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
bit-aloo: I read through the discussion above. The second scenario that GitGab pointed out earlier was what I had in mind. It's clear now. Thanks!

integration-tests/tests/pool_solo_mining.rs
plebhash: hmm shit, there's actually a bunch of linting changes on this PR that are totally unrelated to what the PR really should be doing

the root cause is that I recently added the following to my `AGENTS.md`:

> on every commit you're asked to implement (or rebase), always run `cargo +nightly fmt --all; cargo +stable clippy --fix --allow-dirty --allow-staged` for every crate that's modified by this commit, to make sure it abides by the repository's linting rules

my goal was to pre-emptively avoid breaking `clippy`+`fmt` CI

but now it's adding linting changes to a commit that should only be adding `integration-tests/tests/jds_downstream_state_isolation.rs` (and that would not necessarily break CI if they weren't added to this PR)

I'm cleaning up the PR because we shouldn't have nonsensical unrelated diffs polluting the PR like this

thanks for spotting this!

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: yeah good to improve encapsulation of `DownstreamState`

thanks, done

pool-apps/jd-server/src/lib/job_declarator/token_management/mod.rs
plebhash: great catch, thanks!

should be fixed now

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -77,6 +78,76 @@ struct AllocatedTokenEntry {
...
+        self.allocated_token_entries.retain(|token, entry| {
+            let keep = now.saturating_duration_since(entry.inserted_at) <= token_timeout;
+            if !keep {
+                expired.push((*token, entry.request_id));
+            }
+            keep
+        });
+
+        for (_, request_id) in &expired {
+            self.declared_custom_jobs.remove(request_id);
```
bit-aloo: We should move this inside the if !keep statement above

bit-aloo (commented):
One nit

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
plebhash: done

bit-aloo (approved):
ACK. Thanks

Issue #574, opened 2026-06-18
need shared modules on `bitcoin_core_sv2` to avoid code duplication


plebhash:
#548 is introducing support for multiple Bitcoin Core version on `bitcoin_core_sv2`

as noted on https://github.com/stratum-mining/sv2-apps/pull/548#discussion_r3435297594 that is leading to some code duplication

over time, as more Bitcoin Core releases come out, we can expect for sure:
- many modules to remain identical across releases (resulting in even more code duplication)
- some modules to undergo changes across releases

so one potential alternative is to introduce shared modules:
```
unix_capnp
├── mod.rs
├── v30x_v31x                              # provides shared code for v30x + v31x
│   ├── mod.rs                             # exports shared modules
│   ├── job_declaration_protocol
│   │   ├── mod.rs                         # exports shared JDP modules
│   │   └── mempool.rs                     # shared implementation
│   └── template_distribution_protocol
│       ├── mod.rs                         # exports shared TDP modules
│       └── monitors.rs                    # shared implementation
├── v30x
│   ├── mod.rs
│   ├── job_declaration_protocol
│   │   ├── mod.rs
│   │   ├── error.rs
│   │   ├── handlers.rs
│   │   ├── mempool.rs                     # uses shared code from v30x_v31x/job_declaration_protocol/mempool.rs
│   │   └── monitors.rs
│   └── template_distribution_protocol
│       ├── mod.rs
│       ├── error.rs
│       ├── handlers.rs
│       ├── monitors.rs                    # uses shared code from v30x_v31x/template_distribution_protocol/monitors.rs
│       └── template_data.rs
└── v31x
    ├── mod.rs
    ├── job_declaration_protocol
    │   ├── mod.rs
    │   ├── error.rs
    │   ├── handlers.rs
    │   ├── mempool.rs                     # uses shared code from v30x_v31x/job_declaration_protocol/mempool.rs
    │   └── monitors.rs
    └── template_distribution_protocol
        ├── mod.rs
        ├── error.rs
        ├── handlers.rs
        ├── monitors.rs                    # uses shared code from v30x_v31x/template_distribution_protocol/monitors.rs
        └── template_data.rs
```

the shared `v30x_v31x` module above is specific for IPC connections, so it is placed under `unix_capnp`, while leaving `common` for "runtime API" abstractions

`common` could be renamed into `runtime_api` or some other name to make this idea more explicitly clear and avoid confusion

---

[...]

PR #593, opened 2026-07-02
WIP implement `handle_push_solution` on `jd_server_sv2` + `bitcoin_core_sv2`

### LINKED ISSUE #441: implement `handle_push_solution` on `jd_server_sv2` + `bitcoin_core_sv2`
blocked by https://github.com/bitcoin/bitcoin/issues/34626

plebhash:
NOT READY FOR REVIEW

close #441 

blocked by:
- [ ] Bitcoin Core v32 release
- [ ] `bitcoin-capnp-types` release compatible with v32

builds on top of:
- #617

---

a few conceptual points worthy of note for reviewers:

~~## `JdRequest::PushSolution` must carry `Block`~~

~~before this PR, `JdRequest::PushSolution` carried a `PushSolution` message~~

~~however, Bitcoin Core's new `submitBlock` capnproto method needs a full block~~

~~`BitcoinCoreSv2JDP` does not have enough context to reconstruct a full block from a `PushSolution`, so it's JDS's responsability to do that~~

~~therefore, `JdRequest::PushSolution` now carries a `bitcoin::Block` instead of a `PushSolution` message~~

~~## `JdResponse::Success` must carry `txdata`~~

~~before this PR, `DeclaredCustomJob` carried a `txid_list`, which was used for `merkle_path` validation of `SetCustomMiningJob`~~

~~given that now it's JDS responsability to reconstruct the full block from a `PushSolution`, `txid_list` is no longer sufficient~~

~~so `JdResponse::Success` now carries `txdata` instead of `txid_list`~~

~~`txid_list` is computed from `txdata` for `merkle_path` validation of `SetCustomMiningJob`~~

## `PushSolution` must be handled within isolated Downstream Client context

https://github.com/stratum-mining/sv2-spec/pull/189 established that upon receiving `PushSolution`, JDS MUST try to reconstruct the block based on the last acknowledged `DeclareMiningJob`, and MAY try to reconstruct based on previous ones

we're doing a KISS approach here, where we only try to propagate a solution for the last acknowledged `DeclareMiningJob`

this motivated #590 / #592 as a pre-requisite, because otherwise there would be no way to unambiguously associate a `PushSolution` with a `DeclareMiningJob`

## Bitcoin Core's new `submitSolution` capnp method

Bitcoin Core's new `submitSolution` capnp method is being used at `bitcoin_core_sv2::unix_capnp::v32x::job_declaration_protocol::BitcoinCoreSv2JDP::handle_push_solution`

when compared to the previous implementations, `BitcoinCoreSv2JDP` got a new dedicated `submit_block_thread_ipc_client` so that we don't block solution submission with regular mempool monitoring activity


plebhash:
currently, there's no official Bitcoin Core v32 release

so this PR is temporarily stubbing ITF to use a binary that's only available locally on my machine

once v32 is released I'll fix this PR and CI should be green

all integration tests pass locally though, so nothing to worry about

Sjors:
> we're doing a KISS approach here, where we only try to propagate a solution for the last acknowledged DeclareMiningJob

That seems fine as a first version, but given that new templates may be proposed quite frequently based on fee updates, and that ASIC's may chew on older jobs for a bit, it's a good TODO to try previous templates. Otherwise you completely rely on the JDC/TP to handle the block broadcast for those.

(update: I guess it won't fail if only fees changed, so the only concern is stale blocks)

> so this PR is temporarily stubbing ITF to use a binary that's only available locally on my machine

In the `bitcoin-capnp-types` we build Bitcoin Core from source in CI, which is pretty quick if you enable ccache and use a release tag. For features that are merged, but not yet released, we use the master branch.

https://github.com/2140-dev/bitcoin-capnp-types/blob/master/.github/workflows/ci.yml#L17-L33

And for open pull requests the actual PR branch, e.g. https://github.com/2140-dev/bitcoin-capnp-types/pull/28

> before this PR, DeclaredCustomJob carried a txid_list

I assume `wtxid_list`? (or at least, it's using witness ids)

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,427 @@
...
+
+        let declared_bip34_height = coinbase_tx
+            .input
+            .first()
+            .and_then(|input| {
+                decode_bip34_height_from_coinbase_script_sig(input.script_sig.as_bytes())
+            })
+            // Some templates/coinbase formats do not expose BIP34 height in canonical
+            // scriptSig push form (e.g. opcode-encoded small integers in tests/regtest).
+            // Fall back to coinbase lock_time to avoid panics and keep a stable
```
Sjors: This seems backwards. BIP34 has been active for a decade, the `nLockTime` field is part of BIP54. Bitcoin Core sets the latter, but until such a soft fork is activated, you shouldn't count on it.

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,427 @@
...
+            // Now verify that all wtxids from the declared job are available
+            let missing_wtxids = mempool_mirror.verify(&wtxid_list);
+            if !missing_wtxids.is_empty() {
+                // deliberately ignore potential errors
+                // we don't care if the receiver dropped the channel
+                let _ = response_tx.send(JdResponse::MissingTransactions {
+                    missing_wtxids,
+                    validation_context: initial_validation_context,
+                });
+                return;
```
Sjors: So the idea is to try the whole process again if anything is missing?

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,427 @@
...
+                    let _ = response_tx.send(JdResponse::Error {
+                        error_code: ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR,
+                        validation_context: initial_validation_context,
+                    });
+                    warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                    self.cancellation_token.cancel();
+                    return;
+                }
+            };
+            options.set_check_merkle_root(false);
```
Sjors: Maybe don't skip this in a debug build?

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,427 @@
...
+
+            debug!("checkBlock returned: {}", result);
+            if !result {
+                error!(
+                    reason = ?check_block_reason,
+                    debug = ?check_block_debug,
+                    "Bitcoin Core rejected the block via checkBlock"
+                );
+                debug!(
+                    "Block details - version: {:?}, prev_blockhash: {:?}, bits: {:?}, num_txs: {}",
```
Sjors: In trace logging you may want to log all wtixds?

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,427 @@
...
+            result
+        };
+
+        if !valid_job {
+            // On checkBlock failure, force-refresh template + mirror before classifying the error.
+            // The template monitor updates mempool_mirror asynchronously, so we need to avoid races
+            // where validation can run on context A while chain tip has already moved to context B.
+            // Refreshing here narrows this TOCTOU window and lets us correctly emit
+            // `stale-chain-tip` instead of generic `invalid-job` when context drift occurred.
+            if let Err(e) = self.force_update_mempool_mirror().await {
```
Sjors: This calls `createNewBlock()` under the hood, but afaik all you need is `getTip()`.

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
```diff
@@ -0,0 +1,427 @@
...
+            JdResponse::Success {
+                prev_hash: initial_validation_context.prev_hash,
+                nbits: initial_validation_context.nbits,
+                min_ntime: initial_validation_context.min_ntime,
+                txdata: txdata_for_response,
+            }
+        } else {
+            let stale_at_arrival_by_bip34 = declared_bip34_height != latest_bip34_height;
+            let context_drifted = initial_validation_context.prev_hash
+                != latest_validation_context.prev_hash
```
Sjors: A `prev_hash` mismatch means the tip moved. It could have been a reorg, but it's fine to just assume the template is stale. The BIP22 term is `inconclusive-not-best-prevblk`.

There's no point in comparing `nbits` and `latest_bip34_height`.

Comparing `min_ntime` will give you false positives, because `min_ntime` increases every second.

In other words, as I suggested in my comment above, calling `getTip()` should be enough because you only need to look at `prev_hash` drift.

plebhash:
> > before this PR, DeclaredCustomJob carried a txid_list
> 
> I assume `wtxid_list`? (or at least, it's using witness ids)

that's used for `merkle_path` validation, so it's `txid_list`

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
plebhash: line 98 is sending `JdResponse::MissingTransactions` to JDS, which triggers it to send `ProvideMissingTransactions` to JDC

when JDS receives `ProvideMissingTransactions.Success` from JDC, it sends a new `JdRequest::DeclareMiningJob` (with populated `missing_txs`), which triggers another execution of `BitcoinCoreSv2JDP::handle_declare_mining_job`

but this time the conditional on line 95 won't get triggered, because `mempool_mirror` should have 100% of the declared txs now

---

please note that this entire file is only showing up on the PR diff because of the crate module hierarchy

but there's nothing new here, this is all identical to `v31x` and `v30x` modules

the only part of this file that's relevant for this PR is `handle_push_solution`

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
plebhash: can you elaborate?

what kinds of edge cases do you expect to catch?

---

please note that `DeclareMiningJob` doesn't carry `merkle_root`, it carries `wtxid_list`... which is what we want to validate against the current UTXO set

if you look at line 134 above, you'll see that `Header` is deliberately set with a null merkle root

I mean, I guess in theory yes, we could:
- fetch `txdata` from `wtxid_list` against `mempool_mirror`
- compute `txid_list`
- compute `merkle_root` with a dummy `extranonce`
- use this `merkle_root` on line 134
- flip line 188 to `true`

but the only additional sanity check we'd be doing (outsourced to Bitcoin Core) would be against this extra computation we just added?

---

moreover, please note that this part of the codebase is a direct port of the `v30x` and `v31x` modules, so it's orthogonal to this PR

so in case we get more clarity on whether this is something to be addressed, it should be done on a follow-up issue and not on this PR

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
plebhash: thanks for pointing out

this predates and is orthogonal to this PR though: tracking via https://github.com/stratum-mining/sv2-apps/issues/595

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
plebhash: thanks for the suggestion

this predates and is orthogonal to this PR though: tracking via https://github.com/stratum-mining/sv2-apps/issues/596

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
plebhash: thanks for the suggestion

this predates and is orthogonal to this PR though: tracking via https://github.com/stratum-mining/sv2-apps/issues/597

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
plebhash: thanks for the suggestion

this predates and is orthogonal to this PR though: tracking via https://github.com/stratum-mining/sv2-apps/issues/597

bitcoin-core-sv2/src/unix_capnp/v32x/job_declaration_protocol/handlers.rs
Sjors: > what kinds of edge cases do you expect to catch?

Any bug in how SRI constructs a merkle tree, maybe as part of a fuzzer?

> is a direct port of the v30x and v31x modules

Ok that's quite confusing.

Sjors: Or did you mean to say "builds on:" instead of "blocked by:" in the PR description?

plebhash: > > is a direct port of the v30x and v31x modules
> 
> Ok that's quite confusing.

yeah it's not perfect and it's leading to a lot of code duplication (plus review friction, as we're learning now)

it's currently tracked via https://github.com/stratum-mining/sv2-apps/issues/574

---

> Or did you mean to say "builds on:" instead of "blocked by:" in the PR description?

yes #593 (this PR) builds on top of #592, but #592 does not touch the file where this discussion thread is happening

but I'm taking note to myself to break this effort down into different PRs:
- [#592](https://github.com/stratum-mining/sv2-apps/issues/592) can remain as-is
- one PR closing [#574](https://github.com/stratum-mining/sv2-apps/issues/574)
  - addressing the issue of code duplication on `bitcoin_core_sv2` (and out-of-scope reviews, which is partially the rabbit-hole we find ourselves in right now)
- one PR adding `v32x` module
  - reusing `v30x`/`v31x` modules where functionality remains unchanged
  - no `handle_push_solution` implementation
- one PR adding `handle_push_solution` (which was the original goal of [#593](https://github.com/stratum-mining/sv2-apps/issues/593) here, and for which [#592](https://github.com/stratum-mining/sv2-apps/issues/592) is a pre-requisite)

---

> > what kinds of edge cases do you expect to catch?
> 
> Any bug in how SRI constructs a merkle tree, maybe as part of a fuzzer?

overall I can empathize where you're coming from (especially the overwhelming cognitive overhead, which is definitely my fault)... on a surface level this might seem like low-hanging fruit

but I remain unconvinced

the SRI APIs that are somewhat related to this are:
- `jd_server_sv2::job_declarator::job_validation::bitcoin_core_sv2::DeclaredCustomJob::get_merkle_path` which:
  - would result in circular dependency if called here (`bitcoin_core_sv2` vs `jd_server_sv2`)
- `channels_sv2::merkle_root::merkle_root_from_path` which:
  - isn't really meant to be used here (channel logic should be orthogonal to `bitcoin_core_sv2`)
  - still needs a `path` argument

maybe one could argue these APIs could be moved to a different place where they could be imported here, but I don't think this perceived "opportunity" for differential fuzzying justifies this kind of refactoring efforts

if we eventually decide that these APIs deserve differential fuzzying against Bitcoin Core, IMO that should be done somewhere else, with a properly planned scope

moreover, these APIs are almost entirely built on top of `rust-bitcoin`, which AFAIK is already fuzzed against Core via [`bitcoinfuzz`](https://github.com/bitcoinfuzz/bitcoinfuzz), so the gain would be very little

@Sjors please let me know if you still feel strongly about `options.set_check_merkle_root(true)` and I'll create a follow-up issue... as explained above, the body of `handle_declare_mining_job` (where this discussion is taking place) pre-dates this PR, so it's not something we should address here

Sjors: I don't feel strongly. It think it makes sense in general to run blocks generated with SRI code through the `checkBlock()` IPC, but it doesn't have to be inside the JDC code, it could be somewhere else entirely.

bitcoin-core-sv2/src/runtime_api/job_declaration_protocol/io.rs
```diff
@@ -33,10 +30,11 @@ pub enum JdRequest {
...
     },
-    /// Submit a mining solution to Bitcoin Core (fire-and-forget).
-    PushSolution {
-        push_solution: PushSolution<'static>,
-    },
+    /// Submit a fully assembled block to Bitcoin Core (fire-and-forget).
+    ///
+    /// This request is currently stubbed as a warning-only no-op by the v30.x and v31.x
+    /// backends.
+    PushSolution { block: Block },
```
plebhash, line 37: note to self:

we cannot do `JdRequest::PushSolution` take a `Block`

https://github.com/stratum-mining/sv2-apps/issues/609#issuecomment-5006165234

plebhash: another reason to pivot away from this approach:

I'm noticing a lot of JDS log lines like this on my signet e2e tests:
```
2026-07-19T18:07:28.604132Z  WARN bitcoin_core_sv2::unix_capnp::v32x::job_declaration_protocol::handlers: Bitcoin Core rejected block via submitBlock reason=O
k("high-hash") debug=Ok("proof of work failed")
```

it doesn't happen 100% of the time, because propagation on JDC side is always successful, and I also see JDS log lines like this:
```
2026-07-19T18:12:07.353926Z  WARN bitcoin_core_sv2::unix_capnp::v32x::job_declaration_protocol::handlers: Bitcoin Core rejected block via submitBlock reason=O
k("duplicate") debug=Ok("")
```

I haven't been able to trace the root cause of the `submitBlock` failures with 100% certainty, but my suspicion is that the solution is being applied to the wrong custom job, due to `DeclareMiningJob` vs `PushSolution` races, which is a direct consequence of the KISS approach of only processing `PushSolution` against the "latest ACKd `DeclareMiningJob`"

plebhash:
superseded by #881, rebuilt on top of #856

Issue #598, opened 2026-07-07
bitcoin_core_sv2::BitcoinCoreSv2JDP should leverage TxCollection


Sjors:
Akin to #268, but the prerequisite https://github.com/bitcoin/bitcoin/pull/35671 is expected to land much later than https://github.com/bitcoin/bitcoin/pull/34020.

The `TxCollection` offers two opportunities:

1. Get rid of `mempool.rs` and `monitors.rs` (or keep them only for backward compatibility)
2. Hold on to the reconstructed block template and use the (slightly) more efficient `submitSolution()` instead of `submitBlock()`

Demo:
- #599

plebhash:
I'm particularly curious about why this should be preferred over #268, or as a follow-up

---

up until now, my understanding was: https://github.com/bitcoin/bitcoin/pull/34020 gets merged, then we tackle #268

the design for #268 would be something like this

(note: `bitcoin_core_sv2::BitcoinCoreSv2JDP` is an internal component of JDS, but it's being treated as a distinct entity for full architectural clarity)

### without `ProvideMissingTransactions` trigger

- `DeclareMiningJob` arrives to JDS
- JDS sends a `JdRequest::DeclareMiningJob` to `BitcoinCoreSv2JDP` with:
  - populated `wtxid_list: Vec<Wtxid>` (from `DeclareMiningJob`)
- `BitcoinCoreSv2JDP` sends a `getTransactionsByWTxID()` to Bitcoin Core
- Bitcoin Core has all txs on its mempool, so it returns an array with all `txdata` for the `DeclareMiningJob`
- `BitcoinCoreSv2JDP` builds a `Block` (because now it has the full `txdata` of the `DeclareMiningJob`)
- `BitcoinCoreSv2JDP` calls `checkBlock()`
- `BitcoinCoreSv2JDP` sends `JdResponse::Success` (which includes `txdata`)
- JDC sends `PushSolution` to JDS
- JDS builds `Block` and sends to `BitcoinCoreSv2JDP`
- `BitcoinCoreSv2JDP` sends `submitBlock()` to Bitcoin Core

### with `ProvideMissingTransactions` trigger

- `DeclareMiningJob` arrives to JDS
- JDS sends a `JdRequest::DeclareMiningJob` to `BitcoinCoreSv2JDP` with:
  - populated `wtxid_list: Vec<Wtxid>` (from `DeclareMiningJob`)
[...]

Issue #597, opened 2026-07-03
`bitcoin_core_sv2`: JDP stale-chain-tip detection should use `prev_hash` only (avoid `min_ntime`/`nbits`/`bip34` drift)


plebhash:
feedback from @Sjors :

`DeclareMiningJob` stale classification currently treats any validation-context drift as stale (`prev_hash`, `nbits`, `min_ntime`, `bip34`):

https://github.com/stratum-mining/sv2-apps/blob/453e853f96735d6f824649d98b2f95fda6938f2f/bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/handlers.rs#L289-L321

This can misclassify errors:
- `min_ntime` can increase without a tip change
- `nbits`/`bip34` comparisons are unnecessary for stale-tip detection in this path

Stale tip classification should be restricted to mismatching `prev_hash` only.

---

Finally, there's no need for `force_update_mempool_mirror` to call `createNewBlock()`.

Calling `getTip()` should be sufficient.

---

Note: this should be done on all versions, not only `v31x` (as could be implied by the code snippet above).

plebhash:
note: there's an important nuance that wasn't fully captured on the way this issue was originally framed

we compare `prev_hash` to detect chain tip drift on `ProvideMissingTransactions` round-trips (initial `prev_hash` vs final `prev_hash`)

but there's also the possibility that a `DeclareMiningJob` is stale at arrival, which is not detected via `prev_hash` (especially since this field doesn't exist in the message)

this is detected when `checkBlock` fails with `bad-cb-height` error message, for which case we also return a `DeclareMiningJob` with `error_code = stale-chain-tip`

Sjors:
I still think prevhash should be included in more sv2 messages in order to get rid of this ambiguity. But that's a separate discussion.

GitGab19:
Related to this issue:
- https://github.com/project-loupe/audit-sv2-apps/issues/157
- https://github.com/project-loupe/audit-sv2-apps/issues/161

PR #599, opened 2026-07-07
PoC:  replace the mempool mirror with Bitcoin Core's `TxCollection`

Sjors:
100% vibe coded proof of concept for #598.

Instead of keeping a mempool mirror in sync by polling `waitNext`/`getBlock` every second (the successor to the original `getrawmempool` RPC dump), JDS backend Bitcoin Core's node validates declared jobs directly against its mempool using the new `TxCollection` IPC interface. 

Additionally, he JDS submits solutions against the retained (node-side) template (`submitSolution()` instead of `submitBlock()`). The mempool mirror, its monitor loop, and all full-block transfers are gone: a block never crosses the IPC boundary in either direction. Also, the declared coinbase is fully consensus-checked at declaration time instead of when a solution is submitted.

The v30.x/v31.x backends are unchanged and keep the mirror.

### Depends on
- sv2-apps:
  - #592 — JDS downstream state isolation
  - #593 — `handle_push_solution`; this branch is stacked on top of it
- Bitcoin Core:
  - [bitcoin/bitcoin#35671](https://github.com/bitcoin/bitcoin/pull/35671) — the `TxCollection` interface (draft; an alternative to [bitcoin/bitcoin#34020](https://github.com/bitcoin/bitcoin/pull/34020), which would have kept the mirror but fed it more efficiently)
- [2140-dev/bitcoin-capnp-types#29](https://github.com/2140-dev/bitcoin-capnp-types/pull/29) — capnp bindings for `TxCollection`

### Also

- Partially implements #597 — prev_hash-only stale classification

- `TxCollection` is unlikely to actually land in Bitcoin Core v32, but this PR pretends it does so we don't need to add plumbing for yet another version module.

### Commits

- _TEMPORARY: pin bitcoin-capnp-types v32 to TxCollection PR branch_ - the `TxCollection` bindings only exist on the 2140-dev/bitcoin-capnp-types#29 branch
- _TEMPORARY: require BITCOIN_CORE_V32_BINARY for integration tests_ - no released Bitcoin Core has `TxCollection`; tests need a locally built bitcoin/bitcoin#35671 binary
- _TEMPORARY: CI builds Bitcoin Core from bitcoin#35671_ - build that binary from source in CI (with ccache) instead of stubbing it
- _JDS: validate DeclareMiningJob wtxid list and provided transactions_ - duplicate wtxids and out-of-set provided transactions are protocol violations; Bitcoin Core rejects them with RPC-level errors, which would otherwise let a misbehaving downstream tear down the shared IPC connection
- _JDP: restrict stale-tip classification to prev_hash_ - `nbits`/`min_ntime` drift can misclassify errors (#597), and the TxCollection backend has no template to take those fields from when transactions are missing
- _v32x JDP: **validate declared jobs with TxCollection, drop mempool mirror**_ - the main change: `collectTxs`/`addMissingTxs`/`unknownTxPos`/`makeTemplate` (with the declared coinbase) replace the mirror and `checkBlock`
- _v32x JDP: **submit solutions via retained BlockTemplate**_ - keep the validated template inside Bitcoin Core and submit solutions via `submitSolution`, removing block reconstruction (and per-job `txdata`) from jd-server
- _integration-tests: cover missing-transactions success flow and coinbase validation_ - the `ProvideMissingTransactions` round was only tested on failure paths, and no test mined a block containing a transaction or rejected a consensus-invalid coinbase


Sjors:
Ironically we still need a limited transaction cache on the JDS side, because the current `TxCollection` does not add transactions to its mempool, does not share transactions between templates and does not hold on to (non-mempool) transactions after a template is released. These are all good future improvements that should not change the interface, but they'll non-trivial.

Instead for now the new commit holds on to any transaction the JDS requested, which avoids round-trips back to the miner. It should be much simpler than the original `mempool.rs` though.

Sjors:
Updated to the latest version of https://github.com/2140-dev/bitcoin-capnp-types/pull/29 which bumped `collectTxs()` to `@10`. It also includes `getTransactionsByWitnessID()`, but we don't use that here.

PR #602, opened 2026-07-07
introduce shared modules to `bitcoin_core_sv2`

### LINKED ISSUE #574: need shared modules on `bitcoin_core_sv2` to avoid code duplication

plebhash:
close #574


Sjors:
I'm always a fan of de-duplication. If this lands before #600 it will make e0647db3fcf10e8864cd81076a73de984d7c68eb `test: add Bitcoin Core master IPC bindings` a bit shorter.

plebhash:
ready for review

bitcoin-core-sv2/src/unix_capnp/v31x_v30x/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Shared implementation modules reused by Bitcoin Core v30.x and v31.x runtimes.
+
+pub mod job_declaration_protocol;
+
+// TDP shared monitors are reused via `#[path = "..."]` from v30x/v31x modules so they compile
+// in each version-local `super::*` context; they are not exported from this module tree.
```
GitGab19, line 6: Why are they treated differently wrt JDP ones?

GitGab19: NVM, I got it now.

GitGab19 (approved):

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -9,7 +9,7 @@ use std::{
 use async_channel::{unbounded, Receiver, Sender};
 use core::sync::atomic::Ordering;
 use stratum_apps::{
-    bitcoin_core_sv2::common::template_distribution_protocol::CancellationToken,
+    bitcoin_core_sv2::runtime_api::template_distribution_protocol::CancellationToken,
```
bit-aloo: I think we should not expose the internal structure of `bitcoin-core-sv2` to users, especially the runtime thingy here. Maybe we could export these constructs from a different level instead.

plebhash: can you elaborate?

having this as a public module is pretty much the main premise of this PR, so it seems we're not aligned AT ALL here

perhaps I have a major blindspot in what you're trying to say here, or perhaps you have a major blindspot with regards to what we're trying to achieve here

either way, unless I'm convinced there's some fundamental flaw with this approach, I would like to avoid hard-pivots

bit-aloo: What I mean is this: `bitcoin_core_sv2::template_distribution_protocol::CancellationToken,` instead of runtime_api in between? or something like this `bitcoin_core_sv2::CancellationTokenTDP,`, without letting user know the internal structuring of our crate.

plebhash: ok thanks for clarifying, your first comment wasn't very precise so I misinterpreted it

so IIUC you're talking about the way we're re-exporting `CancellationToken` specifically... I thought you were saying that `runtime_api` shouldn't be a public module, which contradicts the core idea of this PR and would require a hard pivot into a fundamentally different direction

the rationale for re-exporting `CancellationToken` from this crate is to provide one canonical token type at its API boundary, so consumers use the exact type we expect and avoid integration friction/type-mismatch surprises from direct `tokio_util` imports.

but I agree that it's somewhat redundant to re-export it via `bitcoin_core_sv2::runtime_api::template_distribution_protocol::CancellationToken`, while we could simply have one root `bitcoin_core_sv2::CancellationToken`

that's a refactoring artifact, and you did right to point that out, thanks!

should be fixed now

bit-aloo: Cool

PR #612, opened 2026-07-10
`bitcoin_core_sv2`: clean up outdated `TemplateData` on `CoinbaseOutputConstraints`

### LINKED ISSUE #611: `bitcoin_core_sv2`: `CoinbaseOutputConstraints` should trigger cleanup outdated `TemplateData`
currently, `BitcoinCoreSv2TDP` only cleans up stale `TemplateData` after chain tip updates

we should also clean them up after `CoinbaseOutputConstraints`

plebhash:
close #611

### DIFF
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
@@ -42,6 +42,11 @@ impl BitcoinCoreSv2TDP {
             }
         }
 
+        // Capture the current template IDs after the old monitor has exited and
+        // before bootstrapping the new IPC client. This guarantees the snapshot
+        // only contains templates from the previous constraints epoch.
+        let stale_template_ids = self.current_template_ids()?;
+
         self.template_ipc_client_cancellation_token = CancellationToken::new();
         debug!("Created new template_ipc_client_cancellation_token");
 
@@ -54,6 +59,13 @@ impl BitcoinCoreSv2TDP {
             error!("Failed to bootstrap new template IPC client: {:?}", e);
             e
         })?;
+
+        // Retire templates created under previous constraints after the new client
+        // is live. Without this, old template IDs remain usable indefinitely and
+        // accumulate one live template capability per constraints update.
+        if !stale_template_ids.is_empty() {
+            self.process_stale_template_data(stale_template_ids).await;
+        }
 
         debug!("Spawning new monitor_ipc_templates() task");
         self.monitor_ipc_templates();
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
@@ -42,10 +42,7 @@ impl BitcoinCoreSv2TDP {
             }
         }
 
-        // Capture the current template IDs after the old monitor has exited and
-        // before bootstrapping the new IPC client. This guarantees the snapshot
-        // only contains templates from the previous constraints epoch.
-        let stale_template_ids = self.current_template_ids()?;
+        self.process_stale_template_data().await?;
 
         self.template_ipc_client_cancellation_token = CancellationToken::new();
         debug!("Created new template_ipc_client_cancellation_token");
@@ -59,13 +56,6 @@ impl BitcoinCoreSv2TDP {
             error!("Failed to bootstrap new template IPC client: {:?}", e);
             e
         })?;
-
-        // Retire templates created under previous constraints after the new client
-        // is live. Without this, old template IDs remain usable indefinitely and
-        // accumulate one live template capability per constraints update.
-        if !stale_template_ids.is_empty() {
-            self.process_stale_template_data(stale_template_ids).await;
-        }
 
         debug!("Spawning new monitor_ipc_templates() task");
         self.monitor_ipc_templates();
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
@@ -622,6 +622,10 @@ impl BitcoinCoreSv2TDP {
 
     // Spawns a task that processes stale template data after a 10-second grace period.
     //
+    // Takes a snapshot of [`current_template_ids`] at call time, then schedules their
+    // retirement. This ensures the snapshot is always taken at the epoch boundary rather
+    // than relying on the caller to pre-compute the stale set.
+    //
     // The grace period allows in-flight RequestTransactionData and SubmitSolution requests
     // to complete before the template data is retired. After the 10-second window:
     // - Stale template IDs are written to stale_template_ids, causing
@@ -629,7 +633,11 @@ impl BitcoinCoreSv2TDP {
     // - Stale entries are removed from template_data, causing both handle_request_transaction_data
     //   and handle_submit_solution to return errors.
     // - The underlying IPC client capabilities are released via destroy_ipc_client.
-    async fn process_stale_template_data(&self, stale_template_ids: HashSet<u64>) {
+    async fn process_stale_template_data(&self) -> Result<(), BitcoinCoreSv2TDPError> {
+        let stale_template_ids = self.current_template_ids()?;
+        if stale_template_ids.is_empty() {
+            return Ok(());
+        }
         let self_clone = self.clone();
         tokio::task::spawn_local(async move {
             tokio::time::sleep(std::time::Duration::from_secs(10)).await;
@@ -706,5 +714,7 @@ impl BitcoinCoreSv2TDP {
                 }
             }
         });
+
+        Ok(())
     }
 }
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
@@ -620,9 +620,15 @@ impl BitcoinCoreSv2TDP {
         Ok(wait_next_request)
     }
 
-    // spawns a task that processes the stale template data after 10s
-    // we wait 10s in case there's any incoming RequestTransactionData referring to stale templates
-    // immediately after the chain tip change
+    // Spawns a task that processes stale template data after a 10-second grace period.
+    //
+    // The grace period allows in-flight RequestTransactionData and SubmitSolution requests
+    // to complete before the template data is retired. After the 10-second window:
+    // - Stale template IDs are written to stale_template_ids, causing
+    //   handle_request_transaction_data to return an error.
+    // - Stale entries are removed from template_data, causing both handle_request_transaction_data
+    //   and handle_submit_solution to return errors.
+    // - The underlying IPC client capabilities are released via destroy_ipc_client.
     async fn process_stale_template_data(&self, stale_template_ids: HashSet<u64>) {
         let self_clone = self.clone();
         tokio::task::spawn_local(async move {
--- a/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
@@ -43,6 +43,11 @@ impl BitcoinCoreSv2TDP {
             }
         }
 
+        // Capture the current template IDs after the old monitor has exited and
+        // before bootstrapping the new IPC client. This guarantees the snapshot
+        // only contains templates from the previous constraints epoch.
+        let stale_template_ids = self.current_template_ids()?;
+
         self.template_ipc_client_cancellation_token = CancellationToken::new();
         debug!("Created new template_ipc_client_cancellation_token");
 
@@ -55,6 +60,13 @@ impl BitcoinCoreSv2TDP {
             error!("Failed to bootstrap new template IPC client: {:?}", e);
             e
         })?;
+
+        // Retire templates created under previous constraints after the new client
+        // is live. Without this, old template IDs remain usable indefinitely and
+        // accumulate one live template capability per constraints update.
+        if !stale_template_ids.is_empty() {
+            self.process_stale_template_data(stale_template_ids).await;
+        }
 
         debug!("Spawning new monitor_ipc_templates() task");
         self.monitor_ipc_templates();
--- a/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
@@ -43,10 +43,7 @@ impl BitcoinCoreSv2TDP {
             }
         }
 
-        // Capture the current template IDs after the old monitor has exited and
-        // before bootstrapping the new IPC client. This guarantees the snapshot
-        // only contains templates from the previous constraints epoch.
-        let stale_template_ids = self.current_template_ids()?;
+        self.process_stale_template_data().await?;
 
         self.template_ipc_client_cancellation_token = CancellationToken::new();
         debug!("Created new template_ipc_client_cancellation_token");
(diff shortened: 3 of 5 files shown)


bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
```diff
@@ -20,6 +20,13 @@ impl BitcoinCoreSv2TDP {
     ) -> Result<(), BitcoinCoreSv2TDPError> {
         debug!("handle_coinbase_output_constraints() called");
 
+        // Capture the current template IDs before switching to the new IPC client.
+        // These represent templates created under the previous constraints and must
+        // be retired so that old template capabilities are released and cannot be
+        // used via handle_request_transaction_data or handle_submit_solution after
+        // the new constraint epoch is published.
+        let stale_template_ids = self.current_template_ids()?;
+
```
GitGab19: My clanker found a possible issue which can be easily solved by moving this line after the `.await()`:
```
I think `stale_template_ids` should be collected after the old `monitor_ipc_templates` task has been awaited, but before bootstrapping the new IPC client.

Right now the snapshot happens before cancellation/await, so the old monitor can still publish one more template under the previous `CoinbaseOutputConstraints`. That late template would not be included in `stale_template_ids`, so it would survive the delayed cleanup.

Suggested order: cancel old monitor -> await old monitor -> collect `current_template_ids()` -> bootstrap new client -> schedule stale cleanup.
```

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
```diff
@@ -55,6 +62,13 @@ impl BitcoinCoreSv2TDP {
             e
         })?;
 
+        // Retire templates created under previous constraints after the new client
+        // is live. Without this, old template IDs remain usable indefinitely and
+        // accumulate one live template capability per constraints update.
+        if !stale_template_ids.is_empty() {
+            self.process_stale_template_data(stale_template_ids).await;
```
bit-aloo: For stale template processing, could we get the template IDs inside the method itself? Is there a reason we're not doing this atomically instead of having a time gap between creating the list of IDs and actually operating on them?

```rust
async fn process_stale_template_data(&self,/*We wont be requiring this id list anymore. stale_template_ids: HashSet<u64>*/) {
          let ids = self.current_template_ids();
          ...
}
```

bit-aloo:
@plebhash I saw we are using `Rc<RwLock>` in bitcoin_core_sv2 crate, maybe we should start using `SharedLock` here?

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
plebhash: `process_stale_template_data` is also used on chain-tip update flow, and changing signature would blow up scope and force us to do major refactors

fixing this race window via:
```diff
@@ -20,13 +20,6 @@ impl BitcoinCoreSv2TDP {
     ) -> Result<(), BitcoinCoreSv2TDPError> {
         debug!("handle_coinbase_output_constraints() called");

-        // Capture the current template IDs before switching to the new IPC client.
-        // These represent templates created under the previous constraints and must
-        // be retired so that old template capabilities are released and cannot be
-        // used via handle_request_transaction_data or handle_submit_solution after
-        // the new constraint epoch is published.
-        let stale_template_ids = self.current_template_ids()?;
-
         // Break the loop in monitor_ipc_templates() and spawn a new one after bootstrapping the
         // new template IPC client. We no longer care about templates created under previous
         // constraints for future template monitoring.
@@ -49,6 +42,11 @@ impl BitcoinCoreSv2TDP {
             }
         }

+        // Capture the current template IDs after the old monitor has exited and
+        // before bootstrapping the new IPC client. This guarantees the snapshot
+        // only contains templates from the previous constraints epoch.
+        let stale_template_ids = self.current_template_ids()?;
+
```

doesn't literally make `process_stale_template_data` atomic, but does address the race issue

plebhash:
> @plebhash I saw we are using `Rc<RwLock>` in bitcoin_core_sv2 crate, maybe we should start using `SharedLock` here?

can be investigated/discussed, but out of scope here

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
plebhash: should be fixed with the same patch I mentioned here: https://github.com/stratum-mining/sv2-apps/pull/612#discussion_r3604681250

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
bit-aloo: I looked at the other site, where we are using the `process_stale_template_data`, its also a very similar case as we have here. We can sure club them, it won't really increase the blast radius and changed gonna be very contained. 

Can be done here, or I can make a good first issue out of it :sweat_smile: .

plebhash: can you provide a diff of what you have in mind?

plebhash: also, are we pursuing this for objective reasons, or for styling reasons?

bit-aloo: I had something like this in mind, also contains your changes:

```diff
diff --git a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
index 3ef1c932..47b09678 100644
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
@@ -42,6 +42,8 @@ impl BitcoinCoreSv2TDP {
             }
         }
 
+        self.process_stale_template_data().await?;
+
         self.template_ipc_client_cancellation_token = CancellationToken::new();
         debug!("Created new template_ipc_client_cancellation_token");
 
diff --git a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
index a76f2c1c..1a05cd52 100644
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
@@ -620,10 +620,17 @@ impl BitcoinCoreSv2TDP {
         Ok(wait_next_request)
     }
 
-    // spawns a task that processes the stale template data after 10s
-    // we wait 10s in case there's any incoming RequestTransactionData referring to stale templates
-    // immediately after the chain tip change
-    async fn process_stale_template_data(&self, stale_template_ids: HashSet<u64>) {
+    // Spawns a task that processes stale template data after a 10-second grace period.
+    //
+    // The grace period allows in-flight RequestTransactionData and SubmitSolution requests
+    // to complete before the template data is retired. After the 10-second window:
+    // - Stale template IDs are written to stale_template_ids, causing
+    //   handle_request_transaction_data to return an error.
+    // - Stale entries are removed from template_data, causing both handle_request_transaction_data
+    //   and handle_submit_solution to return errors.
+    // - The underlying IPC client capabilities are released via destroy_ipc_client.
+    async fn process_stale_template_data(&self) -> Result<(), BitcoinCoreSv2TDPError> {
+        let stale_template_ids = self.current_template_ids()?;
         let self_clone = self.clone();
         tokio::task::spawn_local(async move {
             tokio::time::sleep(std::time::Duration::from_secs(10)).await;
@@ -700,5 +707,7 @@ impl BitcoinCoreSv2TDP {
                 }
             }
         });
+
+        Ok(())
     }
 }
diff --git a/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs b/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
index ab08dfd8..b0fe2935 100644
--- a/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/handlers.rs
@@ -43,6 +43,8 @@ impl BitcoinCoreSv2TDP {
             }
         }
 
+        self.process_stale_template_data().await?;
+
         self.template_ipc_client_cancellation_token = CancellationToken::new();
         debug!("Created new template_ipc_client_cancellation_token");
 
diff --git a/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs b/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs
index 58dc4306..5b992c9b 100644
--- a/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs
@@ -653,10 +653,17 @@ impl BitcoinCoreSv2TDP {
         Ok(wait_next_request)
     }
 
-    // spawns a task that processes the stale template data after 10s
-    // we wait 10s in case there's any incoming RequestTransactionData referring to stale templates
-    // immediately after the chain tip change
-    async fn process_stale_template_data(&self, stale_template_ids: HashSet<u64>) {
+    // Spawns a task that processes stale template data after a 10-second grace period.
+    //
+    // The grace period allows in-flight RequestTransactionData and SubmitSolution requests
+    // to complete before the template data is retired. After the 10-second window:
+    // - Stale template IDs are written to stale_template_ids, causing
+    //   handle_request_transaction_data to return an error.
+    // - Stale entries are removed from template_data, causing both handle_request_transaction_data
+    //   and handle_submit_solution to return errors.
+    // - The underlying IPC client capabilities are released via destroy_ipc_client.
+    async fn process_stale_template_data(&self) -> Result<(), BitcoinCoreSv2TDPError> {
+        let stale_template_ids = self.current_template_ids()?;
         let self_clone = self.clone();
         tokio::task::spawn_local(async move {
             tokio::time::sleep(std::time::Duration::from_secs(10)).await;
@@ -733,6 +740,8 @@ impl BitcoinCoreSv2TDP {
                 }
             }
         });
+
+        Ok(())
     }
 }
 
diff --git a/bitcoin-core-sv2/src/unix_capnp/v31x_v30x/template_distribution_protocol/monitors.rs b/bitcoin-core-sv2/src/unix_capnp/v31x_v30x/template_distribution_protocol/monitors.rs
index 86f73ac3..0f76f8d3 100644
--- a/bitcoin-core-sv2/src/unix_capnp/v31x_v30x/template_distribution_protocol/monitors.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v31x_v30x/template_distribution_protocol/monitors.rs
@@ -147,15 +147,12 @@ impl BitcoinCoreSv2TDP {
                                     info!("⛓️ Chain Tip changed! New prev_hash: {}", new_prev_hash);
                                     debug!("CHAIN TIP CHANGE DETECTED - old: {}, new: {}", current_prev_hash, new_prev_hash);
 
-                                    let stale_template_ids = match self_clone.current_template_ids() {
-                                        Ok(stale_template_ids) => stale_template_ids,
-                                        Err(e) => {
-                                            error!("Failed to collect stale template ids: {:?}", e);
-                                            warn!("Terminating Sv2 Bitcoin Core IPC Connection");
-                                            self_clone.global_cancellation_token.cancel();
-                                            break;
-                                        }
-                                    };
+                                    if let Err(e) = self_clone.process_stale_template_data().await {
+                                        error!("Failed to collect stale template ids: {:?}", e);
+                                        warn!("Terminating Sv2 Bitcoin Core IPC Connection");
+                                        self_clone.global_cancellation_token.cancel();
+                                        break;
+                                    }
 
                                     match self_clone.publish_template(new_template_data, true, true, false).await {
                                         Ok(()) => {
@@ -169,9 +166,6 @@ impl BitcoinCoreSv2TDP {
                                             break;
                                         }
                                     }
-
-                                    // process the stale template data after 10s
-                                    self_clone.process_stale_template_data(stale_template_ids).await;
                                 } else {
                                     // check if the minimum interval has been reached
                                     if let Some(last_sent_template_instant) = self_clone.last_sent_template_instant {
```

bit-aloo: > also, are we pursuing this for objective reasons, or for styling reasons?

Oopsie, missed before. Simplicity and atomicity, I think :thinking:

plebhash: done

bit-aloo (approved):
ACK

bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/handlers.rs
GitGab19: Yeah this issue is fixed now, but according to my clanker now we introduced another possible issue (IMO it's very unlikely for this to be a problem in production, but I think it's worth evaluating it):
```
I think we still need to keep “capture stale IDs” separate from “start the stale cleanup timer”.

With the current version, `process_stale_template_data()` captures the IDs and immediately starts the 10s grace-period task before the new template is bootstrapped/published. If bootstrapping takes longer than expected, old templates can be removed before a replacement template is live.

Safer order would be: stop old monitor -> capture current template IDs -> bootstrap/publish the new template -> start the 10s cleanup timer for the captured IDs.
```

Now a blocker for me, as I think it's almost impossible to not receive the new template in the 10s timer used.

plebhash: this came from ad8b9b08de04ea94b0b4b47366ff7ba61bc85728 + 6aaa34d8a24bae32ce080ce2345e40fe9485dd66 which @Shourya742 seemed highly opinionated about and also very confident there would be no side effects

so I'll defer it to him to analyze whether we should keep these or not

bit-aloo: Hmm... I did think about this. My reasoning was that the grace period for template expiry is long enough that it shouldn't really be an issue.

It would be nicer if we could move `process_stale_template_data` to run after a new template is published. However, that introduces the risk of expiring a newly created template. Maybe we could associate a chain tip marker with each template. Then, if a template is not marked with the current tip, we can expire it; otherwise, we keep it.

That said, I don't think this belongs in this PR since its purpose is different (logic gonna become a bit different, if we go with marking scheme).

I am happy to drop these commits and I will open a PR later to do whats described.

GitGab19: It seems there's a simpler way to achieve that, but we would have to modify the `process_stale_template_data` you proposed.

Here the clanker suggestion:
```
I think the intended shape is:

```rust
// 1. Capture only templates from the old epoch.
let stale_template_ids = self.current_template_ids()?;

// 2. Publish/activate the replacement template first.
self.bootstrap_template_ipc_client_from_coinbase_output_constraints(
    coinbase_output_constraints,
)
.await?;

// 3. Only now start the 10s grace-period cleanup timer.
self.process_stale_template_data(stale_template_ids).await;

So process_stale_template_data should probably keep taking the captured HashSet<u64> instead of calling current_template_ids() internally. That avoids both problems: we don’t accidentally include the new template, and we don’t start the cleanup timer before the replacement template is live.

```

bit-aloo: Yup, that what we have pre https://github.com/stratum-mining/sv2-apps/commit/ad8b9b08de04ea94b0b4b47366ff7ba61bc85728 + https://github.com/stratum-mining/sv2-apps/commit/6aaa34d8a24bae32ce080ce2345e40fe9485dd66 these commit.

plebhash: tbh I feel the way it is now is fine

the only scenario where the 10s invariant would be broken is if there's some very nasty bug on Bitcoin Core, in which case we would have much bigger problems

I'd merge this PR as it is now (including https://github.com/stratum-mining/sv2-apps/commit/ad8b9b08de04ea94b0b4b47366ff7ba61bc85728 + https://github.com/stratum-mining/sv2-apps/commit/6aaa34d8a24bae32ce080ce2345e40fe9485dd66)

PR #614, opened 2026-07-13
Fix payout verification for split coinbase scriptSig

GitGab19:
This is a follow-up PR to https://github.com/stratum-mining/sv2-apps/pull/497.

The main problem this solves is the payout verification on `NewExtendedMiningJob` messages which contain the last part of the scriptSig in the `coinbase_tx_suffix`, instead of having it entirely in the `coinbase_tx_prefix`.

I checked on the specs, and there's no mention to the fact that a `coinbase_tx_suffix` has to start with `nSequence` bytes, which was the assumption used by the mechanism introduced in https://github.com/stratum-mining/sv2-apps/pull/497.

I also added a new entry into the `translator_mines_when_payout_matches_address_or_donation_identity` IT to ensure we correctly deal with such cases.

The issue has been found while testing our tProxy against [NexusPool](https://nexuspool.io/#connect) endpoint, and this PR fixes that.

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -2669,7 +2669,7 @@ dependencies = [
 
 [[package]]
 name = "integration_tests_sv2"
-version = "0.4.0"
+version = "0.5.0"
 dependencies = [
  "async-channel 1.9.0",
  "clap",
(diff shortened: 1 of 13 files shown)


plebhash:
LGTM, but we should bump `stratum-apps` and `translator_sv2`

bit-aloo (approved):
LGTM. After Plebhash's comment on version bumps.

GitGab19:
Turned this into draft since I want to discuss more about the version bumps we need to do here before merging this.

GitGab19:
Ok, I confirmed with @plebhash that the bumps are correct.

plebhash (approved):

PR #615, opened 2026-07-14
discover ASIC telemetry by miner's username and port 

GitGab19:
This PR unblocks and fixes the issue we faced while working on https://github.com/stratum-mining/sv2-ui/pull/202

It leverages the discovery feature of `asic-rs` to find miners into the configured subnet, and matches them with the clients connected to tProxy and/or JDC by comparing the pool's username returned by `asic-rs` (and the port) with the `authorized_worker_name` (in case of tProxy) and the `user_identity` field (in case of JDC).

It also adds some logic to detect duplicated usernames between different miners which might be running under the same subnet.

I still have to fully review the code, but on my end it works inside docker environment as well.

It would be good to get an early test on macOS from @lucasbalieiro :pray:


GitGab19:
I just tested this on my old macOS, with Docker Desktop, without enabling host networking, and it worked!

photo_2026-07-15_10-15-59

cc @lucasbalieiro

lucasbalieiro:
Unable to reproduce the success you've achieved.

Running with docker desktop v4.81.0 on a MacOs with host network configs disabled

I've builded the docker images based on this PR and used on the sv2-ui at the pr https://github.com/stratum-mining/sv2-ui/pull/202 (dropping the latest commit added there, which changes the networks to host)


But As I expected the things are still being `NAT`ed.
image

GitGab19:
Today we were able to test this together on @lucasbalieiro macOS, and it worked as well.

This PR is now ready for reviews :)

bit-aloo (approved):
ACK

PR #624, opened 2026-07-17
tProxy: build `UserIdentity` TLV only when extension `0x0002` is negotiated inside ChannelManager

### LINKED ISSUE #399: tProxy needs a way to deal with TLV `UserIdentity` bigger than 32 bytes
#295 reported an issue where tProxy was panicking whenever the Sv1 `username` (which later becomes TLV `UserIdentity`) is bigger than 32 bytes

#395 applied some temporary hotpatch, with a few compromises:
- we do a lossy truncation of the stored `user_identity` copy of the `username` field of Sv1 `mining.authorize`, leaving room for collisions
- this truncation happens with a warning, which is triggered even if extension `0x0002` is not active, leaving room for confusing UX

These compromises should be replaced with more sustainable long-term solutions.

We should also define clearer boundaries for:

- the config `user_identity` used for `OpenExtendedMiningChannel`
- the Sv1 worker name from `mining.authorize`
- the 32-byte TLV `UserIdentity` used by extension `0x0002`

---

relevant note: https://github.com/stratum-mining/sv2-apps/issues/295#issuecomment-4154753179

GitGab19:
This PR improves the way tProxy handles `UserIdentity` TLV, by moving its construction in the ChannelManager, used only when the extension `0x0002` is negotiated. 

Given that the idea of the extension is to send the real worker_names of the Sv1 miners up to the Pool, now the TLV is built on the `sv1_worker_name` field, extracted from the larger `sv1_username`.

An example is the following one:
- `sv1_username` --> bc1qtzqxqaxyy6lda2fhdtp5dp0v56vlf6g0tljy2x.Sv2Bitaxe
- `sv1_worker_name` --> Sv2Bitaxe

Both fields are exposed through the monitoring API, which now doesn't expose the removed (useless) UserIdentity TLV anymore.

This PR also allows for some improvements on `sv2-ui`, removing the ambiguities reported by some users related to the UserIdentity (currently truncated) showed there now.

Closes #399.

Some screenshot of the end result:

image

image


plebhash:
this comment should be updated:
https://github.com/stratum-mining/sv2-apps/blob/bf5b3f389a33f06294483ba723658d5404a5002c/integration-tests/tests/extensions.rs#L40-L41

plebhash:
another thing that's getting me a bit confused, not necessarily a finding on this PR

it seems that before this PR, we were gating the `UserIdentity` TLV on `is_non_aggregated()`

isn't that the exact opposite of what extension `0x0002` aims to achieve? I would expect the TLVs to kick in when tProxy is launched in Aggregated mode

this PR fixes that anyways (gates it based on whether the extension was actually negotiated or not, regardless of mode), but I wanted to clarify this detail for my own understanding

plebhash:
also as a follow-up to the question above:

IIUC now the extension can be used both in aggregated mode and non-aggregated mode

is that intentional? if not, can it cause problems when running in non-aggregated mode?

---

edit: asked my clanker and it said it's redundant but benign

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -84,15 +83,16 @@ struct Sv1ServerIo {
     sv1_server_to_downstream_sender: Arc<Mutex<HashMap<DownstreamId, Sender<json_rpc::Message>>>>,
     downstream_to_sv1_server_sender: Sender<(DownstreamId, json_rpc::Message)>,
     downstream_to_sv1_server_receiver: Receiver<(DownstreamId, json_rpc::Message)>,
-    channel_manager_receiver: Receiver<(Mining<'static>, Option<Vec<Tlv>>)>,
-    channel_manager_sender: Sender<(Mining<'static>, Option<Vec<Tlv>>)>,
+    channel_manager_receiver: Receiver<Mining<'static>>,
```
bit-aloo, line 86: why no TLV?

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
```diff
@@ -650,60 +647,47 @@ impl ChannelManager {
...
                     });
 
-                // Check if we should try to include TLV fields
+                // Check if we should include the extension 0x0002 TLV field.
                 let should_send_with_tlv =
-                    contains_type_in_negotiated_extension && tlv_fields.is_some();
+                    contains_type_in_negotiated_extension && sv1_worker_name.is_some();
 
                 let mut sent = false;
                 if should_send_with_tlv {
```
bit-aloo:
```rust
if contains_type_in_negotiated_extension && Some(sv1_worker_name) = sv1_worker_name {
```

miner-apps/translator/src/lib/utils.rs
```diff
@@ -265,27 +266,96 @@ pub(crate) fn is_mining_authorize(msg: &Message) -> bool {
...
-pub(crate) fn tlv_compatible_username(s: &str) -> &str {
-    const MAX_USER_IDENTITY_BYTES: usize = 32;
-    let len = s.len();
-
-    if len <= MAX_USER_IDENTITY_BYTES {
-        return s;
+/// This is intentionally TLV-specific: the full SV1 username and derived worker
+/// name stay unmodified in translator state, and truncation only happens when
+/// extension 0x0002 has been negotiated and the wire TLV is being built.
+pub(crate) fn tlv_user_identity_from_sv1_worker_name(sv1_worker_name: &str) -> UserIdentity {
```
bit-aloo: Lets return a result in this.

miner-apps/translator/src/lib/utils.rs
```diff
@@ -265,27 +266,96 @@ pub(crate) fn is_mining_authorize(msg: &Message) -> bool {
...
+    let tlv_user_identity = truncate_to_tlv_user_identity_len(sv1_worker_name);
+
+    UserIdentity::new(tlv_user_identity)
+        .expect("TLV user identity is truncated to the maximum UserIdentity length")
```
bit-aloo: We can let the errror be shutdown here?

miner-apps/translator/src/lib/utils.rs
```diff
@@ -265,27 +266,96 @@ pub(crate) fn is_mining_authorize(msg: &Message) -> bool {
...
-    const MAX_USER_IDENTITY_BYTES: usize = 32;
-    let len = s.len();
-
-    if len <= MAX_USER_IDENTITY_BYTES {
-        return s;
+/// This is intentionally TLV-specific: the full SV1 username and derived worker
+/// name stay unmodified in translator state, and truncation only happens when
+/// extension 0x0002 has been negotiated and the wire TLV is being built.
+pub(crate) fn tlv_user_identity_from_sv1_worker_name(sv1_worker_name: &str) -> UserIdentity {
+    let tlv_user_identity = truncate_to_tlv_user_identity_len(sv1_worker_name);
```
bit-aloo: Can we inline the truncate_to_tlv_user_identity_len here?

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -712,43 +712,26 @@ impl Sv1Server {
...
+        let sv1_worker_name = downstream
+            .downstream_data
+            .super_safe_lock(|d| d.sv1_worker_name.clone());
 
         self.sv1_server_io
             .channel_manager_sender
             .send((
                 Mining::SubmitSharesExtended(submit_share_extended),
-                tlv_fields,
+                Some(sv1_worker_name),
```
bit-aloo: Shouldn't we check if worker name is empty or not?

bit-aloo (commented):
This is what my clanker said:

```diff
Missing end-to-end coverage for the new split behavior
The extension integration test still uses SRI-miner without a dot, so it does not prove that account.SRI-miner produces TLV SRI-miner. The unit test covers the helper, but not the actual framed TLV path. I’d add/adjust an integration case around [extensions.rs (line 148)](/home/shourya/sv2-apps/integration-tests/tests/extensions.rs:148).
```

GitGab19:
> isn't that the exact opposite of what extension 0x0002 aims to achieve? I would expect the TLVs to kick in when tProxy is launched in Aggregated mode

Yeah that's a fair question: the idea was to use it in non-aggregated mode in order to be able to "transmit" the real Sv1 worker_name (through TLV) upstream (you remember that tProxy doesn't get the real worker_name before the Sv1 `mining.authorize` which happens after the channel opening).

> is that intentional? if not, can it cause problems when running in non-aggregated mode?

It's intentional, because the extension can theoretically help in both cases. I don't see any issues when using it in non-aggregated mode.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Because the (optional) worker_name is sent by the Sv1Server to the ChannelManager, not vice versa.

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
GitGab19: changing this with:

```
if contains_type_in_negotiated_extension {
                    if let Some(sv1_worker_name) = sv1_worker_name
                        .as_deref()
                        .filter(|sv1_worker_name| !sv1_worker_name.is_empty())
```

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Adding the check :+1:

miner-apps/translator/src/lib/utils.rs
GitGab19: Done

plebhash (approved):

bit-aloo (approved):
ACK. Thanks

PR #630, opened 2026-07-20
Validate downstream shares against the advertised pow2 difficulty

ckolivas:
This pull request is dependent on https://github.com/stratum-mining/stratum/pull/2227

Validate downstream shares against the advertised pow2 difficulty and filter the band below the upstream target.

mining.set_difficulty is rounded to an integer power of two, but shares were validated against the exact upstream target. With difficulty rounded down (stratum fix-sv1-difficulty-rounding) shares between the advertised difficulty and the upstream target are expected: validate against the advertised target so honest miners are acknowledged, track the exact upstream target separately, and filter the band from upstream submission with debug logging. With the previous round-up behaviour the band is empty and these changes are inert; without them, round-down would surface the band as miner-visible share rejections.

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -1025,7 +1025,7 @@ checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
 [[package]]
 name = "binary_sv2"
 version = "6.0.0"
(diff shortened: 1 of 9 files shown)


GitGab19 (approved):

bit-aloo (approved):
ACK

PR #632, opened 2026-07-20
fix `bitcoin_core_sv2` unintentional `min_interval` delays on chain-tip updates

### LINKED ISSUE #631: `bitcoin_core_sv2`: `min_interval` is delaying chain-tip updates
`bitcoin_core_sv2` has a configuration parameter called `min_interval`

its intended purpose is to throttle down `NewTemplate` messages, as long as they are NOT related to chain-tip updates

currently, that is not being properly enforced, because chain-tip updates are also being delayed under `min_interval` policy

---

that's because the internal implementation has a fundamental flaw

it's based on a `sleep` that gets naively triggered inside the `else` block of `if new_prev_hash != current_prev_hash { ... } else { ... }`

https://github.com/stratum-mining/sv2-apps/blob/987bac3969a25aa70cf6d5280a4280c076b7f386/bitcoin-core-sv2/src/unix_capnp/v31x_v30x/template_distribution_protocol/monitors.rs#L146

https://github.com/stratum-mining/sv2-apps/blob/987bac3969a25aa70cf6d5280a4280c076b7f386/bitcoin-core-sv2/src/unix_capnp/v31x_v30x/template_distribution_protocol/monitors.rs#L175-L189

the problem is that the `new_prev_hash != current_prev_hash` verification can only happen after `waitNext` has already returned, so while we're waiting for `sleep` there's no `waitNext` at all

and if a chain-tip update happens during the `sleep`, we only detect it after the `sleep` expires — delaying `NewTemplate` + `SetNewPrevHash` by up to `min_interval` seconds.

moreover, the fee-update template fetched before the `sleep` is published after the `sleep`, by which time its `prev_hash` may already be stale — the consumer briefly receives a template for the old tip right before the real one lands.

so this approach is fundamentally flawed, and probably should not be based on a `sleep` at all

---

we should follow the same pattern established on `sv2-tp`, which is:

during the `min_interval` window, `fee_threshold = MAX_MONEY` (Bitcoin Core's consensus max 21M BTC), which essentially means that `waitNext` will only return on chain-tip updates

after the `min_interval` window elapsed, `fee_threshold` returns to the configured value. The `waitNext` timeout is capped at the remaining window so the request wakes exactly when the throttle expires.

plebhash:
close #631

### DIFF
--- a/bitcoin-core-sv2/src/lib.rs
+++ b/bitcoin-core-sv2/src/lib.rs
@@ -41,3 +41,13 @@ pub use tokio_util::sync::CancellationToken;
 
 /// The minimum block reserved weight established by Bitcoin Core.
 pub const MIN_BLOCK_RESERVED_WEIGHT: u64 = 2000;
+
+/// Bitcoin Core's `MAX_MONEY` consensus constant, in satoshis (21,000,000 BTC).
+///
+/// Used as a `fee_threshold` sentinel in `waitNext` requests: Bitcoin Core skips fee-based
+/// template updates when `fee_threshold >= MAX_MONEY`, while still returning a new template
+/// immediately on chain tip changes.
+pub const MAX_MONEY: i64 = 21_000_000 * 100_000_000;
+
+/// Max time a `waitNext` request is allowed to block before timing out (in milliseconds).
+pub const WAIT_NEXT_TIMEOUT_MS: f64 = 10_000.0;
--- a/bitcoin-core-sv2/src/runtime_api/template_distribution_protocol/mod.rs
+++ b/bitcoin-core-sv2/src/runtime_api/template_distribution_protocol/mod.rs
@@ -10,7 +10,9 @@
 //!   `SetNewPrevHash`, and transaction-data responses).
 //!
 //! `fee_threshold` controls template refreshes driven by mempool fee deltas, while `min_interval`
-//! enforces a minimum spacing between mempool-driven template updates.
+//! enforces a minimum spacing between mempool-driven template updates. Chain tip updates are never
+//! throttled: while inside the `min_interval` window, `waitNext` requests are issued with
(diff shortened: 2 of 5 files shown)


GitGab19 (approved):
I checked out this PR, and now the time between receiving the new tip and generating the new template is consistently around 1-2ms 🥳 

```
2026-07-20T21:06:44.547307Z UpdateTip: new best=0000000000000000000085897a779af70a9ed957ce1e7462a00facb8cce3579b height=958929 version=0x20746000 log2_work=96.291143 tx=1401344303 date='2026-07-20T21:06:15Z' progress=1.000000 cache=33.9MiB(215424txo)
2026-07-20T21:06:44.548553Z CreateNewBlock(): block weight: 731208 txs: 666 fees: 452485 sigops 2110
```

```
2026-07-20T21:14:20.752269Z UpdateTip: new best=00000000000000000000fd79263438f37f7938fa182082d500b89010873c9dde height=958930 version=0x2000e000 log2_work=96.291151 tx=1401348708 date='2026-07-20T21:16:23Z' progress=1.000000 cache=33.9MiB(223365txo)
2026-07-20T21:14:20.754072Z CreateNewBlock(): block weight: 1235202 txs: 1032 fees: 593458 sigops 2731
```

```
2026-07-20T21:15:10.231209Z UpdateTip: new best=00000000000000000001b1df831c69fa12d35ede343e17e5804c99ea5f7e35cd height=958931 version=0x21848000 log2_work=96.291159 tx=1401351702 date='2026-07-20T21:15:01Z' progress=1.000000 cache=33.9MiB(227183txo)
2026-07-20T21:15:10.232373Z CreateNewBlock(): block weight: 726705 txs: 840 fees: 259560 sigops 1825
```

tACK

bit-aloo (approved):
Awesome, tACK, way less latency.

PR #636, opened 2026-07-21
Remove reset methods

### LINKED ISSUE #634: JDC `ChannelManager::reset_runtime_state` is redundant
ever since #225, this function (which used to be `ChannelManagerData::reset`, now renamed into `ChannelManager::reset_runtime_state`) is now redundant and can be safely removed

also no point in calling `self.upstream_state.set(UpstreamState::SoloMining)` here:
https://github.com/stratum-mining/sv2-apps/blob/eccb251810e874fd89e743c4f83886bf76743a20/miner-apps/jd-client/src/lib/channel_manager/mod.rs#L761-L770

this codeblock should be simply `info!` then `break`

### LINKED ISSUE #635: tProxy `ChannelManager::reset_state` is redundant
ever since #225, this function is now redundant and can be safely removed

even though it was introduced after #225 via #484 and it's totally orthogonal to receiver draining in `ChannelManagerIo::close()`

bit-aloo:
closes: #634 #635

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/mod.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/mod.rs
@@ -755,7 +755,6 @@ impl ChannelManager {
                     }
                     _ = fallback_token.cancelled() => {
                         info!("Channel Manager: fallback triggered");
-                        self.upstream_state.set(UpstreamState::SoloMining);
                         break;
                     }
                     res = &mut vardiff_future => {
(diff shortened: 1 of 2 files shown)


miner-apps/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -816,13 +752,8 @@ impl ChannelManager {
                         break;
                     }
                     _ = fallback_token.cancelled() => {
-                        info!("Channel Manager: fallback triggered, resetting state");
+                        info!("Channel Manager: fallback triggered");
                         self.upstream_state.set(UpstreamState::SoloMining);
```
plebhash:
```suggestion

```

this line can be removed too

bit-aloo: Removed

plebhash (approved):

PR #639, opened 2026-07-21
Replace custom mutex with shared primitives in IT

bit-aloo:
This is part of ongoing effort to deprecate custom mutex with shared primitive

Part of #499


plebhash (approved):
assuming #641 is going to bump `integration_tests_sv2` (which would overlap with bump requirements here)

LGTM

bit-aloo:
Updating versions here: https://github.com/stratum-mining/sv2-apps/pull/641, so not updating in this PR.

PR #640, opened 2026-07-21
stratum-apps: add `SharedSet` synchronization wrapper

plebhash:
follow-up to #368


bit-aloo (approved):
Might come in handy later, I would also like to a primitive for vector, as we do have its usage across apps.

PR #641, opened 2026-07-21
Replace custom mutex/dashmap with shared primitves in JDS

bit-aloo:
This is part of ongoing effort to deprecate custom mutex with shared primitive

Part of https://github.com/stratum-mining/sv2-apps/issues/499

### DIFF
--- a/stratum-apps/src/sync.rs
+++ b/stratum-apps/src/sync.rs
@@ -246,6 +246,14 @@ impl<K: Eq + Hash + Clone, V> SharedMap<K, V> {
         self.0.remove(key)
     }
 
+    /// Remove a key only when the predicate matches the current entry.
+    pub fn remove_if<F>(&self, key: &K, f: F) -> Option<(K, V)>
+    where
+        F: FnOnce(&K, &V) -> bool,
+    {
+        self.0.remove_if(key, f)
+    }
+
     /// Check if a key exists.
     pub fn contains_key(&self, key: &K) -> bool {
         self.0.contains_key(key)
--- a/stratum-apps/src/sync.rs
+++ b/stratum-apps/src/sync.rs
@@ -247,6 +247,9 @@ impl<K: Eq + Hash + Clone, V> SharedMap<K, V> {
     }
 
     /// Remove a key only when the predicate matches the current entry.
+    ///
+    /// Caution: `f` runs while `DashMap` is mutating internal shards. Avoid
+    /// re-entering this `SharedMap` from inside the predicate.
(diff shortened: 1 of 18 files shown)


stratum-apps/src/sync.rs
```diff
@@ -246,6 +246,14 @@ impl<K: Eq + Hash + Clone, V> SharedMap<K, V> {
...
     }
 
+    /// Remove a key only when the predicate matches the current entry.
+    pub fn remove_if<F>(&self, key: &K, f: F) -> Option<(K, V)>
+    where
+        F: FnOnce(&K, &V) -> bool,
+    {
+        self.0.remove_if(key, f)
+    }
+
```
plebhash, line 259: since we're adding this to `SharedMap`, shouldn't we also take the opportunity to add it to `SharedSet`?

bit-aloo: I was not aware we have this method in `set`, I will update set to include this then, thanks.

stratum-apps/src/sync.rs
```diff
@@ -246,6 +246,14 @@ impl<K: Eq + Hash + Clone, V> SharedMap<K, V> {
         self.0.remove(key)
     }
 
+    /// Remove a key only when the predicate matches the current entry.
```
plebhash, line 249: let's make sure this has a warning about re-entrance like the other methods

(also the one in `SharedSet`)

plebhash:
we should bump:
- `stratum-apps`: `v0.7.0` -> `v0.8.0`
- `jd_server_sv2`: `v0.5.0` -> `v0.6.0`
- `pool_sv2`: `v0.6.0` -> `v0.7.0`
- `integration_tests_sv2`: `v0.5.0` -> `v0.6.0`

and take note on #646

stratum-apps/src/sync.rs
bit-aloo: Its already present in `SharedSet` :laughing:

bit-aloo:
> we should bump:
> 
> * `stratum-apps`: `v0.7.0` -> `v0.8.0`
> * `jd_server_sv2`: `v0.5.0` -> `v0.6.0`
> * `pool_sv2`: `v0.6.0` -> `v0.7.0`
> * `integration_tests_sv2`: `v0.5.0` -> `v0.6.0`
> 
> and take note on #646

It would bump jd-client and tproxy as well, considering they depend on stratum-apps.

stratum-apps/src/sync.rs
bit-aloo: done.

plebhash (approved):

PR #642, opened 2026-07-21
revert shared modules on `bitcoin_core_sv2`

### LINKED ISSUE #619: need to improve shared modules on `bitcoin_core_sv2`
#574 and its PR #602 allowed us to introduce shared modules into `bitcoin_core_sv2`, which reduces code duplication

however, as I work on #268 and #441 I'm noticing that there's still things to improve here

ultimately we should have a crate module structure that's easy and intuitive to navigate and maintain, but we're not there yet

plebhash:
close #619


bit-aloo (approved):
ACK

GitGab19 (approved):

PR #644, opened 2026-07-23
refactor(translator): replace monolithic start loop with TranslatorRuntime typestate machine

### LINKED ISSUE #527: need to refactor `TranslatorSv2::start`
as described by @Shourya742 below:

nulllpc:
Refactors the monolithic `start` function inside the Pool (`miner-apps/translator`) into a structured, typestate-based state machine (`TranslatorRuntime`) to manage bootstrap, component lifecycles, and graceful shutdown.

This PR also properly handles errors instead of calling `unwrap` or `expect`

Fixes https://github.com/stratum-mining/sv2-apps/issues/527

### DIFF
--- a/miner-apps/translator/src/lib/error.rs
+++ b/miner-apps/translator/src/lib/error.rs
@@ -61,6 +61,12 @@ pub enum Action {
     Shutdown,
 }
 
+impl Action {
+    pub fn is_shutdown(self) -> bool {
+        matches!(self, Self::Shutdown)
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum LoopControl {
     Continue,
--- a/miner-apps/translator/src/lib/mod.rs
+++ b/miner-apps/translator/src/lib/mod.rs
@@ -10,36 +10,22 @@
 //! provides the `start` method as the main entry point for running the translator service.
 //! It relies on several sub-modules (`config`, `downstream_sv1`, `upstream_sv2`, `proxy`, `status`,
 //! etc.) for specialized functionalities.
-#![allow(clippy::module_inception)]
-use async_channel::{Receiver, Sender, unbounded};
-use std::{
-    net::SocketAddr,
-    sync::{
-        Arc,
-        atomic::{AtomicBool, Ordering},
-    },
-    time::Duration,
-};
-use stratum_apps::{
-    fallback_coordinator::FallbackCoordinator,
-    payout::PayoutMode,
-    task_manager::TaskManager,
-    utils::types::{GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS, Sv2Frame},
+use error::TproxyErrorKind;
+use std::sync::{
+    Arc,
+    atomic::{AtomicBool, Ordering},
 };
 use tokio::sync::Notify;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, info, warn};
+use tracing::{error, info};
 
 pub use stratum_apps::stratum_core::sv1_api::server_to_client;
 
 use config::TranslatorConfig;
 
-use crate::{
-    error::TproxyErrorKind,
-    sv1::Sv1Server,
-    sv2::{ChannelManager, Upstream},
-    utils::{TproxyMode, UpstreamEntry},
-};
+use crate::utils::TproxyMode;
+
+use translator_runtime::{Init, RuntimeEvent, TranslatorRuntime};
 
 pub mod config;
 pub mod error;
@@ -48,6 +34,7 @@ mod io_task;
 mod monitoring;
 pub mod sv1;
 pub mod sv2;
+mod translator_runtime;
 pub mod utils;
 
 /// The main struct that manages the SV1/SV2 translator.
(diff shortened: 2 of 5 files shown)


miner-apps/translator/src/lib/error.rs
```diff
@@ -52,7 +52,7 @@ pub struct TproxyError<Owner> {
     _owner: PhantomData<Owner>,
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
bit-aloo: We can remove these? with specific method

miner-apps/translator/src/lib/mod.rs
```diff
@@ -74,342 +60,51 @@ impl TranslatorSv2 {
...
     ///
-    /// This method starts the main event loop, which handles connections,
-    /// protocol translation, job management, and status reporting.
+    /// The startup and execution sequence follows:
+    /// 1. **Initialize:** Sets up the translator runtime state machine ([`TranslatorRuntime`]).
+    /// 2. **Bootstrap:** Configures internal channels, starts the SV1 server, initializes the Channel Manager,
+    ///    and establishes upstream SV2 connections.
+    /// 3. **Run & Loop:** Spawns active background loops/servers and handles fallbacks or graceful shutdown.
+    /// 4. **Teardown:** Performs a coordinated graceful cleanup of all services and tasks upon termination.
     pub async fn start(self) {
```
bit-aloo: Lets return error similar to JDC and Pool

miner-apps/translator/src/lib/mod.rs
```diff
@@ -74,342 +60,51 @@ impl TranslatorSv2 {
...
-            channel_manager.clone(),
-            cancellation_token.clone(),
-            fallback_coordinator.clone(),
-            task_manager.clone(),
-        )
-        .await;
+        let runtime = match TranslatorRuntime::<Init>::new(self.clone()) {
+            Ok(runtime) => runtime,
+            Err(_) => return,
+        };
```
bit-aloo, line 94: We should mark tproxy as dead, same as JDC suggestions.

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,597 @@
...
+}
+
+pub(super) struct ChannelManagerReady {
+    io: Io,
+    sv1_server: Arc<Sv1Server>,
+    channel_manager: Arc<ChannelManager>,
+}
+
+pub(super) struct UpstreamReady {
+    sv1_server: Arc<Sv1Server>,
```
bit-aloo, line 69: Non default builds make it deadcode?

```rust
arning: field `sv1_server` is never read
  --> translator/src/lib/translator_runtime.rs:72:5
   |
71 | pub(super) struct UpstreamReady {
   |                   ------------- field in this struct
72 |     sv1_server: Arc<Sv1Server>,
   |     ^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `translator_sv2` (lib) generated 1 warning
```

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,597 @@
...
+    fn from((kind, runtime): (TproxyErrorKind, TranslatorRuntime<State>)) -> Self {
+        Self {
+            kind,
+            runtime: runtime.into_failed(),
+        }
+    }
+}
+
+impl TranslatorRuntime<Init> {
+    pub fn new(translator: TranslatorSv2) -> Result<Self, TproxyErrorKind> {
```
bit-aloo: Can you look into visibility identifiers?

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,597 @@
...
+    utils::{TproxyMode, UpstreamEntry},
+    TranslatorSv2,
+};
+
+struct Io {
+    channel_manager_to_upstream_sender: Sender<Sv2Frame>,
+    channel_manager_to_upstream_receiver: Receiver<Sv2Frame>,
+    upstream_to_channel_manager_sender: Sender<Sv2Frame>,
+    upstream_to_channel_manager_receiver: Receiver<Sv2Frame>,
+    channel_manager_to_sv1_server_sender: Sender<(Mining<'static>, Option<Vec<Tlv>>)>,
```
par1ram: [P1] Update these channels to the current `main` interfaces. On synthetic merge `fc3b0c4a3`, this tuple type makes both `Sv1Server::new` and `ChannelManager::new` fail with E0308: the manager-to-SV1 channel now carries `Mining<'static>` directly, and the SV1-to-manager channel carries `(Mining<'static>, Option<String>)`. I reproduced with `cargo check -p translator_sv2`; current base `44af69e55` passes the same command. Because the existing checks predate that base change, this PR needs a rebase/type update before it can build.

par1ram (changes requested):
I found a blocking integration regression on the latest merge with `main`. `cargo check -p translator_sv2` fails with two E0308 type mismatches in `bootstrap_sv1_server` and `bootstrap_channel_manager`, while the same command passes on the current base commit (`44af69e55`). The existing green checks are from July 23 and do not cover the refreshed merge. Please rebase and update the new `Io` channel types to the current `Sv1Server`/`ChannelManager` interfaces, then rerun CI.

nulllpc:
PR is ready for review @plebhash @bit-aloo

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,601 @@
...
+
+        tokio::select! {
+            biased;
+            _ = cancellation_token.cancelled() => RuntimeEvent::Shutdown,
+            _ = fallback_token.cancelled() => RuntimeEvent::Fallback,
+            _ = tokio::signal::ctrl_c() => {
+                info!("Ctrl+C received — initiating graceful shutdown...");
+                cancellation_token.cancel();
+                RuntimeEvent::Shutdown
+            }
```
plebhash: please refer to https://github.com/stratum-mining/sv2-apps/issues/688

`ctrl_c` should be limited to `main.rs` only


```suggestion
```

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,601 @@
...
+    channel_manager: Arc<ChannelManager>,
+    sv1_server: Arc<Sv1Server>,
+}
+
+pub(super) struct UpstreamReady {
+    sv1_server: Arc<Sv1Server>,
+    channel_manager: Arc<ChannelManager>,
+}
+
+pub(super) struct Running;
```
plebhash, line 73: no `Io` here

was it deliberate? can you explain why?

miner-apps/translator/src/lib/mod.rs
```diff
@@ -74,337 +61,76 @@ impl TranslatorSv2 {
...
-    /// protocol translation, job management, and status reporting.
-    pub async fn start(self) {
+    /// The startup and execution sequence follows:
+    /// 1. **Initialize:** Sets up the translator runtime state machine ([`TranslatorRuntime`]).
+    /// 2. **Bootstrap:** Configures internal channels, initializes the Channel Manager, instantiates
+    ///    the SV1 server, and establishes upstream SV2 connections.
+    /// 3. **Run & Loop:** Activates all background loops/servers (SV1, Channel Manager, Monitoring)
+    ///    and handles fallbacks or graceful shutdown.
+    /// 4. **Teardown:** Performs a coordinated graceful cleanup of all services and tasks upon termination.
+    pub async fn start(self) -> Result<(), TproxyErrorKind> {
```
plebhash: both Pool and JDC use `&self`


```suggestion
    pub async fn start(&self) -> Result<(), TproxyErrorKind> {
```

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,601 @@
...
+    }
+}
+
+pub(super) enum RuntimeEvent {
+    Shutdown,
+    Fallback,
+}
+
+impl TranslatorRuntime<Running> {
+    pub(super) async fn wait(&mut self) -> RuntimeEvent {
```
plebhash: no need for `mut` here

```suggestion
    pub(super) async fn wait(&self) -> RuntimeEvent {
```

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -0,0 +1,601 @@
...
+impl TranslatorRuntime<ChannelManagerReady> {
+    fn bootstrap_sv1_server(self) -> TranslatorRuntime<Sv1ServerReady> {
+        // Safe to unwrap here because downstream_address was already validated in
+        // TranslatorRuntime::new.
+        let downstream_addr_ip = self
+            .translator
+            .config
+            .downstream_address
+            .parse::<std::net::IpAddr>()
+            .unwrap();
```
plebhash: even though this `unwrap` is arguably safe, we could save it by introducing a new `listen_address` as a struct member of `TranslatorRuntime`

parse it 1x on `new` constructor, and then never have to `unwrap` again

miner-apps/translator/src/lib/translator_runtime.rs
nulllpc: My bad, I missed this one

miner-apps/translator/src/lib/translator_runtime.rs
nulllpc: Yes, this was deliberate. At `UpstreamReady`, all components have been bootstrapped. From there, we just start services so no need to carry Io through. Also, we create fresh channels upon fallback so no need for Io either

bit-aloo: Why do we have IO in JDC then?

nulllpc: Because JDC reuses some channels from the old Running state

https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/jd-client/src/lib/jdc_runtime.rs#L911

bit-aloo: I think there's a reason to keep Io in Running for the translator  and Pool, beyond reuse. Once Io is dropped, each channel end is exclusively owned by a component's tasks. If a component dies without signaling its token first (early return), the drop closes the channel instantly  its peer sees the recv error before the token flips. And some of those sites map to shutdown. So a race could lead to a full shutdown.

plebhash: my agent is leaning towards @nulllpc 's rationale here:

> I'd resolve in favor of keeping `Running` as-is (no `Io`):
> 
> - JDC carries `Io` for one concrete reason: `cleanup_for_fallback` reuses the TP channels from the running session ([jdc_runtime.rs#L911](https://github.com/stratum-mining/sv2-apps/blob/main/miner-apps/jd-client/src/lib/jdc_runtime.rs#L911)). The translator recreates every channel on fallback, and Pool's `Running` is already a unit struct — so the rule is "carry Io only if fallback reuses channels".
> - The race can't happen on orderly exits: components cancel the relevant token *before* closing (e.g. CM cancels inside `handle_error_action`, and `channel_manager_io.close()` runs after the loop), and peers check both tokens before classifying a recv error. Signal always precedes close.
> - Where closure *does* arrive before a token flips, that's the designed signal, not a race: `io_task` explicitly closes the inbound channel on TCP errors, and CM maps that recv error to `Fallback` — that's how upstream death triggers failover. Note `async_channel::close()` closes the channel globally, so parked ends in `Running` wouldn't suppress explicit closes anyway.
> - The only thing parked ends would suppress is drop-closure from a panicked component — and that's harm, not protection: upstream panic would no longer trigger fallback (nobody cancels the token, channel never closes), CM would queue shares into an unbounded channel nobody drains, and sv1-server death would leave a zombie proxy instead of exit(1) + supervisor restart. Also worth noting the sv1→cm sender lived inside `Sv1Server` before this PR too, so drop→shutdown on that channel is pre-existing behavior, not something this PR introduced.
>
> If there's a concrete sequence where a peer misclassifies a closure on a non-panic path, happy to reopen — but I couldn't construct one.

bit-aloo: oh! cool agent did the analysis (kinda happy :laughing:  that we don't have task lifetime fragmentation across the apps). Then maybe we can remove the non TP channels in JDC running IO state. Ofcourse, not on this PR.

bit-aloo (approved):
ACK

PR #648, opened 2026-07-27
Add display impl for errors variants in apps

bit-aloo:
I don't really like seeing phantom types in error. Its an internal detail, and shouldn't be exposed in logs. This PR adds impl display trait for error variants across apps.


par1ram (approved):
Reviewed the four error hierarchies and the logging call-site conversions; I found no blocking correctness issues. On the current merge with `main`, the 33 existing jd-client/translator tests pass, the pool/JDS crates build and test successfully, and four temporary contract tests confirmed that representative `Display` outputs preserve owner/action/cause context without leaking `PhantomData`.

Non-blocking: please consider retaining `downstream_id` as a dedicated tracing field in the `Action::Disconnect` branches even though it is included in `%e`; this keeps operational filtering intact. A small table-driven `Display` test would also make this new logging contract explicit, especially since pool and JDS currently have no unit tests.

PR #649, opened 2026-07-27
ci: add docs.rs compatibility checks

### LINKED ISSUE #524: add CI for `docs.rs`
in #519 it was discovered that `docs.rs` (which runs builds in very opinionated ways via [`docs_rs_builder`](https://github.com/rust-lang/docs.rs)) was failing for multiple crates

the root cause of those failures was documented via https://github.com/stratum-mining/sv2-apps/issues/519#issuecomment-4525746708

#521 is fixing the root cause, which should make the `docs.rs` builds succeed with the current state of the codebase

I tried a few different approaches for adding a CI that would emulate the internal limitations of `docs_rs_builder`, but the different trade-offs were non-trivial so #521 scope was limited to fixing the root cause of #519 only

this issue tracks adding CI that enforces sanity of `docs.rs` on the long term... ideally it should either:
- emulate internal limitations of `docs_rs_builder` (no-network sandboxed environment)
- literally use `docs_rs_builder`

but personally speaking (and speaking from my own experience, where it was not trivial to achieve either of those goals), I'm also ok with KISS approaches

par1ram:
## Summary

- add a GitHub Actions workflow that checks docs.rs compatibility for all seven published application crates
- prefetch dependencies, then run `cargo docs-rs --offline` inside an isolated Linux network namespace
- document the reusable `docs-rs-check.sh` helper

## Why

docs.rs builds crates without network access. A regular documentation build can therefore miss regressions where a dependency or build script attempts a download, as happened in #519.

This adds a focused regression check while keeping the implementation smaller than running the full docs.rs infrastructure.

Closes #524.

## Validation

- `cargo +nightly docs-rs --offline` passed for:
  - `stratum-apps`
  - `pool_sv2`
  - `jd_server_sv2`
  - `jd_client_sv2`
  - `translator_sv2`
  - `integration_tests_sv2`
  - `bitcoin_core_sv2`
- `actionlint .github/workflows/docs-rs.yaml`
- `shellcheck scripts/docs-rs-check.sh`
- `bash -n scripts/docs-rs-check.sh`
- `git diff --check`

The full Linux network-namespace path is exercised by the new workflow on this pull request.

### DIFF
--- a/scripts/docs-rs-check.sh
+++ b/scripts/docs-rs-check.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Run docs.rs-like documentation builds for one or more crates.
+# Arguments:
+#   paths to Cargo.toml files
+set -euo pipefail
+
+if [ "$#" -eq 0 ]; then
+  echo "Usage: $0 <manifest-path> [<manifest-path> ...]" >&2
+  exit 2
+fi
+
+for manifest_path in "$@"; do
+  if [ ! -f "$manifest_path" ]; then
+    echo "Manifest not found: $manifest_path" >&2
+    exit 1
+  fi
+
+  echo "Building docs for ${manifest_path}"
+  cargo +nightly docs-rs --manifest-path="$manifest_path" --offline
+done
--- a/.github/workflows/docs-rs.yaml
+++ b/.github/workflows/docs-rs.yaml
@@ -1,0 +1,33 @@
+name: Docs.rs
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  docs-rs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Build docs.rs sandbox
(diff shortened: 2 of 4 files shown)


scripts/docs-rs-check.sh
```diff
@@ -0,0 +1,38 @@
...
+  exit 1
+fi
+
+# Execute docs build inside a separate network namespace to emulate docs.rs
+# no-network constraints. Preserve Rust/cargo env so cargo remains available
+# inside the sudo context.
+echo "Building docs for ${manifest_path} in no-network sandbox"
+export MANIFEST_PATH="$manifest_path"
+export CARGO_BIN="$cargo_bin"
+sudo --preserve-env=MANIFEST_PATH,CARGO_BIN,PATH,CARGO_HOME,RUSTUP_HOME,HOME unshare --net --mount-proc bash -lc \
```
bit-aloo: We don't wanna use sudo on ci

plebhash:
@par1ram on this project we try to keep a reasonable level of git hygiene

if a reviewer asks you to change something, please don't add a new commit pivoting commit history to a new direction... simply rebase the original commit while atomically incorporating the review

in other words: https://github.com/stratum-mining/sv2-apps/pull/649/commits/ecd98fa70c672c686b204085ca31ac0179a26aec shouldn't exist as a standalone commit

PR #664, opened 2026-08-04
Use assoc error type from sv1 crate in translator

bit-aloo:
companion of: https://github.com/stratum-mining/stratum/pull/2268

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -1025,7 +1025,7 @@ checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
 [[package]]
 name = "binary_sv2"
 version = "7.0.0"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#30ca82ac20d7a543cc777be39dc2e19c621eff4b"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#42e774928681e3a81a3b88aab7a90f2f54a14d24"
 dependencies = [
  "buffer_sv2",
  "derive_codec_sv2",
@@ -1210,7 +1210,7 @@ dependencies = [
 [[package]]
 name = "buffer_sv2"
 version = "3.0.1"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#30ca82ac20d7a543cc777be39dc2e19c621eff4b"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#42e774928681e3a81a3b88aab7a90f2f54a14d24"
 dependencies = [
  "aes-gcm",
 ]
@@ -1347,7 +1347,7 @@ dependencies = [
 [[package]]
 name = "channels_sv2"
 version = "8.0.0"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#30ca82ac20d7a543cc777be39dc2e19c621eff4b"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#42e774928681e3a81a3b88aab7a90f2f54a14d24"
 dependencies = [
  "binary_sv2",
  "bitcoin",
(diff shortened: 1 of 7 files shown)


GitGab19 (commented):
Two minor nits found by the clanker.

---

## Two suggested patches

While reviewing this together with #530, I'd like to ask for two small additions here.
Rationale: `TproxyErrorKind` currently has two look-alike variants (`DownstreamNotFound`
holding a *request id*, `DownstreamNotPresent` holding a *downstream id*). #530's
skip-guards match on `DownstreamNotFound(_)` while this PR's handlers return
`DownstreamNotPresent` — after #530 rebases, the same "downstream is gone" condition
would flow under two different variants and half the guards would miss it. These two
patches remove the ambiguity at the source (and fix a pre-existing mislabeled log on
main).

### Patch 1 — rename the request-id variant, fix the misuse in `mod.rs`

`error.rs`:

```diff
     /// Downstream mapped to request id not found
-    DownstreamNotFound(u32),
+    RequestIdNotFound(u32),
```
```diff
-            DownstreamNotFound(request_id) => write!(
+            RequestIdNotFound(request_id) => write!(
                 f,
                 "Downstream id associated to request id: {request_id} not found"
             ),
```

`sv1/sv1_server/mod.rs` — `handle_open_channel_request` (this one is about a
*downstream id*, so it's currently mislabeled and logs "request id: N not found";
it moves to the correct variant and the cast disappears):

```diff
             return Err(TproxyError::disconnect(
-                TproxyErrorKind::DownstreamNotFound(downstream_id as u32),
+                TproxyErrorKind::DownstreamNotPresent(downstream_id),
                 downstream_id,
             ));
```

`sv1/sv1_server/mod.rs` — `OpenExtendedMiningChannelSuccess` (genuinely a
request-id lookup failure, takes the rename):

```diff
                 let Some((_, downstream_id)) = downstream_id else {
-                    return Err(TproxyError::log(TproxyErrorKind::DownstreamNotFound(
+                    return Err(TproxyError::log(TproxyErrorKind::RequestIdNotFound(
                         m.request_id,
                     )));
                 };
```

These are the only two uses of `DownstreamNotFound` in the repo
(`DownstreamNotFoundWithChannelId` is a different variant, untouched). After this,
`grep DownstreamNotFound` returns nothing — so #530's rebase can't compile until its
`with_registered_downstream` helper and guards are consciously switched to
`DownstreamNotPresent`, which guarantees the reconciliation happens.

### Patch 2 — centralize the downstream lookup in `downstream_message_handler.rs`

The `let Some(downstream) = self.downstreams.get(...) else { return Err(...) }` block
now repeats 11× in this file. One helper (outside the `hotpath::measure_all` impl)
makes `DownstreamNotPresent` originate from a single expression:

```rust
impl Sv1Server {
    /// Look up a registered downstream for an `IsServer` handler, mapping a
    /// missing entry to `TproxyErrorKind::DownstreamNotPresent`.
    fn registered_downstream(
        &self,
        downstream_id: DownstreamId,
    ) -> Result<Ref<'_, DownstreamId, Downstream>, TproxyErrorKind> {
        self.downstreams
            .get(&downstream_id)
            .ok_or(TproxyErrorKind::DownstreamNotPresent(downstream_id))
    }
}
```

Each of the 11 sites then collapses to:

```diff
-        let Some(downstream) = self.downstreams.get(&downstream_id) else {
-            return Err(TproxyErrorKind::DownstreamNotPresent(downstream_id));
-        };
+        let downstream = self.registered_downstream(downstream_id)?;
```

Notes: the returned `Ref` behaves identically to the inline `.get()` guard at every
site, including `authorize()` holding it across the nested `is_authorized()` call
(two read guards on the same shard, same as today); `?` converts via the reflexive
`From` since `Self::Error = TproxyErrorKind`. Needs `dashmap::mapref::one::Ref`,
`DownstreamId`, and `Downstream` added to the imports. #530's rebase will rework this
helper into its `SharedMap` closure form — that conflict is intentional, it forces the
two lookup paths to merge into one.

bit-aloo:
I am taking up the first suggestion. For the second, we are only using the map’s getter APIs, so I wouldn’t really consider it duplication that needs to be backed by a helper.

GitGab19 (approved):
ACK, way better now, thanks for pushing this changes and properly unblock #530

Issue #645, opened 2026-07-23
JDS needs tighter validation against `DeclareMiningJob.coinbase_tx_prefix`


plebhash:
JDS's `DeclaredCustomJob::get_coinbase_tx` is a bit too naive:

https://github.com/stratum-mining/sv2-apps/blob/fff6abcaf02338035fadeb2a4ebf01c63a31118d/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L172-L209

if a `DeclareMiningJob` comes with `coinbase_tx_prefix` that don't satisfy some naive assumptions, that can lead to panics

more specifically:
- if `coinbase_tx_prefix` is smaller than the expected 43 bytes, `&declared_coinbase_tx_prefix[COINBASE_PREFIX_LEN..]` panics with index-out-of-bounds 
- adjacent arithmetic (`prefix.len() - script_sig_offset`, `script_sig_size - script_sig_bytes_in_prefix`) can also underflow depending on the contents of `DeclareMiningJob`
- if `scriptSig length` is declared as a sufficiently large value (which would actually break Bitcoin consensus), it could lead to unbounded memory allocation

in summary, JDS should defensively check against all these edge cases before allocating memory, and simply reject the `DeclareMiningJob` in case of invalid values

plebhash:
expanding scope a bit, based on this comment from @bit-aloo https://github.com/stratum-mining/sv2-apps/pull/677#pullrequestreview-4895441159

two points beyond the parse-safety fixes already described here:

**1. ordering.** `get_coinbase_tx` reconstruction runs before the structural guards, not after. the `input.len() != 1` check sits at [`bitcoin_core_ipc.rs:546`](https://github.com/stratum-mining/sv2-apps/blob/b26cf780046c2747fed64905114b80ce0464a6c6/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L546-L550), but the reconstruction it should gate happens first at [:531](https://github.com/stratum-mining/sv2-apps/blob/b26cf780046c2747fed64905114b80ce0464a6c6/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L531). beyond making the parse panic/alloc-safe, the cheap O(1) structural rejections (input count, a sane size cap) should run before the byte reconstruction, so a malformed job doesn't pay for reconstruction first.

[...]

PR #677, opened 2026-08-04
`bitcoin-core-sv2`: harden JDP `handle_declare_mining_job` against untrusted `DeclareMiningJob` inputs

### LINKED ISSUE #675: `bitcoin-core-sv2`: need to harden JDP `handle_declare_mining_job` against untrusted `DeclareMiningJob` inputs
`handle_declare_mining_job` (in both version trees `v30x` and `v31x`) blindly trusts its `DeclareMiningJob` input, relying on guards that only exist in the in-tree `jd-server-sv2` caller — but `bitcoin_core_sv2` is a public library whose `JdRequest::DeclareMiningJob` API accepts arbitrary input.

## 1. Malformed coinbase takes down the JDP request loop

`jd-server-sv2` rejects a declared coinbase that does not carry exactly one input before it ever reaches the IPC layer:

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L544-L550

the handler itself has no such check, and a zero-input coinbase breaks it twice.

### 1a. Panic on zero-input coinbase

logging `coinbase_tx.input[0].script_sig` via direct indexing inside a `tracing::debug!`:

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs#L46-L49

A zero-input coinbase transaction panics the handler when DEBUG logging is active, tearing down the JDP request loop and halting all job-declaration validation.

Line 53 just below already uses `.first()` with an explicit "avoid panics" fallback — line 48 is the inconsistent access.

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs#L51-L63

Fix: use `.first()` at line 48 as well.

Note this is not made redundant by the guard in 1b: the `debug!` fires before the handler has read the validation context it needs to answer with an error, so the guard cannot be hoisted above it.

### 1b. Same teardown via block serialization, with no logging required

`.first()` stops the panic but not the outage. The zero-input coinbase still reaches block assembly:

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs#L136-L141

rust-bitcoin serializes a transaction with no inputs in BIP141 form on purpose, to dodge the ambiguity between a segwit marker and a legacy input count. That encoding advertises a witness flag with no witness data behind it, which BIP141 forbids, so Bitcoin Core's deserializer rejects it outright: `checkBlock` comes back as a capnp remote exception — `std::exception: Superfluous witness record: iostream error` — which lands here:

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs#L172-L186

so the client gets `internal-error` and the handler then cancels the whole IPC connection. Every subsequent request fails on a closed channel. Unlike 1a this needs no DEBUG logging, so fixing only 1a converts a panic into a quieter outage with the same blast radius.

Fix: reject a coinbase that does not carry exactly one input before assembling the block, answering with `ERROR_CODE_DECLARE_MINING_JOB_INVALID_COINBASE_TX_INPUT`, mirroring the guard `jd-server-sv2` already applies. Note this reclassifies multi-input coinbases, which deserialize fine and today reach `checkBlock` and are rejected there, to the more precise error code — no change for the in-tree caller, which already filters them out.

## 2. Rejected declarations retain client-supplied transactions

we're inserting every client-supplied `missing_txs` entry into the process-wide `MempoolMirror` *before* the declaration passes `checkBlock`:

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs#L67

If validation fails, the handler returns `JdResponse::Error` without removing the inserted transactions.

`MempoolMirror` only clears `txdata` on prev-hash changes, so repeated rejected declarations with unique invalid transactions grow shared memory until the next tip change — a DoS vector against all JDP users.

Fix: don't insert `missing_txs` into the mirror before validation. Stage them in a local map, verify `wtxid_list` and assemble the block from the mirror plus the staged overlay, and only commit them via `add_transactions` after `checkBlock` succeeds.

A single commit point on the success path removes the need for rollback on every error branch, and guarantees unvalidated transactions never enter shared state — including on the `MissingTransactions` early-return, which today also retains partially-supplied transactions.

Clients already resend the full tx set on retry (`jd-server-sv2` forwards `ProvideMissingTransactionsSuccess` transactions into the re-declaration), so the round-trip is unaffected. Only transactions actually named in `wtxid_list` should be committed — staged entries the client sent but did not declare were never inside the block `checkBlock` validated.

## Notes

- All fixes must be applied to **both** the `v30x` and `v31x` trees — the flagged code is identical in each.
- A regression test in `integration-tests/tests/bitcoin_core_ipc_jdp_io.rs` can cover the retention fix directly: declare a job with a zero-output tx, receive `Error`, then declare again with the same wtxid and an empty tx list, asserting the wtxid is reported missing rather than served from the mirror. Note the failure path force-refreshes the mirror, but `MempoolMirror::update` only clears `txdata` on a prev-hash change, so the rejected tx really does survive today.
- The zero-input coinbase scenario belongs in the same file, but only once 1b is fixed: without the guard it cancels the IPC connection and takes the other scenarios in that shared runtime down with it. Running it *first* makes every scenario after it double as proof that the rejection left the connection alive.
- 1a alone is not observable in CI — `start_tracing()` defaults to INFO and `tracing::debug!` does not evaluate its arguments below the enabled level, so only the 1b guard is testable.

plebhash:
close #675 
close https://github.com/project-loupe/audit-sv2-apps/issues/61
close https://github.com/project-loupe/audit-sv2-apps/issues/30

### DIFF
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs
@@ -14,8 +14,10 @@ use stratum_core::{
         hashes::Hash,
     },
     job_declaration_sv2::{
-        ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR, ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB,
-        ERROR_CODE_DECLARE_MINING_JOB_STALE_CHAIN_TIP, PushSolutionOwned,
+        ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR,
+        ERROR_CODE_DECLARE_MINING_JOB_INVALID_COINBASE_TX_INPUT,
+        ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB, ERROR_CODE_DECLARE_MINING_JOB_STALE_CHAIN_TIP,
+        PushSolutionOwned,
     },
 };
 use tokio::sync::oneshot;
@@ -24,7 +26,8 @@ use tracing::{debug, error, info, warn};
 impl BitcoinCoreSv2JDP {
     /// Validates a declared mining job by checking transaction availability and block structure.
     ///
-    /// Adds missing transactions to the mempool mirror, verifies all transactions are available,
+    /// Rejects a coinbase that does not carry exactly one input, adds missing transactions to the
+    /// mempool mirror, verifies all transactions are available,
     /// assembles a test block, and uses Bitcoin Core's `checkBlock` to validate the block
     /// structure. Returns success with current template parameters or an error if validation
     /// fails.
@@ -45,7 +48,7 @@ impl BitcoinCoreSv2JDP {
         );
         debug!(
             "Declared coinbase scriptSig: {:?}",
-            coinbase_tx.input[0].script_sig
+            coinbase_tx.input.first().map(|input| &input.script_sig)
         );
 
         let declared_bip34_height = coinbase_tx
@@ -85,6 +88,24 @@ impl BitcoinCoreSv2JDP {
             let initial_bip34_height = mempool_mirror
                 .get_current_bip34_height()
                 .expect("current_bip34_height must be set");
+
+            // A coinbase must carry exactly one input. Reject anything else before assembling
+            // the block: a zero-input coinbase re-serializes into bytes Bitcoin Core's
+            // deserializer reads as a SegWit marker ("Superfluous witness record"), which comes
+            // back as a capnp remote exception and tears down the whole IPC connection.
+            if coinbase_tx.input.len() != 1 {
+                warn!(
+                    "Rejecting DeclareMiningJob: declared coinbase has {} inputs, expected exactly 1",
+                    coinbase_tx.input.len()
+                );
+                // deliberately ignore potential errors
+                // we don't care if the receiver dropped the channel
+                let _ = response_tx.send(JdResponse::Error {
+                    error_code: ERROR_CODE_DECLARE_MINING_JOB_INVALID_COINBASE_TX_INPUT,
+                    validation_context: initial_validation_context,
+                });
+                return;
+            }
 
             // Now verify that all wtxids from the declared job are available
             let missing_wtxids = mempool_mirror.verify(&wtxid_list);
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/job_declaration_protocol/handlers.rs
@@ -6,6 +6,7 @@ use crate::{
         BitcoinCoreSv2JDP, mempool::decode_bip34_height_from_coinbase_script_sig,
     },
 };
+use std::collections::HashMap;
 use stratum_core::{
     bitcoin::{
         Block, Transaction, TxMerkleNode, Txid, Wtxid,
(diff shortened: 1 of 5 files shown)


bit-aloo (commented):
Clankers review:

> Committing after the await can reintroduce stale txs
The old code inserted into the mirror before checkBlock; the new code inserts after it. The template monitor runs on the same LocalSet and MempoolMirror::update clears txdata whenever prev_hash changes ([mempool.rs:31-33](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/bitcoin-core-sv2/src/unix_capnp/v31x/job_declaration_protocol/mempool.rs#L31-L33)). If the tip moves while check_block_request.send().promise is awaited, the commit at line 286 writes transactions validated against the old tip into a freshly-cleared mirror, where they persist until the next tip change.

> Biggest remaining hole in the threat model is upstream of this diff
The branch hardens the IPC layer, but DeclaredCustomJob::get_coinbase_tx() in [bitcoin_core_ipc.rs:169-214](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L169-L214) parses fully client-controlled bytes with no bounds checks, and it runs before the request ever reaches the code you hardened:

> [:181](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L181) &declared_coinbase_tx_prefix[43..] — panics outright when coinbase_tx_prefix is shorter than 43 bytes. It's a B0_64K, so an empty prefix is a legal message.
[:197](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L197) and [:200](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L200) — two unchecked usize subtractions that underflow on a short prefix or an understated script_sig_size.
[:205](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs#L205) vec![0; full_extranonce_size] — the size derives from a client-supplied VarInt with no cap, so a large value is an allocation abort.
I checked [job_declaration_message_handler.rs:108-183](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/pool-apps/jd-server/src/lib/job_declarator/job_declaration_message_handler.rs#L108-L183): the only gates before this are token parsing and token-allocation, so any downstream that completed setup reaches it. Pre-existing and outside your diff — but it's the same "untrusted DeclareMiningJob" surface, and it fires earlier than everything you fixed.

plebhash:
> Clankers review:
> 
> > Committing after the await can reintroduce stale txs
> > The old code inserted into the mirror before checkBlock; the new code inserts after it. The template monitor runs on the same LocalSet and MempoolMirror::update clears txdata whenever prev_hash changes (mempool.rs:31-33). If the tip moves while check_block_request.send().promise is awaited, the commit at line 286 writes transactions validated against the old tip into a freshly-cleared mirror, where they persist until the next tip change.
> 
> > Biggest remaining hole in the threat model is upstream of this diff
> > The branch hardens the IPC layer, but DeclaredCustomJob::get_coinbase_tx() in bitcoin_core_ipc.rs:169-214 parses fully client-controlled bytes with no bounds checks, and it runs before the request ever reaches the code you hardened:
> 
> > :181 &declared_coinbase_tx_prefix[43..] — panics outright when coinbase_tx_prefix is shorter than 43 bytes. It's a B0_64K, so an empty prefix is a legal message.
> > :197 and :200 — two unchecked usize subtractions that underflow on a short prefix or an understated script_sig_size.
> > :205 vec![0; full_extranonce_size] — the size derives from a client-supplied VarInt with no cap, so a large value is an allocation abort.
> > I checked job_declaration_message_handler.rs:108-183: the only gates before this are token parsing and token-allocation, so any downstream that completed setup reaches it. Pre-existing and outside your diff — but it's the same "untrusted DeclareMiningJob" surface, and it fires earlier than everything you fixed.

thanks

the point about TOCTOU was solved:

> We moved the mirror commit to after the post-`checkBlock` read of the current context, and gated it on `latest_validation_context.prev_hash == initial_validation_context.prev_hash`. If the tip moved while `checkBlock` was in flight, that check fails and we drop the staged transactions instead of committing them into the freshly-cleared mirror — the client resends them on retry. So transactions only enter shared state when they were validated against the tip that's still current.

the point about JDS is closely related to the scope already tracked via #645, so I left this comment to make sure we also cover it: https://github.com/stratum-mining/sv2-apps/issues/645#issuecomment-5258914571

bit-aloo (approved):
ACK

PR #678, opened 2026-08-04
`bitcoin-core-sv2`: fix `u32` overflow in `coinbase_weight`

### LINKED ISSUE #676: `bitcoin-core-sv2`: need to fix `u32` overflow in `coinbase_weight` wraps before cast to `u64`
`bootstrap_template_ipc_client_from_coinbase_output_constraints` (in both version trees `v30x` and `v31x`) computes the `block_reserved_weight` sent to Bitcoin Core over IPC as:

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs#L530-L534

https://github.com/stratum-mining/sv2-apps/blob/c990ef7ea67cc18ea6d40ab1a9f976afe2d9365d/bitcoin-core-sv2/src/unix_capnp/v31x/template_distribution_protocol/mod.rs#L547-L551

`coinbase_output_max_additional_size` is a `u32` carried over the SV2 wire in `CoinbaseOutputConstraints`, and `WEIGHT_FACTOR` is also `u32`. The multiplication is therefore evaluated in `u32` and only *afterwards* widened to `u64`. For any input `>= 0x4000_0001` (~1 GiB) the multiplication overflows: in release builds it wraps silently, in debug builds it panics with "attempt to multiply with overflow".

Impact in release builds: the computed `coinbase_weight` is a wrapped value unrelated to the requested size. E.g. `0x4000_0000` wraps to `0` and gets clamped to `MIN_BLOCK_RESERVED_WEIGHT`, so Bitcoin Core is told to reserve only the minimum coinbase weight while the peer asked for far more. Either way the template is built against a mismatched constraint, leading to mining failures for that session.

The in-tree callers derive `coinbase_output_max_additional_size` from locally-trusted state, but `bitcoin_core_sv2` is a public library and the value is peer-controlled at the protocol level, so the library shouldn't rely on caller-side sanity.

Fix: change `WEIGHT_FACTOR` in `unix_capnp/mod.rs` to `u64`, matching `MIN_BLOCK_RESERVED_WEIGHT` which is already `u64`

Notes

- The fix must be applied to both the `v30x` and `v31x` trees — the flagged code is identical in each.

plebhash:
close #676
close https://github.com/project-loupe/audit-sv2-apps/issues/16

### DIFF
--- a/bitcoin-core-sv2/src/unix_capnp/mod.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/mod.rs
@@ -14,7 +14,7 @@ pub mod v31x;
 const MIN_BLOCK_RESERVED_WEIGHT: u64 = 2000;
 
 /// BIP141 weight factor (witness scale factor), used to convert vsize to weight units.
-const WEIGHT_FACTOR: u32 = 4;
+const WEIGHT_FACTOR: u64 = 4;
 
 /// Grace period before stale template data is retired after a chain tip change, in seconds.
 ///
--- a/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
+++ b/bitcoin-core-sv2/src/unix_capnp/v30x/template_distribution_protocol/mod.rs
@@ -527,8 +527,8 @@ impl BitcoinCoreSv2TDP {
                 e
             })?;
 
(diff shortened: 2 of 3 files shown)


bit-aloo (approved):
ACK

PR #633, opened 2026-07-20
Refactor(jd-client): replace monolithic start loop with JdcRuntime typestate machine

nulllpc:
Refactors the monolithic `start` function inside the Pool (`miner-apps/jd-client`) into a structured, typestate-based state machine (`JdcRuntime`) to manage bootstrap, component lifecycles, and graceful shutdown.

This PR also properly handles errors instead of calling `unwrap` or `expect`

Fixes https://github.com/stratum-mining/sv2-apps/issues/526

PR #685, opened 2026-08-05
`PoolRuntime` startup and shutdown gaps, aligned with #633

plebhash:
follow-up to #582 

aims to align patterns and patch gaps found on #633

### DIFF
--- a/pool-apps/pool/src/lib/channel_manager/mod.rs
+++ b/pool-apps/pool/src/lib/channel_manager/mod.rs
@@ -270,34 +270,9 @@ impl ChannelManager {
         // todo: let start_downstream_server accept Arc, instead of clone.
         let this = Arc::new(self);
 
-        // Wait for initial template and prevhash before accepting connections
-        loop {
-            let has_required_data = this
-                .last_future_template
-                .with(|template| template.is_some())
-                .map_err(PoolError::shutdown)?
-                && this
-                    .last_new_prev_hash
-                    .with(|prevhash| prevhash.is_some())
-                    .map_err(PoolError::shutdown)?;
-
-            if has_required_data {
-                info!("Required template data received, ready to accept connections");
-                break;
-            }
-
-            warn!("Waiting for initial template and prevhash from Template Provider...");
-            warn!("Is the Bitcoin node undergoing IBD?");
-            select! {
-                biased;
-                _ = cancellation_token.cancelled() => {
-                    info!("Channel Manager: received shutdown while waiting for templates");
-                    return Ok(());
-                }
-                _ = tokio::time::sleep(std::time::Duration::from_secs(1)) => {}
-            }
-        }
-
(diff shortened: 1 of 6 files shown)


nulllpc:
LGTM

nulllpc (approved):

pool-apps/pool/src/lib/mod.rs
```diff
@@ -71,6 +77,16 @@ impl PoolSv2 {
         Ok(())
     }
 
+    /// Marks the pool as stopped and releases anyone blocked on [`PoolSv2::shutdown`].
+    ///
+    /// Called by [`PoolRuntime::shutdown`] once teardown completes, and directly by
+    /// [`PoolSv2::start`] on the failure path where no runtime exists to tear down.
+    fn mark_stopped(&self) {
+        self.is_alive.store(false, Ordering::Release);
```
bit-aloo, line 85: considering this storee is `release`, can we make corresponding load `acquire`?

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -310,6 +285,41 @@ impl ChannelManager {
         let task_manager_clone = task_manager.clone();
         let cancellation_token_clone = cancellation_token.clone();
         task_manager.spawn(async move {
+            // Wait for initial template and prevhash before accepting connections
+            let ready = loop {
```
bit-aloo, line 289: This has a behavior change now: we start accepting connections even before we have templates and prevhash ready. This might cause a delay in the Noise handshake or even a disconnect. But I guess that's okay, considering that otherwise the user would get an immediate failure, whereas now, if anything, the failure would just be delayed.

bit-aloo: Just a thought.

bit-aloo (commented):
Looks good. Just a few nits. This handles all the bootstrap-related errors perfectly. But now I’m a bit concerned about errors in the running state. Currently, we return `Ok` even if something goes wrong internally. We might need to change that behavior, but it’s not a concern for this PR.

pool-apps/pool/src/lib/channel_manager/mod.rs
plebhash: yeah I did consider this

tbh I think the previous behavior had a bit more global cohesion, but I'm also ok with what we have here

pool-apps/pool/src/lib/mod.rs
plebhash: done, but also doing the same on tProxy and JDC for consistency, which is already expanding the scope of this PR beyond what was originally scoped

if further changes are to be requested, please consider creating a follow-up issue

470df94ee1b84e99718753fa79e836898ef39477
33d09e6ab4ad2b128035b56fea51a419d87ce77e

bit-aloo (approved):
ACK

PR #689, opened 2026-08-05
fix bad SIGINT patterns

### LINKED ISSUE #688: `noise_connection` installs a process-wide SIGINT handler, breaking Ctrl+C for every consumer
`noise_connection::Connection::{spawn_reader,spawn_writer}` both `select!` on `tokio::signal::ctrl_c()` as their shutdown signal:

https://github.com/stratum-mining/sv2-apps/blob/2089ba75507c5e6c5348f0f32e49f6cd1fc8de6a/stratum-apps/src/network_helpers/noise_connection.rs#L67-L100

https://github.com/stratum-mining/sv2-apps/blob/2089ba75507c5e6c5348f0f32e49f6cd1fc8de6a/stratum-apps/src/network_helpers/noise_connection.rs#L102-L140

The first time either future is polled, tokio registers a process-global SIGINT handler. From that moment on, SIGINT no longer terminates the process — it only completes those futures (the reader/writer tasks exit, the process keeps running).

This means any binary that opens even one noise connection silently loses default Ctrl+C behavior.

If the process later gets stuck for an unrelated reason, it cannot be interrupted (observed while testing `interoperability-tests-sv2` against an unreachable endpoint: repeated Ctrl+C, and even an explicit `kill -INT <pid>`, had no effect). See https://github.com/stratum-mining/interoperability-tests-sv2/pull/2#issuecomment-5185463970

## **Reproduce:**
1. Establish one noise `Connection` (any initiator/responder pair).
2. Block the main task on anything (e.g. a long `sleep`).
3. Press Ctrl+C → process does not terminate.

## **Suggested fix:**
Signal handling is an application-level decision; a transport library shouldn't claim process signals. Replace the `ctrl_c()` arms with an internal shutdown mechanism: a `CancellationToken` passed into `Connection::new` 

Binaries that want graceful shutdown on SIGINT can hook that up themselves.

plebhash:
close #688

### DIFF
--- a/integration-tests/lib/mining_device/mod.rs
+++ b/integration-tests/lib/mining_device/mod.rs
@@ -37,6 +37,7 @@ use stratum_apps::{
     sync::SharedLock,
 };
 use tokio::net::TcpStream;
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info};
 
 // Fast SHA256d midstate hasher
@@ -134,9 +135,13 @@ pub async fn connect(
     info!("Pool tcp connection established at {}", address);
     let address = socket.peer_addr().unwrap();
     let initiator = Initiator::new(pub_key.map(|e| e.0));
-    let (receiver, sender) = Connection::new(socket, HandshakeRole::Initiator(initiator))
-        .await
-        .unwrap();
+    let (receiver, sender) = Connection::new(
+        socket,
+        HandshakeRole::Initiator(initiator),
+        CancellationToken::new(),
+    )
+    .await
+    .unwrap();
     info!("Pool noise connection established at {}", address);
     Device::start(
         receiver,
--- a/integration-tests/lib/sv1_sniffer.rs
+++ b/integration-tests/lib/sv1_sniffer.rs
@@ -77,7 +77,6 @@ impl SnifferSV1 {
             let downstream_to_sniffer_connection =
                 ConnectionSV1::new(downstream_stream, CancellationToken::new()).await;
             select! {
-                _ = tokio::signal::ctrl_c() => { },
                 _ = Self::recv_from_down_send_to_up_sv1(
                     downstream_to_sniffer_connection.receiver(),
                     sniffer_to_upstream_connection.sender(),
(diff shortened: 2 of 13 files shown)


plebhash:
the scope of this PR (more specifically the top commit, currently hashed as https://github.com/stratum-mining/sv2-apps/pull/689/commits/fc3cfc01fa6bff54889393faff9010b73a874ba2) overlaps with:
- #633 
- #685
- #644

but let's see what gets merged first, and then rebase accordingly... no strong opinions on the specific order

bit-aloo:
Even with the changes here, a second `ctrl + C` will get absorbed, if the any such task are blocked?

integration-tests/lib/mining_device/mod.rs
```diff
@@ -134,9 +135,13 @@ pub async fn connect(
     info!("Pool tcp connection established at {}", address);
     let address = socket.peer_addr().unwrap();
     let initiator = Initiator::new(pub_key.map(|e| e.0));
-    let (receiver, sender) = Connection::new(socket, HandshakeRole::Initiator(initiator))
-        .await
-        .unwrap();
+    let (receiver, sender) = Connection::new(
+        socket,
+        HandshakeRole::Initiator(initiator),
+        CancellationToken::new(),
```
bit-aloo, line 141: Shouldn't we be storing this cancellation token somewhere?

integration-tests/lib/sv1_sniffer.rs
```diff
@@ -98,67 +97,54 @@ impl SnifferSV1 {
...
-                        panic!(
-                            "Timeout: keepalive mining.notify (job_id containing '#') not found"
-                        );
-                    } else {
-                        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
-                        continue;
-                    }
+        loop {
+            let has_notify = match direction {
+                MessageDirection::ToUpstream => {
```
bit-aloo, line 131: Should we add a cancellation lookup in these select blocks?

integration-tests/lib/sv1_sniffer.rs
```diff
@@ -98,67 +97,54 @@ impl SnifferSV1 {
...
-                            }
-                        }
-                        MessageDirection::ToDownstream => {
-                            if self.messages_from_upstream.has_message(message).await {
-                                break;
-                            }
-                        }
+        loop {
+            match direction {
+                MessageDirection::ToUpstream => {
```
bit-aloo, line 102: Should we add cancellation lookup here?

integration-tests/lib/utils.rs
```diff
@@ -90,9 +95,13 @@ pub async fn create_upstream(
     stream: tokio::net::TcpStream,
 ) -> Option<(Receiver<MessageFrame>, Sender<MessageFrame>)> {
     let initiator = Initiator::without_pk().expect("This fn call can not fail");
-    Connection::new::<AnyMessageOwned>(stream, HandshakeRole::Initiator(initiator))
-        .await
-        .ok()
+    Connection::new::<AnyMessageOwned>(
+        stream,
+        HandshakeRole::Initiator(initiator),
+        CancellationToken::new(),
```
bit-aloo, line 101: We should do something with this? if we are passing it?

stratum-apps/src/lib.rs
bit-aloo, line 52: Now, network helpers depend on tokio_util? we can feature gate it under tokio-util feature. Though I now don't remember why we feature gated tokio-util.

bit-aloo (commented):
Changes makes sense. Some suggestions and questions.

integration-tests/lib/mining_device/mod.rs
plebhash: clanker answer, which I agree with:

> Not needed here — the mining device is test-harness code with no shutdown API; it's spawned fire-and-forget and torn down when the test's runtime drops (which aborts the reader/writer tasks). The connection also exits on its own via socket EOF / channel closure. The token param is for the real roles to plumb in their app-level cancellation token; a fresh dummy token is the same pattern `create_downstream`/`create_upstream` and the SV1 sniffer already use.

integration-tests/lib/sv1_sniffer.rs
plebhash: > No — these wait helpers are awaited inline by the test body (not detached tasks) and are already bounded by the 60s timeout panic, and this matches the SV2 `Sniffer::wait_for_message_type`, which has always been a plain poll loop. There's also no token in `SnifferSV1` to look up, and nothing that would cancel it. The old `ctrl_c` arm was actually harmful: Ctrl+C made the select complete so the wait *returned as if the message had arrived*, and it registered a process-wide signal handler from test-lib code — the pattern this PR removes.

integration-tests/lib/sv1_sniffer.rs
plebhash: > Same answer as above — after removing the `ctrl_c` arm the select had only one branch left, so it's now a plain loop; there's no token in `SnifferSV1` to select on and nothing that would cancel it. The timeout panic bounds the wait.

integration-tests/lib/utils.rs
plebhash: same as `mining_device`

stratum-apps/src/lib.rs
plebhash: > It's already gated that way — `tokio-util` is optional, and `network`/`fallback-coordinator`/`sv1` each enable it (`network = ["tokio-util", "core"]`); the new re-export is behind `#[cfg(feature = "tokio-util")]`. The dep also isn't new to network helpers: `sv1_connection.rs` already used `tokio_util` codec + `CancellationToken` before this PR. The reason for gating is the role bundles that exclude networking (`jd_server`, `mining_device`) — keeps tokio-util out of those builds. The re-export exists so apps can name `CancellationToken` via `stratum_apps::tokio_util` instead of depending on tokio-util directly and risking version skew against the type `Connection::new` expects.

bit-aloo:
> Even with the changes here, a second `ctrl + C` will get absorbed, if the any such task are blocked?

Curious to hear about this?

bit-aloo (approved):
ACK

PR #691, opened 2026-08-06
fix: remove useless conversion after `stratum-core` bump

plebhash:
https://github.com/stratum-mining/stratum/pull/2285 should have had this as a companion

I merged https://github.com/stratum-mining/sv2-apps/pull/690 but started getting this kind of failure:
https://github.com/stratum-mining/sv2-apps/actions/runs/31106372569/job/92636877422

### DIFF
--- a/miner-apps/translator/src/lib/utils.rs
+++ b/miner-apps/translator/src/lib/utils.rs
@@ -93,9 +93,7 @@ pub fn validate_sv1_share(
         full_extranonce.as_ref(),
         job.merkle_branch.as_ref(),
     )
-    .ok_or(TproxyErrorKind::InvalidMerkleRoot)?
-    .try_into()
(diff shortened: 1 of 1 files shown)


bit-aloo (approved):
Ack. But why the CI didn't fail

bit-aloo:
> Ack. But why the CI didn't fail

That's a clippy failure...

plebhash:
> > Ack. But why the CI didn't fail
> 
> That's a clippy failure...

https://github.com/stratum-mining/stratum/pull/2285 CI didn't fail because it only runs cross-repo integration tests, no clippy

#690 would have failed but I bypassed it, as I usually do for the bot-generated `stratum-core`-sync PRs, under the premise that if anything is going to wrong, we're going to be alerted post-merge anyways (which is what happened here)

but that's just me trying to expedite things a bit, while taking calculated risks

happy to stop doing this if people prefer to stick with a more conservative workflow

PR #698, opened 2026-08-07
speed up integration tests

GitGab19:
This is a test PR with some improvements on the IT in order to speed up tests execution and improve flakiness.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
+          # Hash every Cargo workspace without registering the shared target more than once.
+          key: cargo-inputs-${{ hashFiles('**/Cargo.toml', '**/Cargo.lock') }}
           workspaces: integration-tests -> ../target
 
       - name: Clippy check
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,6 +15,8 @@ jobs:
   clippy:
     name: Clippy Check
     runs-on: ubuntu-latest
+    env:
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -26,6 +28,12 @@ jobs:
         uses: dtolnay/rust-toolchain@1.88
         with:
           components: clippy
+
+      - name: Cache Rust build
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: "false"
+          workspaces: integration-tests -> ../target
 
       - name: Clippy check
         run: |
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -34,6 +34,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
+          # Hash every Cargo workspace without registering the shared target more than once.
+          key: cargo-inputs-${{ hashFiles('**/Cargo.toml', '**/Cargo.lock') }}
           workspaces: integration-tests -> ../target
 
       - name: Install capnp dependencies (Ubuntu)
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -48,6 +48,9 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         run: cargo binstall --no-confirm --disable-telemetry --disable-strategies quick-install --locked cargo-nextest@0.9.100
 
+# Failure triage: re-run a single test with captured logs shown live:
+      #   RUST_LOG=debug cargo nextest run --manifest-path=integration-tests/Cargo.toml \
+      #     -E 'test(<name>)' --no-capture
       - name: Integration Tests
         run: |
-         RUST_BACKTRACE=1 RUST_LOG=debug cargo nextest run --manifest-path=integration-tests/Cargo.toml --nocapture
+          RUST_BACKTRACE=1 RUST_LOG=debug cargo nextest run --manifest-path=integration-tests/Cargo.toml
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -48,7 +48,7 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         run: cargo binstall --no-confirm --disable-telemetry --disable-strategies quick-install --locked cargo-nextest@0.9.100
 
-# Failure triage: re-run a single test with captured logs shown live:
+      # Failure triage: re-run a single test with captured logs shown live:
       #   RUST_LOG=debug cargo nextest run --manifest-path=integration-tests/Cargo.toml \
       #     -E 'test(<name>)' --no-capture
       - name: Integration Tests
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -8,6 +8,9 @@ on:
 
 name: Integration Tests
 
+env:
+  CARGO_TARGET_DIR: ${{ github.workspace }}/target
+
 jobs:
   ci:
     runs-on: ${{ matrix.os }}
@@ -26,6 +29,12 @@ jobs:
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@1.88
+
+      - name: Cache Rust build
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: "false"
+          workspaces: integration-tests -> ../target
 
       - name: Install capnp dependencies (Ubuntu)
         if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/lockfiles.yaml
+++ b/.github/workflows/lockfiles.yaml
@@ -11,6 +11,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
 
     steps:
       - uses: actions/checkout@v4
@@ -20,6 +22,12 @@ jobs:
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache Rust build
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: "false"
+          workspaces: integration-tests -> ../target
 
       - name: Build with locked dependencies
         run: |
--- a/.github/workflows/lockfiles.yaml
+++ b/.github/workflows/lockfiles.yaml
@@ -27,6 +27,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
+          # Hash every Cargo workspace without registering the shared target more than once.
+          key: cargo-inputs-${{ hashFiles('**/Cargo.toml', '**/Cargo.lock') }}
           workspaces: integration-tests -> ../target
 
       - name: Build with locked dependencies
--- a/.github/workflows/msrv.yaml
+++ b/.github/workflows/msrv.yaml
@@ -22,6 +22,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
+          # Hash every Cargo workspace without registering the shared target more than once.
+          key: cargo-inputs-${{ hashFiles('**/Cargo.toml', '**/Cargo.lock') }}
           workspaces: integration-tests -> ../target
 
       - name: Install capnp dependencies
--- a/.github/workflows/msrv.yaml
+++ b/.github/workflows/msrv.yaml
@@ -13,12 +13,16 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CARGO_INCREMENTAL: 0
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
       RUSTFLAGS: -C debuginfo=0
 
     steps:
       - uses: actions/checkout@v4
-      - uses: Swatinem/rust-cache@v1.2.0
       - uses: dtolnay/rust-toolchain@1.88
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: "false"
+          workspaces: integration-tests -> ../target
 
       - name: Install capnp dependencies
         run: sudo apt-get install -y capnproto libcapnp-dev
--- a/.github/workflows/semver-check.yaml
+++ b/.github/workflows/semver-check.yaml
@@ -28,6 +28,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
+          # Hash every Cargo workspace without registering the shared target more than once.
+          key: cargo-inputs-${{ hashFiles('**/Cargo.toml', '**/Cargo.lock') }}
           workspaces: integration-tests -> ../target
 
       - name: Install dependencies
--- a/.github/workflows/semver-check.yaml
+++ b/.github/workflows/semver-check.yaml
@@ -11,6 +11,8 @@ on:
 jobs:
   semver-check:
     runs-on: ubuntu-latest
+    env:
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
 
     steps:
       - name: Checkout repository
@@ -22,21 +24,11 @@ jobs:
       - name: Install capnp dependencies
         run: sudo apt-get install -y capnproto libcapnp-dev
 
-      - name: Cache Cargo registry
-        uses: actions/cache@v4
+      - name: Cache Rust build
+        uses: Swatinem/rust-cache@v2
         with:
-          path: ~/.cargo/registry
-          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-registry-
-
-      - name: Cache Cargo index
-        uses: actions/cache@v4
-        with:
-          path: ~/.cargo/git
-          key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-index-
+          cache-bin: "false"
+          workspaces: integration-tests -> ../target
 
       - name: Install dependencies
         run: sudo apt-get update && sudo apt-get install -y cmake
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_TARGET_DIR: ${{ github.workspace }}/target
 
 jobs:
   test:
@@ -25,23 +26,13 @@ jobs:
       - name: Install capnp dependencies (macOS)
         if: matrix.os == 'macos-latest'
         run: brew install capnp
-      - name: Cache cargo registry
-        uses: actions/cache@v4
-        with:
-          path: ~/.cargo/registry
-          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
-      - name: Cache cargo index
-        uses: actions/cache@v4
-        with:
-          path: ~/.cargo/git
-          key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
-      - name: Cache cargo build
-        uses: actions/cache@v4
-        with:
-          path: target
-          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
       - name: Install Rust
         uses: dtolnay/rust-toolchain@1.88
+      - name: Cache Rust build
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: "false"
+          workspaces: integration-tests -> ../target
       - name: Test stratum-apps workspace
         run: cargo test --manifest-path=stratum-apps/Cargo.toml --all-features
       - name: Test pool
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -32,6 +32,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
+          # Hash every Cargo workspace without registering the shared target more than once.
+          key: cargo-inputs-${{ hashFiles('**/Cargo.toml', '**/Cargo.lock') }}
           workspaces: integration-tests -> ../target
       - name: Test stratum-apps workspace
         run: cargo test --manifest-path=stratum-apps/Cargo.toml --all-features
--- a/integration-tests/.config/nextest.toml
+++ b/integration-tests/.config/nextest.toml
@@ -5,8 +5,14 @@
 # and that's a reliable indication that we shouldn't merge this PR
 retries = { backoff = "fixed", count = 3, delay = "2s" }
 
-# only run one test at a time, which allows a human-friendly experience for inspecting logs
-test-threads = 1
+# Number of test processes to run concurrently.
+#
+# "num-cpus" matches the runner (4 vCPU on ubuntu-latest, 3 vCPU on
+# macos-latest — both free-tier public repos per GitHub docs).
+#
+# For sequential human-readable logs (one test at a time), override locally:
+#     NEXTEST_TEST_THREADS=1 cargo nextest run ...
+test-threads = "num-cpus"
 
 # label as slow if a test runs for more than 60s
 # kill it after 120s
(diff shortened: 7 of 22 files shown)


bit-aloo:
We definitely got a nice speedup here, IT went from 30 mins to 20 mins! 🔥

plebhash:
thanks for the initiative @GitGab19 

SRI community desperately needs to speed up our development process, since we're running on GitHub free tiers and Loupe reports are arriving at a pace in which our meatlayer processes are not being able to keep up with

we're all very grateful for Anthropic's FOSS support with out free-tier Claude Max accounts, but we've also been kinda disappointed at Fable and Opus performance (even when locked into the borderline mandatory Claude Code walled gardened harness, which sorta goes agaist FOSS spirit)

and Kimi K3 has been kicking ass across the entire Bitcoin Ecosystem lately, so we might as well try to leverage on our benefit too

---

so I'll try to leverage Kimi K3 for the creation of a plan, to be executed later by budget-friendly  DeepSeek V4 Pro (both inside [`opencode`](https://opencode.ai/) harness which has been delivering great performance for my agentic workflows):

<details>

<summary>PROMPT to Kimi K3:</summary>

> gitgab19 took some great initiative on #698, let's try to refine it further
> 
> # context
> 
> SRI community is facing development workflow bottlenecks
> 
> SRI community desperately needs to speed up our development process, since we're running on GitHub free tiers and Loupe reports are arriving at a pace in which our meatlayer processes are not being able to keep up with
> 
> it's worth expanding on the implications of Github Runners free tiers:
> - concurrent PRs (which happen A LOT on SRI community workflow) tend to stack up
> - limited number of runners means whenever PR stack goes beyond a certain threshold, there's no further concurrent CI execution, and a stale queue starts forming
> 
> IUCC #698 is trimming some fat from Integration Tests CI global execution
> 
> I think the axis we can explore now is the great prior work from @bit-aloo, which made the genious suggestion of leveraging `cargo nextest`, which IMHO has been underexplored
> 
> up until now, we have been deliberately forcing `test-threads = 1`, while aiming for human-readable local logs (which can only happen if they're sequential)
> 
> let's allow ourselves to lift this restriction, aiming to optimize for concurrent execution, so that the stack is drained faster
> 
> as long as the final `cargo nextest` report has a list of which tests failed, if needed they can later be re-executed in isolation for log inspection
> 
> assuming VM resources of a free Github Runner, let's see how `cargo nextest` can accelerate execution of the global Integration Tests CI (aka ITFCI) stack, without sacrificing determinism or increasing flakiness
> 
> concurrent execution should also deliberately avoid flakiness by making sure concurrent CPU mining do not unintendly/blindly exhaust compute resources (which would reduce determinism)
> 
> in other words: let's optimize FOSS-friendly (aka free) CI compute resources, across the axis of:
> - `cargo nextest`-based optimization
> - global ITFCI execution concurrency
> 
> btw I'm sharing this prompt at https://github.com/stratum-mining/sv2-apps/pull/698#issuecomment-5226260042
> 
> # parameters
> 
> - `cargo nextest` config syntax
> - Github Runner VM resources (free tier, look up online docs instead of speculating)
> - @gitgab19 work on #698
> 
> # methodology
> 
> try to maintain @gitgab19 prior work as much as possible
> 
> add new commits, progressively expanding `speed-up-integration-tests` branch, which we will push to plebhash fork, so they can be potentially cherry-picked into #698

</details>

I'll come back later with some commit suggestions to be cherry-picked

plebhash:
<details>

<summary> PLAN by Kimi K3: </summary>

> 
> # Plan: concurrent ITFCI execution via `cargo nextest` (extending #698)
> 
> Branch: `speed-up-integration-tests` (currently at `b19dfe97`, on top of @gitgab19's 4 commits from #698).
> Goal: lift `test-threads = 1` so the Integration Tests CI drains faster on free-tier GitHub runners, **without** new flakiness sources.
> 
> ## Context / why this works now
> 
> - nextest already runs **each test in its own process**. `test-threads = 1` is the only thing forcing serialization today.
> - @gitgab19's #698 commits already removed the big sequential-bottlenecks (fixed sleeps → readiness gates, 1s → 200ms polls), which also *lowered per-test wall time variance* — the precondition for safe concurrency.
> - GitHub free-tier runner specs (verified against [official docs](https://docs.github.com/en/actions/reference/runners/github-hosted-runners), public repos):
>   - `ubuntu-latest`: **4 vCPU**, 16 GB RAM, 14 GB SSD
>   - `macos-latest`: **3 vCPU (M1)**, 7 GB RAM, 14 GB SSD
> - `test-threads = "num-cpus"` therefore means 4 on Linux, 3 on macOS — self-tuning, no matrix-specific config.
> 
> ## Hazards found while auditing the harness (must fix BEFORE lifting the limit)
> 
> ### H1. Cross-process port allocation race (determinism killer)
> `integration-tests/lib/utils.rs::get_available_port()` probes with `TcpListener::bind("127.0.0.1:0")`, records the port in a `static UNIQUE_PORTS: Mutex<HashSet>`, then **drops the socket**. The static only dedupes *within one process* — under nextest each test is its own process, so with N>1 concurrent tests two processes can both probe the same free port in the window before either role binds it. Worse: bitcoind datadirs are keyed by port (`.bitcoin-{port}`), so a port collision means **two bitcoinds sharing one datadir** → corruption, not just a bind error.
> 
> ### H2. Cold-cache artifact download race (CI-first-run killer)
> `template_provider.rs` downloads/unpacks Bitcoin Core, sv2-tp and `high_diff_chain` into the shared `integration-tests/template-provider/` dir, guarded only by `if !bin.exists()`. On a fresh CI runner, N concurrent test processes all see "missing" and unpack simultaneously — and `tarball::unpack` writes a **fixed-name** `temp.tar.gz` into the destination. Corrupted archives / partial untars on essentially every cold CI run.
> 
> ### H3. CPU oversubscription by mining threads
> `mining_device` auto mode = `logical_cpus - 1` hashing threads **per test process** (`integration-tests/lib/mining_device/mod.rs::worker_count()`). On the 4-vCPU runner with 4 concurrent tests each spawning a mining device: ~12 busy hashing threads → contention, and timing-sensitive share-rate assertions (`SHARES_PER_MINUTE = 120`) go flaky. (`sv1_minerd` is already pinned to `--threads 1`.)
> 
> ### Non-hazards (verified, no action needed)
> - No hardcoded ports anywhere in `integration-tests/` (all via `get_available_address()`).
> - `static LOGGER: OnceCell` — fine, per-process under nextest.
> - Stale-datadir cleanup from #698 (`efc4dca3`) — compatible, keep as-is.
> 
> ---
> 
> ## Commit plan (progressive, each commit leaves the branch green)
> 
> ### Commit 1 — `test(integration): make port allocation race-free across test processes`
> 
> Rewrite `get_available_port()` in `integration-tests/lib/utils.rs` to reserve ports with an **advisory file lock held for the process lifetime**:
> 
> 1. Create lock dir once: `std::env::temp_dir().join("sv2-it-ports")`.
> 2. Probe a free port with `bind("127.0.0.1:0")` as today, **keep the probe socket open**.
> 3. Open/create `{lockdir}/{port}.lock` and try `libc::flock(fd, LOCK_EX | LOCK_NB)`:
>    - success → keep the `File` in a process-global `Vec<File>` (never dropped → lock held until process exit, kernel releases even on `kill -9`, so no stale-lock cleanup needed), drop probe socket, return port;
>    - failure → another test process holds this port: drop everything, loop.
> 4. Remove the now-obsolete `UNIQUE_PORTS` static (the lock dir supersedes it; within-process uniqueness is guaranteed because a held lock makes subsequent `flock` attempts fail).
> 
> Details for the executor:
> - Add `libc = "0.2"` to `integration-tests/Cargo.toml` (already in the transitive tree via `corepc-node`; zero new compile cost). `flock` exists on both Linux and macOS.
> - Keep the `File` handles in a `static HELD_LOCKS: Lazy<Mutex<Vec<std::fs::File>>>` — storing the `File` keeps the fd (and thus the lock) alive.
> - Exact order matters: probe → flock → drop probe. Never drop the probe before the flock succeeds, or another process can re-probe the same port.
> - Existing rustdoc on `get_available_address`/`get_available_port` must be updated to describe the locking scheme (per AGENTS.md: keep rust docs in sync).
> 
> ### Commit 2 — `test(integration): serialize template-provider artifact downloads`
> 
> In `integration-tests/lib/template_provider.rs` (`BitcoinCore::start` and `TemplateProvider::start`):
> 
> - Guard each download+unpack block (bitcoin-core tarball, sv2-tp tarball, `high_diff_chain`) with a blocking exclusive `libc::flock` on a per-artifact lockfile, e.g. `template-provider/.locks/bitcoin-31.0.lock`.
> - **Double-check inside the lock**: re-test `bin.exists()` / `high_diff_chain_dir.exists()` after acquiring the lock — the first waiter must not re-download what the lock holder just unpacked.
> - While here: give `tarball::unpack`'s temporary file a per-process-unique name (e.g. `temp-{pid}.tar.gz`) so even a bug elsewhere can't make two unpacks share the temp file. Cheap insurance, one line.
> 
> Note: this makes the *first* CI run slightly serialized at startup (one process downloads, others wait on the lock, then take the fast path). Do **not** add a CI pre-warm step speculatively — measure first.
> 
> ### Commit 3 — `test(integration): pin harness mining devices to one worker thread`
> 
> In `integration-tests/lib/mod.rs::start_mining_device_sv2()`, call `crate::mining_device::set_cores(1)` before spawning. One line; covers every test that mines through the harness helper. The standalone `mining_device` **binary** and benches keep auto mode (they're not part of ITFCI).
> 
> Expected effect: with `test-threads = num-cpus`, worst-case hashing load ≈ 1 thread per test slot + test runtime threads ≈ the machine's core count, not N×(cores−1).
> 
> ### Commit 4 — `test(integration): run nextest at num-cpus`
> 
> `integration-tests/.config/nextest.toml`:
> 
> ```toml
> # was: test-threads = 1 (human-readable sequential logs)
> # Sequential logs are still available locally via:
> #   cargo nextest run --test-threads=1 --no-capture
> test-threads = "num-cpus"
> ```
> 
> - Keep `retries = 3` — it now also absorbs any residual rare flakes; a test that fails 4× in a row is still a deterministic failure signal.
> - `slow-timeout`/`terminate-after`: **leave untouched initially**; only relax (e.g. `terminate-after = 3`) if the validation runs below show CPU-contended tests getting killed at 120s. Don't tune speculatively.
> - Do **not** add `threads-required` overrides yet — same reason. It's the documented escape hatch if one heavy test (e.g. a multi-bitcoind JD test) measurably thrashes.
> 
> ### Commit 5 — `ci(integration): drop --nocapture so concurrent output stays readable`
> 
> `.github/workflows/integration-tests.yaml`: remove `--nocapture` from the `cargo nextest run` invocation. With concurrency, live stdout from N tests interleaves into garbage; nextest's default captures per-test output and prints it (with `RUST_LOG=debug` logs intact) only for **failing/flaky** tests at the end — which is exactly the "final report lists failures, re-run individually for logs" workflow we want. Debug-rerun instructions go in the commit message and as a comment above the step:
> 
> ```
> # failure triage: re-run a single test with captured logs shown live:
> #   RUST_LOG=debug cargo nextest run --manifest-path=integration-tests/Cargo.toml \
> #     -E 'test(<name>)' --no-capture
> ```
> 
> ---
> 
> ## Validation protocol (run between/after commits)
> 
> Local (dev machine has more cores than CI, so *also* simulate the runner):
> 
> ```bash
> # build once
> cargo nextest run --manifest-path=integration-tests/Cargo.toml --no-run
> 
> # baseline (current behavior) — record wall time
> cargo nextest run --manifest-path=integration-tests/Cargo.toml --test-threads=1
> 
> # simulated ubuntu-latest runner
> cargo nextest run --manifest-path=integration-tests/Cargo.toml --test-threads=4
> # simulated macos-latest runner
> cargo nextest run --manifest-path=integration-tests/Cargo.toml --test-threads=3
> 
> # flake check: 3 consecutive runs at 4 threads, all must pass with zero retries-fired diffs
> for i in 1 2 3; do cargo nextest run --manifest-path=integration-tests/Cargo.toml --test-threads=4; done
> ```
> 
> Acceptance criteria:
> - Wall time at `--test-threads=4` meaningfully below the `--test-threads=1` baseline (expect roughly 2–3×, not linear — many tests are I/O-wait-bound on bitcoind).
> - No new FLAKY entries across the 3 repeat runs. If a specific test goes flaky: first try `threads-required = 2` override for it; only as a last resort relax its timing assertion — and say so in the commit message.
> - `cargo +1.88 clippy --manifest-path integration-tests/Cargo.toml --all-targets` and `cargo fmt --check` clean (CI gates).
> 
> Then push to the fork and compare the GitHub `Integration Tests` job wall time against a recent `main` run, on both `ubuntu-latest` and `macos-latest` (macOS has 3 vCPU/7 GB — it's the constraint; if macOS regresses while Linux improves, consider `test-threads = 2` override via a `[profile.ci-macos]`... only if data says so).
> 
> ---
> 
> ## Estimated improvement (from real CI data)
> 
> Baseline measured on `main`, ubuntu-latest, run [31215815112](https://github.com/stratum-mining/sv2-apps/actions/runs/31215815112) (2026-08-07):
> 
> | Component | Time |
> |---|---|
> | Compile (`cargo nextest run` build phase) | 3m 27s |
> | Test execution, `test-threads = 1` | **25m 40s** (1540s, 108 tests, 0 failed) |
> | **Total `Integration Tests` step** | **~29m 20s** |
> 
> Per-test distribution extracted from the same log: mean ≈ 14s, max = 41s
> (`translator_integration::non_aggregated_translator_correctly_deals_with_group_channels`),
> with a thick band of 30–40s tests (`pool_solo_mining_*`, `*_group_*_channels`). The suite is
> dominated by **I/O-wait** (bitcoind/sv2-tp startup, readiness gates, message waits), not CPU —
> exactly the workload that parallelizes well.
> 
> Model for `test-threads = "num-cpus"`:
> 
> | Factor | Effect |
> |---|---|
> | Ideal 4-way parallelism (ubuntu) | 1540s / 4 ≈ 385s |
> | CPU contention (1 mining thread + tokio runtime + bitcoind bursts per test slot) | ×1.2–1.4 per-test slowdown |
> | Cold-start artifact downloads (bitcoin-core ×2, sv2-tp, high_diff_chain) | overlapped with other tests under the new flock; ≈ net-neutral vs. today |
> | Tail effect (108 tests, max 41s → ≤ ~1 idle slot at the end) | +~30s |
> | Retries under contention | assumed flat (acceptance criterion blocks merge otherwise) |
> 
> | Runner | Serial today | Expected concurrent | Speedup |
> |---|---|---|---|
> | ubuntu-latest (4 vCPU) | 25m 40s | **~8–11 min** | **~2.5–3×** |
> | macos-latest (3 vCPU M1) | ~29m 30s | **~11–14 min** | **~2–2.5×** |
> 
> Including the unchanged ~3.5min compile, the `Integration Tests` job lands at roughly
> **12–15 min instead of ~29–30 min per OS**. Queue-level effect (the actual bottleneck this PR
> targets): each PR's ITFCI occupancy halves, so a stacked PR queue drains **~2× faster** even
> before GitHub-side runner concurrency is considered.
> 
> Bounds: pessimistic case (contention ×1.5 + one retry wave) ≈ 13–15 min test time on ubuntu —
> still ≥1.7× faster than today. Optimistic case (I/O-wait dominance confirmed) ≈ 7 min.
> 
> Longest-pole check: 41s max test × 1.5 worst-case contention ≈ 62s — comfortably under the
> existing 120s `terminate-after`, so no test should hit the kill timeout purely from concurrency.
> 
> These numbers are estimates; the validation protocol above is what confirms them — record the
> branch's first CI run and compare `Summary [ ... ]` against 1540s before declaring victory.
> 
> ## Explicitly out of scope (YAGNI for this PR)
> 
> - Splitting the suite into nextest partitions across matrix jobs — bigger win but orthogonal; revisit only if concurrency alone doesn't drain the queue enough.
> - `ubuntu-slim` / job-level restructuring.
> - Caching `template-provider/` binaries in CI (`actions/cache`) — compatible follow-up, independent of this change.
> - Changing `SHARES_PER_MINUTE` or any test semantics.

</details>

plebhash:
<details>

<summary>EXECUTION report by DeekSeek V4 Pro:</summary>

> # Execution report: concurrent ITFCI via `cargo nextest`
>
> Branch: `speed-up-integration-tests` (plebhash fork).
> Base: `GitGab19/speed-up-integration-tests` @ `b19dfe97` (#698).
>
> ## What was committed
>
> ```text
> 8917834a style(integration): cargo fmt + Cargo.lock update for libc dep
> 1cd62cac ci(integration): drop --nocapture so concurrent output stays readable
> 76659cf8 test(integration): run nextest at num-cpus instead of serial
> d37ff847 test(integration): pin harness mining devices to one worker thread
> 0e569385 test(integration): serialize template-provider artifact downloads across processes
> 66a17269 test(integration): make port allocation race-free across test processes
> b19dfe97 test(integration): stop probing Bitcoin Core's IPC socket with a connect   ← #698 top
> efc4dca3 test(integration): gate on readiness instead of fixed sleeps, drop datadirs
> ca5fd5f9 test(integration): poll on a 200ms interval instead of 1s
> 5181a166 test(integration): fix latent ordering races in the test harness
> ```
>
> ### 1. Cross-process-safe port allocation (`66a17269`)
>
> **Files**: `integration-tests/lib/utils.rs`, `integration-tests/Cargo.toml`
>
> Replaced the in-process-only `UNIQUE_PORTS` dedup set with per-port `flock` lockfiles
> under `$TMPDIR/sv2-it-ports/` held for the process lifetime. Two concurrent nextest
> processes can no longer race on the same ephemeral port. Lockfiles are kernel-released
> even on `kill -9`, so stale files are harmless.
>
> Added `libc = "0.2"` dep (already in the crate's transitive tree via `corepc-node`).
>
> ### 2. Serialized artifact downloads (`0e569385`)
>
> **Files**: `integration-tests/lib/template_provider.rs`, `integration-tests/lib/utils.rs`
>
> Added `with_exclusive_lock()` helper: blocking exclusive flock + inner `exists()`
> re-check so only the first process actually downloads; the rest wait and skip.
> Applied to three critical sections:
> - Bitcoin Core tarball (v30.2 / v31.0)
> - sv2-tp tarball (v1.1.0)
> - `high_diff_chain` unpack
>
> Also made `tarball::unpack` write `temp-{pid}.tar.gz` instead of a static filename.
>
> ### 3. Mining device pinned to 1 thread (`d37ff847`)
>
> **File**: `integration-tests/lib/mod.rs`
>
> `start_mining_device_sv2()` now calls `crate::mining_device::set_cores(1)`.
> Every test that mines through the harness gets a single hashing thread, which
> keeps CPU oversubscription under control on 3–4 vCPU runners. Standalone binary
> and benches are untouched.
>
> ### 4. `test-threads = "num-cpus"` (`76659cf8`)
>
> **File**: `integration-tests/.config/nextest.toml`
>
> Replaced `test-threads = 1` with `test-threads = "num-cpus"` (4 on ubuntu-latest,
> 3 on macos-latest per GitHub docs). Added comment documenting `NEXTEST_TEST_THREADS=1`
> for local sequential debugging.
>
> ### 5. Dropped `--nocapture` in CI (`1cd62cac`)
>
> **File**: `.github/workflows/integration-tests.yaml`
>
> Removed `--nocapture` from the `cargo nextest run` step. nextest captures per-test
> stdout/stderr and prints it only for failures — the exact "final report lists failures,
> re-run individually" workflow requested.  A YAML comment above the step documents the
> triage rerun command.
>
> > **Rebase note:** the original commit had a YAML indentation bug (`- name:` at column 0)
> > that caused "workflow file issue" on the first CI push.  Rebased to fix the indent
> > and move the triage comment out of the `run:` block into a YAML comment.
>
> ### 6. fmt + lockfile (`8917834a`)
>
> **Files**: `integration-tests/Cargo.lock`, `integration-tests/lib/template_provider.rs`,
> `integration-tests/lib/utils.rs`
>
> Applied `cargo fmt` (indentation fixes for the new closures) and updated Cargo.lock
> for the new `libc` dependency.
>
> ## Validation
>
> - `cargo check --manifest-path integration-tests/Cargo.toml` — clean
> - `cargo clippy --manifest-path integration-tests/Cargo.toml --all-targets` — only
>   pre-existing warnings, none from these changes
> - `cargo fmt --manifest-path integration-tests/Cargo.toml -- --check` — clean
> - `integration-tests.yaml`: validated with `python3 -c "import yaml; ..."` — parses correctly
>
> ## CI observation plan
>
> 1. Push to plebhash fork → trigger Integration Tests workflow.
> 2. Compare the `Summary [ ... ]` line against the baseline of **1540s** (from
>    the `main` run at [31215815112](https://github.com/stratum-mining/sv2-apps/actions/runs/31215815112)):
>    - Expected on ubuntu-latest: ~460–660s (8–11 min test time vs 25m40s).
>    - Expected on macos-latest: ~660–840s (11–14 min vs ~29m30s).
> 3. Check for green across both OS matrix entries; zero FLAKY lines.
> 4. If a specific test goes flaky: note it, re-run that test serially to confirm
>    it isn't inherently broken, then consider `threads-required = 2` override.
> 5. If the entire run is flaky: revert `test-threads` to 1 (only last-config commit)
>    and investigate — but keep the port-lock and download-lock commits (commit 1+2)
>    because they are correctness fixes regardless of concurrency.
>
> ## Open for follow-up
>
> - **Per-test `threads-required` overrides** — only if measurements show specific
>   heavy tests (multi-bitcoind JD tests, etc.) thrashing.
> - **Artifact cache** (bitcoin-core / sv2-tp binaries in CI via `actions/cache`) —
>   independent optimization; the flock already prevents first-run download races.
> - **Suite partitioning across matrix jobs** — bigger architectural change;
>   revisit only if per-job wall time is still unacceptable after this.

</details>

proposed commits for cherry-picking:

- [66a1726](https://github.com/stratum-mining/sv2-apps/commit/66a17269e047c6a2c649e1dca008a72a6f7474bd)
- [0e56938](https://github.com/stratum-mining/sv2-apps/commit/0e569385f93c1dac3c23548e9160702795a1a71b)
- [d37ff84](https://github.com/stratum-mining/sv2-apps/commit/d37ff84733cac980e2cfdd28182b2411425741a7)
- [76659cf](https://github.com/stratum-mining/sv2-apps/commit/76659cf880a2df00ad7f7dc67e2c30da65084589)
- [1cd62ca](https://github.com/stratum-mining/sv2-apps/commit/1cd62cac6b19b2cf2b1eb3f91ee03e26e35e18bf)
- [8917834](https://github.com/stratum-mining/sv2-apps/commit/8917834a3147ce21d50bdfce07c2b434333a4f17)

plebhash:
PERFORMANCE report by DeekSeek V4 Pro:

> # Performance report: concurrent ITFCI
>
> **Run:** [plebhash/sv2-apps#4 · Actions #31263476121](https://github.com/plebhash/sv2-apps/actions/runs/31263476121)
> **Branch:** `speed-up-integration-tests` @ `8917834a`
> **Baseline:** `stratum-mining/main` @ [31215815112](https://github.com/stratum-mining/sv2-apps/actions/runs/31215815112) — `test-threads = 1`, serial execution
>
> ## 🏗️ Full job time
>
> | Platform | Total job |
> |---|---|
> | 🐧 ubuntu-latest | **~9m 38s** |
> | 🍎 macos-latest | **~9m 24s** |
>
> Total Integration Tests job time went from **~29m 30s** to **~9m 30s** — roughly
> **3.15× faster end-to-end**.

from ~30min to ~9min

tbh 9 min feels still a bit prohibitive IMO, but great improvements overall

---

a good next step would be to try to explore the following area, which has been left out of plan scope as YAGNI (and is also somewhat aligned to what was done on https://github.com/stratum-mining/sv2-apps/pull/600):

> Splitting the suite into nextest partitions across matrix jobs — bigger win but orthogonal; revisit only if concurrency alone doesn't drain the queue enough.

not sure I want to push for it now though, feels like an area that deserves deeper reasoning before we get our hands dirty

for example, one important aspect worth of consideration: this will also affect the dynamics of how the global org-level Github Runner quota will be consumed, which impacts how concurrent PRs affect eachother (which the model called "queue" on the quoted sentence above)

GitGab19:
@plebhash I cherry-picked your commits and added https://github.com/stratum-mining/sv2-apps/pull/698/commits/a1896b981df20c77285502f878cb79f989461262 and https://github.com/stratum-mining/sv2-apps/pull/698/commits/5701391352b4d0607f1e508c6e860bc8f10b64e2 on top of them.

I did that to address the flakiness exposed by parallel execution of your PR (https://github.com/plebhash/sv2-apps/pull/4).

The first makes minerd setup and teardown concurrency-safe, replaces a fixed startup delay with a readiness check, and properly reaps child processes to prevent nextest LEAK reports.
The second handles a legitimate vardiff UpdateChannel arriving before SubmitSharesExtended, without changing the sniffer’s general behavior.

The latest Ubuntu and macOS runs completed without flaky or leaky tests.

The results are:
- ubuntu runner --> 8m 52s
- macos runner --> 10m 49s

GitGab19:
Update: with the cache commit (https://github.com/stratum-mining/sv2-apps/pull/698/commits/f334f5b44c5e06c34658ad73244e696317bea4ae), now we're running all the jobs in ~8min!

plebhash (approved):

bit-aloo (approved):
Super ACK

PR #706, opened 2026-08-09
Noise-sv2 hardening 

bit-aloo:
companion https://github.com/stratum-mining/stratum/pull/2283

### DIFF
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -1025,7 +1025,7 @@ checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
 [[package]]
 name = "binary_sv2"
 version = "7.0.0"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#59a3e3e3bacd24ba40548022046ff8720157dd70"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#e13179476941ea9707d93d514b95b29105391598"
 dependencies = [
  "buffer_sv2",
  "derive_codec_sv2",
@@ -1210,7 +1210,7 @@ dependencies = [
 [[package]]
 name = "buffer_sv2"
 version = "3.0.1"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#59a3e3e3bacd24ba40548022046ff8720157dd70"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#e13179476941ea9707d93d514b95b29105391598"
 dependencies = [
  "aes-gcm",
 ]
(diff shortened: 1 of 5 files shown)


GitGab19 (commented):
Clanker review:

## Review findings

### `try_write_frame()` can desynchronize the Noise transport

[`try_write_frame()` encrypts the frame before calling `try_write()`](https://github.com/stratum-mining/sv2-apps/blob/204ca1754438c4b198668c30ead4103e2134d940/stratum-apps/src/network_helpers/noise_stream.rs#L204-L218).

Encryption advances the Noise nonce. If the socket returns `WouldBlock`, the method returns `Ok(false)` but discards both the ciphertext and the consumed frame. The next frame then uses a later nonce while the peer is still expecting the discarded one, permanently desynchronizing the connection.

Partial writes have a similar problem: a ciphertext prefix may reach the socket while the remainder is discarded.

This behavior predates this PR, but it is directly relevant to the transport-state hardening. The write half should retain pending ciphertext and its write offset until the complete encrypted frame has been sent.

### Add direct coverage for the split transport transition

The migration to separate `TransportEncryptState` and `TransportDecryptState` appears correct, but there is no focused test covering it.

A bidirectional client/server test should complete the handshake and exchange multiple frames concurrently. This would protect against swapped transport states, incorrect nonce progression, and regressions in future refactors.

bit-aloo:
> Clanker review:
> 
> ## Review findings
> ### `try_write_frame()` can desynchronize the Noise transport
> [`try_write_frame()` encrypts the frame before calling `try_write()`](https://github.com/stratum-mining/sv2-apps/blob/204ca1754438c4b198668c30ead4103e2134d940/stratum-apps/src/network_helpers/noise_stream.rs#L204-L218).
> 
> Encryption advances the Noise nonce. If the socket returns `WouldBlock`, the method returns `Ok(false)` but discards both the ciphertext and the consumed frame. The next frame then uses a later nonce while the peer is still expecting the discarded one, permanently desynchronizing the connection.
> 
> Partial writes have a similar problem: a ciphertext prefix may reach the socket while the remainder is discarded.
> 
> This behavior predates this PR, but it is directly relevant to the transport-state hardening. The write half should retain pending ciphertext and its write offset until the complete encrypted frame has been sent.
> 
> ### Add direct coverage for the split transport transition
> The migration to separate `TransportEncryptState` and `TransportDecryptState` appears correct, but there is no focused test covering it.
> 
> A bidirectional client/server test should complete the handshake and exchange multiple frames concurrently. This would protect against swapped transport states, incorrect nonce progression, and regressions in future refactors.

For first, as its predates this PR, I would open an issue for that and for second one, I might wanna change the current network helpers setup a bit once the changes to codec-sv2, maybe then I will add the test.

plebhash:
issue tracking `try_write_frame` bug: https://github.com/stratum-mining/sv2-apps/issues/734

plebhash (approved):

GitGab19 (approved):
We can now take the commit from https://github.com/stratum-mining/sv2-apps/pull/738 and proceed with the merge of this one.

PR #724, opened 2026-08-14
Handle malformed template provider transactions without panic

### LINKED ISSUE #723: Template Provider can panic JDC by sending malformed raw transaction
`handle_request_tx_data_success` parses every entry of the Template Provider's RequestTransactionDataSuccess.transaction_list via `consensus::deserialize(raw_tx).expect("invalid tx")`. bitcoin::consensus::deserialize::<Transaction> returns Err on any byte string it cannot decode as a Bitcoin transaction, so a single malformed entry in the TP's response converts a recoverable parse error into a panic.

bit-aloo:
closes: #723 
closes: https://github.com/project-loupe/audit-sv2-apps/issues/17

### DIFF
--- a/miner-apps/jd-client/src/lib/channel_manager/template_message_handler.rs
+++ b/miner-apps/jd-client/src/lib/channel_manager/template_message_handler.rs
@@ -368,8 +368,16 @@ impl HandleTemplateDistributionMessagesFromServerOwnedAsync for ChannelManager {
 
         let tx_list: Vec<Transaction> = transactions_data
             .iter_bytes()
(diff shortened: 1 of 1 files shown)


GitGab19 (commented):
Can you add regression tests indicated by Loupe?

bit-aloo:
> Can you add regression tests indicated by Loupe?

Loupe test for this is vague, like it just checks if expect is there or not. I don't think it makes sense to have a file parser test.

GitGab19:
I meant to have a test asserting that we don't panic anymore, simulating the exact case which trigger this.

But it's fine.

GitGab19 (approved):

PR #733, opened 2026-08-18
Pool+JDC: handle new `SeenSharesBudgetExhausted `error variant from `channels_sv2` share validation

plebhash:
companion https://github.com/stratum-mining/stratum/pull/2303


bit-aloo (approved):
ACK

PR #735, opened 2026-08-19
apps: pass `max_past_jobs` to channel constructors

gimballock:
Companion to stratum-mining/stratum#2307, which adds `max_past_jobs: Option<usize>` to the server and client `ExtendedChannel`/`StandardChannel` constructors. `None` and `Some(0)` both select the crate default `MAX_PAST_JOBS`, so nothing changes for callers that pass neither.

Four commits:

1. **`apps: pass max_past_jobs to channel constructors`** — pure plumbing. Every one of the 10 call sites passes `None`, so behaviour is unchanged. 2 in pool, 3 in jd-client, 5 in translator (one a test helper). Group channels are untouched — they take no such parameter.
2. **`TMP: point stratum-core at ...`** — temporarily repoints the `stratum-core` dependency at the companion branch so CI can build, following #733. **Must be reverted** once #2307 merges and the pin is bumped to the merged rev.
3. **`pool: make the past-jobs cap settable from config`** — adds `max_past_jobs` to the pool config, so two otherwise-identical deployments can differ by exactly one value.
4. **`apps: expose max_past_jobs on tProxy and JDC too`** — the same setting in the other two roles, for symmetry (see the discussion below).

Each role keeps its own local config idiom: pool and jd-client via a private field plus getter, translator via a public field. Every shipped config example documents the setting **commented out**, so the library default stays in force unless an operator opts in, and each role has a test asserting exactly that.

Each role also logs the cap actually in force at startup. When unset it comes from `channels_sv2`, so it otherwise appears nowhere in the application's own config and is not recoverable from a running process.

### Why expose it at all

The cap is a retention *window* — `cap / template rate` — and the template rate is a deployment property the library cannot see. The same constant buys 50 s of late-share tolerance at a 1 s template cadence and over 16 minutes at 20 s.

Sizing it also depends on what sits downstream, which differs per role: a pool faces proxies or miners, a translator faces real SV1 firmware, a JDC faces a miner or another proxy. Each sees a different submission-latency distribution, so one value chosen for one role is not automatically right for another.

### DIFF
--- a/pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
+++ b/pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
@@ -25,6 +25,12 @@ pool_signature = "Stratum V2 SRI Pool"
 shares_per_minute = 6.0
 # How many shares do we want to acknowledge in a batch
 share_batch_size = 10
+# Past jobs retained per channel for late-share validation (optional). Unset or 0 uses the
+# channels_sv2 default of 16; set `ceil(16s / your job interval)` to reclaim ~4.5 kB per job.
+# Keep the default if this pool serves job-declaration clients: each SetCustomMiningJob
+# retires the active job, so the client sets the rate.
+# max_past_jobs = 3
+
 
 # Monitoring HTTP server address for exposing channel data (optional)
 monitoring_address = "127.0.0.1:9090"
--- a/pool-apps/pool/src/lib/channel_manager/mod.rs
+++ b/pool-apps/pool/src/lib/channel_manager/mod.rs
@@ -111,6 +111,8 @@ pub struct ChannelManager {
     pool_tag_string: String,
     share_batch_size: usize,
     shares_per_minute: SharesPerMinute,
+    /// Past jobs retained per channel; `None` uses the `channels_sv2` default.
+    max_past_jobs: Option<usize>,
     coinbase_reward_script: CoinbaseRewardScript,
     /// Protocol extensions that the pool supports (will accept if requested by clients).
     supported_extensions: Vec<u16>,
@@ -183,6 +185,14 @@ impl ChannelManager {
             downstream_receiver,
         };
 
+        // Record the cap actually in force. When unset it comes from the `channels_sv2`
+        // default, so it appears nowhere in this application's own config and is
+        // otherwise not recoverable from a running pool.
+        match config.max_past_jobs() {
+            Some(n) if n > 0 => info!(max_past_jobs = n, "past-jobs retention cap configured"),
+            _ => info!("past-jobs retention cap: using channels_sv2 default"),
+        }
+
         let channel_manager = ChannelManager {
             downstreams: SharedMap::new(),
             extranonce_allocator: SharedLock::new(extranonce_allocator),
@@ -194,6 +204,7 @@ impl ChannelManager {
             channel_manager_io,
             share_batch_size: config.share_batch_size(),
             shares_per_minute: config.shares_per_minute(),
+            max_past_jobs: config.max_past_jobs(),
             pool_tag_string: config.pool_signature().to_string(),
             coinbase_reward_script: config.coinbase_reward_script().clone(),
             supported_extensions: config.supported_extensions().to_vec(),
--- a/integration-tests/Cargo.lock
+++ b/integration-tests/Cargo.lock
@@ -1025,7 +1025,7 @@ checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
 [[package]]
 name = "binary_sv2"
 version = "7.0.0"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#941d004aaa034efd1773109fd1794812b32605c3"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#d3a05cfa8b51c6923df81f42bdac7f76b2147830"
 dependencies = [
  "buffer_sv2",
  "derive_codec_sv2",
@@ -1210,7 +1210,7 @@ dependencies = [
 [[package]]
 name = "buffer_sv2"
 version = "3.0.1"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#941d004aaa034efd1773109fd1794812b32605c3"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#d3a05cfa8b51c6923df81f42bdac7f76b2147830"
 dependencies = [
  "aes-gcm",
 ]
@@ -1347,7 +1347,7 @@ dependencies = [
 [[package]]
 name = "channels_sv2"
 version = "8.0.0"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#941d004aaa034efd1773109fd1794812b32605c3"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#d3a05cfa8b51c6923df81f42bdac7f76b2147830"
 dependencies = [
  "binary_sv2",
  "bitcoin",
@@ -1460,7 +1460,7 @@ dependencies = [
 [[package]]
 name = "codec_sv2"
 version = "7.0.0"
-source = "git+https://github.com/stratum-mining/stratum?branch=main#941d004aaa034efd1773109fd1794812b32605c3"
+source = "git+https://github.com/stratum-mining/stratum?branch=main#d3a05cfa8b51c6923df81f42bdac7f76b2147830"
 dependencies = [
  "binary_sv2",
  "buffer_sv2",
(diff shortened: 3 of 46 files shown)


pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
```diff
@@ -25,6 +25,14 @@ pool_signature = "Stratum V2 SRI Pool"
 shares_per_minute = 6.0
 # How many shares do we want to acknowledge in a batch
 share_batch_size = 10
+# Past jobs retained per channel for late-share validation (optional).
+#
+# Unset — or 0 — uses the channels_sv2 default. The cap is really a retention window,
+# `cap / template rate`, so the value that suits a deployment depends on its template
+# cadence and its miners' submission latency, neither of which the library can see.
+# A share can only reach back as far as `1 + ceil(latency / template period)` jobs.
+# max_past_jobs = 50
```
plebhash: too verbose, can we make it more concise?

pool-apps/pool/src/lib/channel_manager/mod.rs
```diff
@@ -183,6 +185,15 @@ impl ChannelManager {
             downstream_receiver,
         };
 
+        // Log the effective retention cap once at startup. An A/B that varies this between
+        // otherwise-identical deployments has to be able to prove which side ran which value;
+        // reading it back off the config file is not the same as reading it out of the process
+        // that used it.
```
plebhash: this comment is a bit nonsensical

plebhash:
I'm confused with the approaches taken by Pool vs tProxy/JDC

why does only Pool get the optional config parameter?

pool-apps/pool/config-examples/mainnet/pool-config-bitcoin-core-ipc-example.toml
gimballock: it should better match the surrounding file's conventions of verbosity

pool-apps/pool/src/lib/channel_manager/mod.rs
gimballock: oops, test details leaking into the comments here. cleaned up now

gimballock:
> I'm confused with the approaches taken by Pool vs tProxy/JDC
> 
> why does only Pool get the optional config parameter?

oversight, I was running the test off of the pool and focusing on that. fixed now

plebhash:
thanks @gimballock 

I order to reduce the back-and-forth and expedite this getting merged, I did some minor changes to your branch and pushed to https://github.com/plebhash/sv2-apps/tree/feat/configurable-max-past-jobs

I wasn't able to force-push them here, can you cherry-pick them in favor of the current commits?

basically I reworded some comments

gimballock:
Done — your commit is now the branch tip (`7815f09d`), your wording verbatim. Thanks for taking the reword on yourself; the two structural fixes in there were real bugs on my side:

- the `max_past_jobs` block had orphaned the `# LAN subnet ...` comment from its `[miner_telemetry]` section in all 19 JDC/tProxy examples;
- in `TranslatorConfig` I'd wedged the field between the `supported_extensions` doc comment and `supported_extensions` itself, so that comment was documenting the wrong field.

While checking your text against the pinned `stratum-core` rev I found an inaccuracy that both of us had — my original said it too, so this isn't on your reword. `channels_sv2` has **two** defaults:

- `server::jobs::job_store::MAX_PAST_JOBS = 16`
- `client::MAX_PAST_JOBS = 50`

Which one a role gets depends on the channel kind, and the roles differ:

| role | channels created | actual default |
|---|---|---|
| pool | server only | 16 ✅ as documented |
| tProxy | `client::ExtendedChannel` only, upstream *and* downstream | **50** |
| JDC | `client::ExtendedChannel` upstream, `new_for_job_declaration_client` (server) downstream | **50 and 16** — one key, both |

So "the channels_sv2 default of 16" was simply wrong for tProxy, and half-right for JDC.

The sizing advice inherits the same problem. `ceil(16s / job interval)` comes from server-side measurement of a retention *window*. The client default isn't a window — 50 was chosen to match the server cap, per its own doc, "so a proxy's client channel never evicts a job its upstream still accepts." An operator who follows our advice and sets `max_past_jobs = 3` on a tProxy makes it reject late SV1 shares as `InvalidJobId` that the pool would still have credited. That's the opposite of a memory optimisation with no downside, and it's exactly the operator our commented-out example was aimed at.

Pushed `9db4a770` on top: your pool text untouched (16 is right there), tProxy and JDC corrected to the client default with the share-loss tradeoff stated instead of the window formula. Comments only, and both `every_example_documents_the_cap_commented_out` tests plus the three pool config tests still pass.

Separately, and for #2307 rather than here: `client::MAX_PAST_JOBS`'s doc justifies 50 as "matches the server-side cap," which stopped being true when that PR moved the server cap to 16. Worth a follow-up there.

The `TMP:` commit still needs reverting once #2307 merges.

plebhash:
@gimballock please see https://github.com/stratum-mining/stratum/pull/2307#issuecomment-5516256311

client side has already been set to 16 on my branch, so I think 9db4a7701aec002b190cf5e3efb4ad3b88be205c is not necessary here.

gimballock:
Correcting my last comment, and the branch: **head is back to your `7815f09d`**, unmodified. Sorry for the churn — I pushed two commits on top and have removed both.

The first (`9db4a770`) claimed the client-side default was 50 against a server-side 16, and told tProxy/JDC operators not to lower it. That was true of the rev this branch pins, but you'd already fixed the underlying problem in stratum-mining/stratum#2307 by moving `client::MAX_PAST_JOBS` to 16 as well — which I hadn't seen when I wrote it. With both constants at 16 your original wording is simply correct, and my "correction" was correcting the library's inconsistency rather than anything in this PR. I've cherry-picked that #2307 branch too, so the client default is 16 there now.

The second (`31d54b59`) was a one-word change naming the tProxy's `T` as the upstream's job interval. Right as far as it went, but it asserted the JDC examples "already read correctly," and on re-reading they're the ambiguous ones:

**One `max_past_jobs` key on JDC covers two channel kinds with two different job rates.** The upstream channel is a `client::ExtendedChannel` whose job rate is the JDC's own `SetCustomMiningJob` declare rate. The downstream channels come from `new_for_job_declaration_client` — server-side — and their rate is the JDC's own job production. So `ceil(16s / your job interval)` has two candidate `T`s there, and the smaller cap governs whichever channel is tighter against its own rate.

Not proposing a commit for it — the config comments defer to `channels_sv2`, and after #2307 the library doc carries the custom-job caveat on both the client and server constants, which is the right place for it. Flagging it in case you think the JDC example deserves a sentence; if not, this PR is your cherry-pick and nothing else.

Still outstanding, unchanged: the `TMP:` commit needs reverting and the pin bumped once #2307 merges.

plebhash:
let's replace 8e2d5e6a743d8bc1554dfdc5a6c0ec636949894f with the commit from https://github.com/stratum-mining/sv2-apps/pull/828

gimballock:
Done — `8e2d5e6a` is gone, replaced by `7961ceec` (your `c67bc72e` from #828). Branch is now two commits on `9d49b984`:

```
2c1d3a61 apps: expose max_past_jobs on Pool, tProxy and JDC
7961ceec chore(deps): update stratum-core to d3a05cf
```

This is better than a swap of one pin for another, because #2307 merged in the meantime and `d3a05cf` *is* its merge commit. So the fork pin isn't just replaced, it's no longer needed at all:

- both `Cargo.toml` files are back on `stratum-mining/stratum` `branch = "main"` — `grep -r marafoundation` across every `Cargo.toml`/`Cargo.lock` returns nothing;
- all four lockfiles sit at `d3a05cf`;
- the feature commit no longer touches any lockfile, so the dep bump and the feature are cleanly separated;
- **the "must be reverted before merge" blocker in the PR description is discharged.**

I also dropped the lockfile churn my own commit had been carrying and rewrote its first paragraph, which still described #2307 as unmerged.

Verified against `d3a05cf`: `MAX_PAST_JOBS` is 16 on both sides as merged (`server::jobs::job_store` and `client`), so the 30 config examples and doc comments that say "the channels_sv2 default of 16" are now correct for every role — pool, tProxy and JDC alike. Tests green: 48 translator, 11 JDC, 5 pool, plus both `every_example_documents_the_cap_commented_out` tests; clippy clean on both workspaces.

One thing worth recording, since it was invisible until an hour ago: this PR had **no CI at all** for its whole life. Every workflow here triggers on `pull_request`, which needs `refs/pull/735/merge`, and GitHub can't compute that ref while a PR conflicts. The branch had been conflicting against main since the `TMP:` pin started colliding with the automated `chore(deps)` commits that rewrite the same lockfile lines. Rebasing onto current main dispatched 10 checks for the first time. Worth knowing that a stale TMP-pin branch silently gets zero coverage rather than red checks.

plebhash (approved):

PR #741, opened 2026-08-19
Fix tProxy issues found by Loupe

### LINKED ISSUE #743: tProxy: bound SV1 handshake messages queued before channel establishment
## Problem

Before the corresponding upstream SV2 channel is established, tProxy stores incoming SV1 handshake messages in `DownstreamData::queued_sv1_handshake_messages`.

This queue is a `Vec` owned by each downstream connection and previously had no limit. A miner could continue sending messages while the upstream channel request was pending, causing memory consumption to grow for that connection until the channel opened or the process ran out of memory.

The queue is specific to tProxy’s SV1-to-SV2 handshake state machine; the other applications do not maintain this per-miner SV1 message buffer.

## Expected behavior

The pre-channel queue should accept only the small number of messages required to complete a valid SV1 handshake. Once that limit is exceeded, tProxy should disconnect the offending downstream without affecting other miners.

Queued messages must still be processed in order after the upstream channel opens.

## Acceptance criteria

- Introduce a documented per-downstream queue limit based on the valid SV1 handshake sequence.
- Disconnect only the downstream that exceeds the limit.
- Preserve normal `mining.subscribe`, `mining.authorize`, and related handshake processing.
- Add tests covering the limit, overflow behavior, and isolation between downstreams.

## References

- Loupe finding: project-loupe/audit-sv2-apps#26

### LINKED ISSUE #744: tProxy: make SV1 handshake completion idempotent
## Problem

A downstream can send `mining.authorize` more than once. tProxy previously allowed every successful authorization to run the handshake-completion side effects again.

That could drain cached handshake messages, apply pending difficulty state more than once, or otherwise repeat state transitions that should occur only when the downstream changes from unauthenticated to authenticated.

Repeated authorization requests may receive a normal protocol response, but they must not replay channel-establishment side effects.

## Expected behavior

Handshake completion must occur exactly once for each downstream connection. The authenticated state transition should be guarded atomically, while subsequent authorization requests should follow an explicitly documented idempotent policy.

## Acceptance criteria

- Guard handshake completion with a false-to-true authentication transition.
- Do not drain cached messages or reapply pending state on repeated authorization.
- Preserve the response expected by valid miners that repeat `mining.authorize`.
- Add tests showing that repeated authorization cannot trigger handshake completion twice.

## References

- Loupe finding: project-loupe/audit-sv2-apps#127

### LINKED ISSUE #745: tProxy: continue difficulty broadcasts after a downstream disconnects
## Problem

When tProxy distributes a difficulty update to multiple SV1 downstreams, sending to a disconnected receiver can fail. The broadcast path previously returned at the first such failure, preventing later healthy downstreams from receiving the update.

A single stale receiver could therefore interfere with difficulty delivery to unrelated miners.

JDC already forwards independently prepared downstream messages one at a time and continues after an individual delivery failure. This particular aborting broadcast path is local to tProxy.

## Expected behavior

A failed downstream delivery should remove or disconnect that downstream and allow the broadcast to continue. It must not become a global shutdown or prevent delivery to the remaining peers.

## Acceptance criteria

- Treat a closed receiver as a downstream-local failure.
- Continue delivering the same difficulty update to all remaining downstreams.
- Clean up the stale downstream state.
- Add a deterministic test with a closed receiver ordered before a healthy receiver.

## References

- Loupe finding: project-loupe/audit-sv2-apps#130

### LINKED ISSUE #811: require an explicit payout-verification policy
## Problem

verify_payout is a security-sensitive configuration option, but its Serde default silently sets it to false when the field is absent.

Existing or abbreviated configurations can therefore disable coinbase payout validation without an explicit operator decision. When validation is disabled, an authenticated but malicious or compromised upstream can send jobs paying an unexpected script without tProxy detecting the diverted reward.

This is distinct from validation of unknown or misspelled configuration fields: the configuration can be otherwise valid while genuinely omitting verify_payout.

## Expected behavior

The operator must explicitly choose whether payout verification is enabled.

Omitting the setting should produce an actionable configuration error rather than silently selecting the insecure behavior. An explicit opt-out may remain supported, but its consequences must be documented.

## Acceptance criteria

- Remove the implicit false default for verify_payout.
- Fail configuration loading when the setting is absent.
- Keep both explicit true and explicit false configurations supported if intentional opt-out remains part of the policy.
- Document the security implications of disabling payout verification.
- Update all example and test configurations to state the policy explicitly.
- Recommend verification when the upstream identity contains an expected payout address.
- Add tests for a missing value, explicit enablement, and explicit disablement.
- Ensure startup clearly reports the selected policy without exposing sensitive identity data.

## Loupe findings

- [project-loupe/audit-sv2-apps#185](https://github.com/project-loupe/audit-sv2-apps/issues/185)

### LINKED ISSUE #812: enforce a collision-free channel and group ID namespace
## Problem

tProxy routes extended channels and group channels through what is effectively a shared numeric namespace, but two upstream-controlled paths do not enforce uniqueness across that namespace.

In non-aggregated mode, OpenExtendedMiningChannelSuccess can reuse the ID of an existing extended channel. The unconditional insertion replaces the first miner's channel state while stale group membership and SV1 routing continue to reference the same ID.

SetGroupChannel can also introduce a group ID that matches a live extended channel. Because later handlers often check group channels first, jobs, targets, previous hashes, and close messages for the colliding ID can be applied to the wrong scope.

These collisions can strand miners, mix channel state, broadcast channel-specific work, or allow one channel lifecycle to tear down another.

## Expected behavior

Every live individual channel, group channel, and reserved aggregated identifier must occupy a collision-free routing namespace.

An upstream message introducing a collision must be rejected before any state is mutated or any success is forwarded downstream.

## Acceptance criteria

- Reject an upstream channel-open success whose channel ID is already live.
- Reject a group-channel ID that collides with an individual channel, another group, or a reserved aggregate identifier.
- Perform all collision checks before mutating channel maps, group membership, SV1 routing, or accounting state.
- Return a typed, controlled protocol error without leaving partial state.
- Permit ID reuse only after the previous channel or group has been completely removed.
- Preserve the existing aggregated-channel reserved-ID protection.
- Add tests for duplicate extended-channel success responses.
- Add tests for group-to-extended, group-to-group, and reserved-ID collisions.
- Verify jobs, targets, previous hashes, shares, and close messages remain scoped to the intended channel after a rejected collision.
- Add a test for valid reuse after complete cleanup.

## Loupe findings

- [project-loupe/audit-sv2-apps#187](https://github.com/project-loupe/audit-sv2-apps/issues/187)
- [project-loupe/audit-sv2-apps#188](https://github.com/project-loupe/audit-sv2-apps/issues/188)

Loupe [#187](https://github.com/project-loupe/audit-sv2-apps/issues/187) covers duplicate upstream extended-channel IDs. Loupe [#188](https://github.com/project-loupe/audit-sv2-apps/issues/188) covers group IDs that shadow live extended channels.

### LINKED ISSUE #813: generate one aggregated keepalive job per tick
## Problem

In aggregated mode, the SV1 keepalive loop iterates over every downstream and mutates the shared aggregated channel's last job once for each miner.

Each iteration increments nTime, appends another job to the same shared history, and sends that particular mutation to one downstream. A single timer tick therefore advances the shared job once per connected miner instead of once per aggregated channel.

With enough connected miners, one tick can advance the shared job to the maximum future-time limit. Later miners receive no keepalive, and subsequent ticks cannot generate another one from the already capped job. The loop also clones and retains one full notification per downstream.

## Expected behavior

Each keepalive tick should generate at most one new job for the aggregated channel and broadcast that exact job to every eligible downstream.

The shared aggregated history should be updated once, independently of the number of connected miners.

## Acceptance criteria

- Read and mutate the aggregated channel's last job only once per tick.
- Increment nTime at most once per aggregated keepalive interval.
- Append at most one keepalive job to aggregated history per tick.
- Broadcast the same generated notification to every eligible downstream.
- Avoid holding shared channel or downstream locks while performing network sends.
- Preserve the existing per-channel behavior outside aggregated mode.
- Handle an empty downstream set without mutating history unnecessarily.
- Add regression tests with two downstreams and with enough downstreams to reach the future-time cap under the old behavior.
- Verify history growth is constant per tick rather than proportional to connected miners.
- Verify all downstreams receive the same job ID and nTime.

## Loupe findings

- [project-loupe/audit-sv2-apps#195](https://github.com/project-loupe/audit-sv2-apps/issues/195)

GitGab19:
This PR:
- Clarify the intentional vardiff target-ordering policy.
- Make SV1 handshake completion idempotent.
- Continue difficulty broadcasts after an individual downstream disconnects.
- Bound SV1 handshake messages buffered before channel establishment.
- Document that payout verification defaults to disabled and that this trusts the upstream payout policy.
- Reject upstream channel and group identifiers that collide in non-aggregated mode.
- Generate and retain one shared aggregated keepalive job per tick, then deliver it to every eligible downstream.

Closes project-loupe/audit-sv2-apps#26
Closes #743
Closes project-loupe/audit-sv2-apps#126
Closes project-loupe/audit-sv2-apps#127
Closes #744
Closes project-loupe/audit-sv2-apps#130
Closes #745
Closes project-loupe/audit-sv2-apps#187
Closes project-loupe/audit-sv2-apps#185
Closes #811 
Closes project-loupe/audit-sv2-apps#188
Closes #812
Closes project-loupe/audit-sv2-apps#195
Closes #813

### DIFF
--- a/integration-tests/tests/bitcoin_core_ipc_jdp_io.rs
+++ b/integration-tests/tests/bitcoin_core_ipc_jdp_io.rs
@@ -6,8 +6,8 @@
 //! - `DeclareMiningJob` returns `Error(stale-chain-tip)` when the declared BIP34 height is
 //!   intentionally mismatched.
 //! - `DeclareMiningJob` does not retain client-supplied transactions when validation fails.
-//! - `DeclareMiningJob` rejects a coinbase that does not carry exactly one input, without
-//!   tearing down the IPC connection.
+//! - `DeclareMiningJob` rejects a coinbase that does not carry exactly one input, without tearing
+//!   down the IPC connection.
 //!
 //! File structure:
 //! - top: version-specific `#[tokio::test]` wrappers.
--- a/miner-apps/translator/README.md
+++ b/miner-apps/translator/README.md
@@ -95,8 +95,11 @@ Make sure the machine running the Translator Proxy has its clock synced with an 
   address as the username.
 
 #### **Solo/Donation Payout Verification**
-Payout verification is disabled by default. Set `verify_payout = true` for solo mining or
-donation configurations where `user_identity` intentionally encodes an on-chain payout address:
+Payout verification defaults to `false` for compatibility with standard pool mining. When payout
+verification is disabled, tProxy trusts the upstream server's coinbase payout policy and does not
+verify that jobs pay the miner.
+Set `verify_payout = true` for solo mining, or donation configurations where `user_identity`
+intentionally encodes an on-chain payout address:
 
 - `sri/solo/<payout_address>/<worker>`: tProxy verifies every upstream extended job pays 100% of spendable coinbase outputs to `<payout_address>`
 - `<payout_address>[.worker]`: legacy solo mode, verified by checking that at least 90% of spendable coinbase outputs go to `<payout_address>`
--- a/miner-apps/translator/README.md
+++ b/miner-apps/translator/README.md
@@ -115,6 +115,14 @@ If verification fails, tProxy triggers upstream fallback instead of forwarding t
   - When `true`: Translator manages difficulty adjustments based on share submission rates
   - When `false`: Upstream manages difficulty, translator forwards SetTarget messages to miners
 
+- `job_keepalive_interval_secs`: Idle interval before a miner receives a keepalive
+  `mining.notify`; `0` disables keepalives. Only miners whose own interval has elapsed
+  receive one. In aggregated mode, tProxy generates at most one new shared keepalive
+  per interval. A miner becoming due between advances receives the existing
+  shared job. A late joiner also receives the current shared job without resetting
+  shared history or the keepalive schedule. Normal upstream jobs are forwarded
+  without waiting for keepalive deadlines.
+
 #### **Miner Telemetry**
 Translator Proxy can enrich the monitoring API with telemetry from the ASICs connected to its SV1
 port. This is useful when you want the UI to show each miner's management IP, firmware, hashrate,
--- a/miner-apps/translator/src/lib/sv1/downstream.rs
+++ b/miner-apps/translator/src/lib/sv1/downstream.rs
@@ -5,14 +5,7 @@ use crate::{
 use async_channel::{Receiver, Sender};
 #[cfg(feature = "monitoring")]
 use std::net::IpAddr;
-use std::{
-    future::Future,
-    sync::{
-        Arc,
-        atomic::{AtomicBool, Ordering},
-    },
-    time::Instant,
-};
+use std::{future::Future, sync::Arc, time::Instant};
 use stratum_apps::{
     channel_utils::ReceiverCleanup,
     fallback_coordinator::FallbackCoordinator,
@@ -63,6 +56,17 @@ impl DownstreamIo {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum Sv1HandshakeState {
+    /// The original authorize-triggered notification flush has not started.
+    Pending,
+    /// One caller is flushing cached mining notifications. New notifications stay cached until
+    /// that caller observes an empty notify cache under the same lock and enables forwarding.
+    Completing,
+    /// Cached mining notifications were delivered and normal forwarding is enabled.
+    Complete,
+}
+
 #[derive(Debug)]
 pub struct DownstreamData {
     pub channel_id: Option<ChannelId>,
@@ -81,6 +85,7 @@ pub struct DownstreamData {
     pub sv1_worker_name: String,
     pub cached_set_difficulty: Option<json_rpc::Message>,
     pub cached_notify: Option<json_rpc::Message>,
+    pub(super) sv1_handshake_state: Sv1HandshakeState,
     // Next advertised SV1 target, applied when the corresponding
     // mining.set_difficulty is sent with a new mining.notify.
     pub pending_target: Option<Target>,
@@ -120,6 +125,7 @@ impl DownstreamData {
             sv1_worker_name: String::new(),
             cached_set_difficulty: None,
             cached_notify: None,
+            sv1_handshake_state: Sv1HandshakeState::Pending,
             pending_target: None,
             pending_hashrate: None,
             stable_hashrate: false,
@@ -171,8 +177,6 @@ pub struct Downstream {
     pub downstream_id: DownstreamId,
     pub downstream_data: SharedLock<DownstreamData>,
     pub downstream_io: DownstreamIo,
-    // Flag to track if SV1 handshake is complete (subscribe + authorize)
-    pub sv1_handshake_complete: Arc<AtomicBool>,
     /// Per-connection cancellation token (child of the global token).
     /// Cancelled when this downstream's task loop exits, causing
     /// the associated SV1 I/O task to shut down.
@@ -274,7 +278,6 @@ impl Downstream {
             downstream_id,
             downstream_data,
             downstream_io: downstream_channel_io,
-            sv1_handshake_complete: Arc::new(AtomicBool::new(false)),
             downstream_cancellation_token,
         }
     }
@@ -387,102 +390,75 @@ impl Downstream {
         match self.downstream_io.sv1_server_receiver.recv().await {
             Ok(message) => {
                 let downstream_id = self.downstream_id;
-                let handshake_complete = self.sv1_handshake_complete.load(Ordering::SeqCst);
 
-                // Handle messages based on message type and handshake state
                 if let Message::Notification(notification) = &message {
-                    // For notifications (mining.set_difficulty, mining.notify), only send if
-                    // handshake is complete
-                    if handshake_complete {
-                        match notification.method.as_str() {
-                            "mining.set_difficulty" => {
-                                // Cache the Sv1 set_difficulty message to be sent before the next
-                                // notify
-                                debug!(
-                                    "Down: Caching mining.set_difficulty to send before next mining.notify"
-                                );
-                                self.downstream_data
-                                    .with(|d| d.cached_set_difficulty = Some(message))
-                                    .map_err(TproxyError::shutdown)?;
-                                return Ok(());
-                            }
-                            "mining.notify" => {
-                                let (pending_set_difficulty, notify_opt) = self
-                                    .downstream_data
-                                    .with(|d| {
-                                        let cached_set_difficulty = d.cached_set_difficulty.take();
-
-                                        // Prepare the notify message and update state
-                                        let notify_result = server_to_client::Notify::try_from(
+                    match notification.method.as_str() {
+                        "mining.set_difficulty" => {
+                            // Difficulty changes are always paired with the next notify. Keeping
+                            // the handshake state and cache under the same lock prevents handshake
+                            // completion from draining a newly arrived difficulty by itself.
+                            let handshake_state = self
+                                .downstream_data
+                                .with(|data| {
+                                    data.cached_set_difficulty = Some(message);
+                                    data.sv1_handshake_state
+                                })
+                                .map_err(TproxyError::shutdown)?;
+                            debug!(
+                                ?handshake_state,
+                                "Down: Caching mining.set_difficulty to send before next mining.notify"
+                            );
+                            return Ok(());
+                        }
+                        "mining.notify" => {
+                            let messages_to_send = self
+                                .downstream_data
+                                .with(|data| {
+                                    if data.sv1_handshake_state != Sv1HandshakeState::Complete {
+                                        data.cached_notify = Some(message.clone());
+                                        let notify = server_to_client::Notify::try_from(
                                             notification.clone(),
-                                        );
-                                        if let Ok(mut notify) = notify_result {
-                                            if cached_set_difficulty.is_some() {
-                                                notify.clean_jobs = true;
-                                            }
-                                            d.last_job_version_field = Some(notify.version.0);
+                                        )
+                                        .expect("this must be a mining.notify");
+                                        data.last_job_version_field = Some(notify.version.0);
+                                        return None;
+                                    }
 
-                                            // Update target and hashrate if we're sending
-                                            // set_difficulty
-                                            if cached_set_difficulty.is_some() {
-                                                if let Some(new_target) = d.pending_target.take() {
-                                                    d.target = new_target;
-                                                }
-                                                if let Some(new_hashrate) =
-                                                    d.pending_hashrate.take()
-                                                {
-                                                    d.hashrate = Some(new_hashrate);
-                                                }
-                                            }
-                                            // Update last job received time for keepalive tracking
-                                            d.last_job_received_time = Some(Instant::now());
-                                            (cached_set_difficulty, Some(notify))
-                                        } else {
-                                            (cached_set_difficulty, None)
+                                    let cached_set_difficulty = data.cached_set_difficulty.take();
+                                    let mut notify =
+                                        server_to_client::Notify::try_from(notification.clone())
+                                            .expect("this must be a mining.notify");
+                                    if cached_set_difficulty.is_some() {
+                                        notify.clean_jobs = true;
+                                        if let Some(new_target) = data.pending_target.take() {
+                                            data.target = new_target;
                                         }
-                                    })
-                                    .map_err(TproxyError::shutdown)?;
+                                        if let Some(new_hashrate) = data.pending_hashrate.take() {
+                                            data.hashrate = Some(new_hashrate);
+                                        }
+                                    }
+                                    data.last_job_version_field = Some(notify.version.0);
+                                    data.last_job_received_time = Some(Instant::now());
+                                    Some((cached_set_difficulty, Message::from(notify)))
+                                })
+                                .map_err(TproxyError::shutdown)?;
 
-                                if let Some(set_difficulty_msg) = &pending_set_difficulty {
-                                    debug!(
-                                        "Down: Sending pending mining.set_difficulty before mining.notify"
-                                    );
-                                    self.downstream_io
-                                        .downstream_sv1_sender
-                                        .send(set_difficulty_msg.clone())
-                                        .await
-                                        .map_err(|e| {
-                                            error!(
-                                                "Down: Failed to send mining.set_difficulty to downstream: {:?}",
-                                                e
-                                            );
-                                            TproxyError::disconnect(TproxyErrorKind::ChannelErrorSender, downstream_id)
-                                        })?;
-                                }
-
-                                if let Some(notify) = notify_opt {
-                                    debug!("Down: Sending mining.notify");
-                                    self.downstream_io
-                                        .downstream_sv1_sender
-                                        .send(notify.into())
-                                        .await
-                                        .map_err(|e| {
-                                            error!("Down: Failed to send mining.notify to downstream: {:?}", e);
-                                            TproxyError::disconnect(TproxyErrorKind::ChannelErrorSender, downstream_id)
-                                        })?;
-                                }
+                            let Some((pending_set_difficulty, notify)) = messages_to_send else {
+                                debug!("Down: SV1 handshake not complete, caching mining.notify");
                                 return Ok(());
-                            }
-                            _ => {
-                                // Other notifications - forward if handshake complete
+                            };
+
+                            if let Some(set_difficulty) = pending_set_difficulty {
+                                debug!(
+                                    "Down: Sending pending mining.set_difficulty before mining.notify"
+                                );
                                 self.downstream_io
                                     .downstream_sv1_sender
-                                    .send(message.clone())
+                                    .send(set_difficulty)
                                     .await
-                                    .map_err(|e| {
+                                    .map_err(|error| {
                                         error!(
-                                            "Down: Failed to send notification to downstream: {:?}",
-                                            e
+                                            "Down: Failed to send mining.set_difficulty to downstream: {error:?}"
                                         );
                                         TproxyError::disconnect(
                                             TproxyErrorKind::ChannelErrorSender,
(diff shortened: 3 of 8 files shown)


miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -1436,17 +1458,22 @@ impl Sv1Server {
...
             }
         }
+
+        if let Some(downstream_id) = disconnected_downstream {
+            return Err(TproxyError::disconnect(
+                TproxyErrorKind::ChannelErrorSender,
+                downstream_id,
+            ));
+        }
+
```
bit-aloo, line 1498: Why we did this? early return was the same?

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -578,18 +578,27 @@ impl Downstream {
...
+                ))
             })
             .map_err(TproxyError::shutdown)?;
+        let Some((cached_set_difficulty, cached_notify, downstream_id)) = completion else {
+            debug!(
+                "Down: Ignoring repeated SV1 handshake completion for downstream {}",
+                self.downstream_id
+            );
+            return Ok(());
+        };
```
bit-aloo: Lets do the early completion check like this:


```suggestion
    if self
        .sv1_handshake_complete
        .swap(true, Ordering::SeqCst)
    {
        debug!(
            "Down: Ignoring repeated SV1 handshake completion for downstream {}",
            self.downstream_id
        );
        return Ok(());
    }

    let (cached_set_difficulty, cached_notify) = self
        .downstream_data
        .with(|d| {
            (
                d.cached_set_difficulty.take(),
                d.cached_notify.take(),
            )
        })
        .map_err(TproxyError::shutdown)?;
```

miner-apps/translator/src/lib/translator_runtime.rs
```diff
@@ -565,14 +664,26 @@ impl TranslatorRuntime<Running> {
...
             biased;
             _ = cancellation_token.cancelled() => RuntimeEvent::Shutdown,
+            requested_endpoint = self.protocol_reconnect_receiver.recv() => {
+                match requested_endpoint {
+                    Ok(requested_endpoint) => RuntimeEvent::ProtocolReconnect {
+                        requested_endpoint,
+                    },
+                    Err(_) => RuntimeEvent::Shutdown,
+                }
+            },
```
bit-aloo: We should design this change around errors, since in our implementation, errors define the lifetime of components, whether they should shut down, fall back, or disconnect. I think this is very similar to fallback. With fallback, we move to the next upstream in the list; here, we reconnect to the same upstream. None of our systems currently handle this case, but the core design should be similar to reconnecting to a previously available upstream that we skipped because it was offline or unavailable for some other reason.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -685,16 +693,29 @@ impl Sv1Server {
...
             }
+            let message_was_queued =
+                self.with_registered_downstream(downstream_id, |downstream| {
+                    downstream
+                        .downstream_data
+                        .with(|data| {
+                            if data.queued_sv1_handshake_messages.len()
+                                >= MAX_QUEUED_SV1_HANDSHAKE_MESSAGES
+                            {
+                                return false;
```
bit-aloo: We can return error from here, directly

bit-aloo (commented):
I need to think around the reconnect design.

miner-apps/translator/src/lib/translator_runtime.rs
GitGab19: I removed the `480b6c0b fix(translator): handle channel endpoint changes without panicking` and `51e27cb1 fix(translator): reject reconnect requests without panicking` commits from this PR, so that we're gonna properly agree on the correct design and implement the fix here on tProxy as well later.

We have https://github.com/stratum-mining/sv2-apps/issues/780 which is tracking exactly this, so let's open the discussion there.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: The early return stopped the broadcast at the first disconnected downstream, so healthy downstreams later in the iteration did not receive the new difficulty. We now complete the broadcast and then report the disconnected peer for cleanup. The regression test covers this case explicitly.

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: Done

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Done

GitGab19:
Just pushed three more commits to cover the new #811, #812, and #813 found by Loupe.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
bit-aloo: With the current set of changes, we would only consider the first downstream in the queue. If there are multiple downstreams, we would only return an error for the first one. What we have here is better. Although it doesn’t disconnect every downstream that errors out, it at least gives all of them a chance to broadcast the message. I was thinking of designing this similarly to joins, where each of these would run concurrently instead of sequentially as they do now. Once all the downstream operations complete, we would send the appropriate disconnection requests.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -81,6 +81,29 @@ impl HandleMiningMessagesFromServerOwnedAsync for ChannelManager {
             })?
             .1;
 
+        // In non-aggregated mode, upstream assigns both IDs used in the shared routing namespace.
+        // Reject collisions before they can overwrite or reinterpret live channel state.
+        if !self.mode.is_aggregated()
+            && (m.channel_id == m.group_channel_id
```
bit-aloo: This one should be ok?

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -1607,99 +1632,126 @@ impl Sv1Server {
...
-                            .with_mut(&ch_id, |jobs| jobs.push(keepalive_notify.clone()));
-                    }
+        Ok(())
+    }
 
-                    Some(keepalive_notify)
-                });
+    fn create_keepalive_job(
+        &self,
+        channel_id: Option<ChannelId>,
```
bit-aloo: Should channel_id be an option?

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: What do you mean?

bit-aloo: message channel_id and group_channel_id can be the same

GitGab19: They cannot be the same, as they both share the same connection namespace, from the specs.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Removed the option.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Good point, but I'd keep this PR minimal. This PR only fixes the starvation (the actual Loupe finding).

Also, the returned `Disconnect` isn't the only cleanup path, given that each downstream task runs its own `on_disconnect` when its receiver closes, and a failed send usually means that receiver already closed, so cleanup should still be done there.

The broader "who owns fan-out cleanup" question spans tProxy, Pool and JDC, and making only this tProxy function do multi-peer cleanup would make it inconsistent with the other two. That's already tracked in #777, so I'd leave standardization there.

WDYT?

bit-aloo: For sure, I was just mentioning I did thought about this problem before. For now, this is ok.

integration-tests/tests/bitcoin_core_ipc_jdp_io.rs
```diff
@@ -6,8 +6,8 @@
 //! - `DeclareMiningJob` returns `Error(stale-chain-tip)` when the declared BIP34 height is
 //!   intentionally mismatched.
 //! - `DeclareMiningJob` does not retain client-supplied transactions when validation fails.
-//! - `DeclareMiningJob` rejects a coinbase that does not carry exactly one input, without
-//!   tearing down the IPC connection.
+//! - `DeclareMiningJob` rejects a coinbase that does not carry exactly one input, without tearing
+//!   down the IPC connection.
```
plebhash, line 10: is this here on purpose?

miner-apps/translator/README.md
```diff
@@ -95,8 +95,10 @@ Make sure the machine running the Translator Proxy has its clock synced with an
   address as the username.
 
 #### **Solo/Donation Payout Verification**
-Payout verification is disabled by default. Set `verify_payout = true` for solo mining or
-donation configurations where `user_identity` intentionally encodes an on-chain payout address:
+Payout verification defaults to `false` for compatibility with standard pool mining. In that mode,
+tProxy trusts the upstream's coinbase payout policy and does not verify that jobs pay the miner.
+Set `verify_payout = true` for solo mining or donation configurations where `user_identity`
+intentionally encodes an on-chain payout address:
```
plebhash: small rewording to improve readability

```suggestion
Payout verification defaults to `false` for compatibility with standard pool mining. When payout verification is disabled,
tProxy trusts the upstream server's coinbase payout policy and does not verify that jobs pay the miner.
Set `verify_payout = true` for solo mining, or donation configurations where `user_identity`
intentionally encodes an on-chain payout address:
```

plebhash:
https://github.com/stratum-mining/sv2-apps/pull/741/commits/d713bdce85acd9f5c13a92f72847bb741c96eced seems to be going in a different direction than what the scope of #811 proposes

no hard opinions on which direction is best, just pointing out the contradiction in case its not intentional

plebhash:
commits have titles but no message bodies

as described on [this link](https://chris.beams.io/posts/git-commit/) listed on `CONTRIBUTING.md`, we should start enforcing this on all commits, so that we always have in-tree documented rationales for all code changes

(highly recommended to have this on `AGENTS.md`/`CLAUDE.md`)

GitGab19:
> [d713bdc](https://github.com/stratum-mining/sv2-apps/commit/d713bdce85acd9f5c13a92f72847bb741c96eced) seems to be going in a different direction than what the scope of #811 proposes
> 
> no hard opinions on which direction is best, just pointing out the contradiction in case its not intentional

Yeah it's intentional, but thanks for pointing it out.

> commits have titles but no message bodies

Completely agree, I'll add bodies soon.

miner-apps/translator/README.md
GitGab19: Done

integration-tests/tests/bitcoin_core_ipc_jdp_io.rs
GitGab19: This was `cargo fmt`

plebhash (approved):

GitGab19:
I just modified the keepalive jobs commit: [3dafaf3a](https://github.com/stratum-mining/sv2-apps/commit/3dafaf3ae2dfb1772f33d13a6cf74507d2ba9e2d).

Keepalive jobs now go only to miners whose own interval has elapsed, reusing the shared job when no new mutation is due. Late joiners also receive the latest shared job without resetting its history or rewinding nTime.

plebhash (approved):

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -578,69 +566,333 @@ impl Downstream {
...
+        Ok(())
+    }
+
+    async fn send_cached_handshake_messages(
+        &self,
+        set_difficulty: Option<json_rpc::Message>,
+        notify: Option<json_rpc::Message>,
+    ) -> TproxyResult<(), error::Downstream> {
+        let mut did_send_difficulty = false;
+        if let Some(set_difficulty) = set_difficulty {
```
bit-aloo, line 627: Can we make this method start with processing when set difficulty is none, and the some case become the natural flow of the method. Will become more readbale.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -81,6 +81,38 @@ impl HandleMiningMessagesFromServerOwnedAsync for ChannelManager {
...
+            || m.group_channel_id == AGGREGATED_CHANNEL_ID
+            || m.channel_id == m.group_channel_id;
+        // In non-aggregated mode, wire channel IDs are stored directly in the local maps, so also
+        // reject collisions with any live channel or group before state can be overwritten.
+        if reserved_or_self_collision
+            || (!self.mode.is_aggregated()
+                && (self.extended_channels.contains_key(&m.channel_id)
+                    || self.group_channels.contains_key(&m.channel_id)
+                    || self.extended_channels.contains_key(&m.group_channel_id)))
+        {
```
bit-aloo: This conditioning maybe we can have this as part of a method,  otherwise somewhat hard to read

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -578,69 +566,333 @@ impl Downstream {
...
+                        return None;
+                    };
+                    Some((data.cached_set_difficulty.take(), Some(notify)))
+                })
+                .map_err(TproxyError::shutdown)?;
+
+            let Some((set_difficulty, notify)) = next_messages else {
+                break;
+            };
+            self.send_cached_handshake_messages(set_difficulty, notify)
```
bit-aloo: Why are we sending cahced handshake message again? shouldn't one be enough? rest can be discarded?

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -63,6 +56,16 @@ impl DownstreamIo {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum Sv1HandshakeState {
+    /// The miner has not completed both `mining.subscribe` and `mining.authorize`.
+    Pending,
+    /// Cached setup notifications are being delivered in protocol order.
+    Completing,
```
bit-aloo, line 65: Can you explain where we use Completing an why?

bit-aloo: And why we moved it away from Atomic we had for this earlier?

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -1574,132 +1640,175 @@ impl Sv1Server {
...
+            .unwrap_or_else(|| last_job.job_id.clone());
+        let base_time = self
+            .get_original_job(&original_job_id, channel_id)
+            .as_ref()
+            .map(|job| job.time.0)
+            .unwrap_or(last_job.time.0);
+
+        // Keep nTime within Bitcoin Core's maximum future-time policy relative to the original
+        // upstream job.
+        const MAX_FUTURE_BLOCK_TIME: u32 = 2 * 60 * 60;
```
bit-aloo: lets define ait at the top of the module

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -1574,132 +1640,175 @@ impl Sv1Server {
...
+            Some(sender) => sender.send(notify.into()).await.is_ok(),
+            None => false,
+        };
+        if !sent {
+            warn!(
+                "Failed to send keepalive job to downstream {}",
+                downstream_id
+            );
+            return Ok(());
+        }
```
bit-aloo:
```suggestion
let Some(sender) = self
    .sv1_server_io
    .sv1_server_to_downstream_sender
    .get_cloned(&downstream_id)
else {
    warn!("No sender found for downstream {}", downstream_id);
    return Ok(());
};

if sender.send(notify.into()).await.is_err() {
    warn!("Failed to send keepalive job to downstream {}", downstream_id);
    return Ok(());
}
```

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -1574,132 +1640,175 @@ impl Sv1Server {
...
                     };
 
-                    if let Some(ch_id) = job_channel_id {
-                        // Use with_mut (not with_mut_or_default) so we never
-                        // re-create a valid_sv1_jobs entry for a channel that was
-                        // already cleaned up.
-                        self.valid_sv1_jobs
-                            .with_mut(&ch_id, |jobs| jobs.push(keepalive_notify.clone()));
+                    if !needs_keepalive {
+                        return None;
```
bit-aloo:
```suggestion
let Some(last_time) = d.last_job_received_time else {
    return None;
};

if last_time.elapsed() < interval {
    return None;
}
```

bit-aloo (commented):
Few questions and suggestions..

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: These aren’t the same messages being sent again. The first pair is removed with `.take()`, but another `mining.notify` can be cached while we’re sending it. The loop flushes that newer work before enabling normal forwarding. Discarding it could leave the miner on an outdated job.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: Yeah this is something you also pointed out on #823, I'll add a helper for this

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: `Completing` means that one caller is flushing the cached notifications, while normal forwarding is still disabled. Setting the old flag before the flush lets newer notifications overtake it; setting it afterward needs coordination with messages cached during the flush. Keeping the state and cache under the same lock handles that transition and prevents repeated completion.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Sure, I'll do that

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Yes, I'll apply this suggestion soon

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Yes, I'll apply this suggestion soon

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: I kept the `if let` here because we still need to send the cached notify when there's no difficulty. Handling `None` first would duplicate the notify path.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: Done, moved the checks into `colliding_channel_id`

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: I also renamed the helper to `send_cached_mining_notifications` and clarified in the docs that these aren't repeated handshake responses.

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: Expanded the docs to explain this: `Completing` keeps normal forwarding disabled while one caller flushes the cache, with the state and cache coordinated under the same lock.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Done

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Done

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Done, using `?` for the missing timestamp and returning early when the miner's interval hasn't elapsed.

bit-aloo (approved):
ACK. Clanker suggested some non blocker, but I don't think they were interesting to block work here.

Issue #777, opened 2026-08-20
clean up channel and downstream state on every terminal path


GitGab19:
## Affected applications

- tProxy
- JDC
- Pool
- JDS audit

## Problem

Channel and downstream identifiers are duplicated across several maps: request correlation, channel ownership, job/template mapping, vardiff state, share-sequence counters, group membership, and sender registries.

Some disconnect, close, tip-activation, and terminal-response paths remove only part of this state. JDC can retain stale per-channel job-to-template mappings after a new tip is activated. In tProxy, closing a non-aggregated channel can retain its share-sequence counter, and removing the last member of a group can leave an empty group behind.

Fan-out paths also handle failed downstream sends inconsistently. They generally continue delivery to healthy recipients, but explicit cleanup can cover only one failed recipient or be left entirely to the downstream task's independent disconnect callback. This leaves cleanup ownership unclear when several recipients fail during the same broadcast.

Existing tracker issues report similar stale-state behavior in tProxy, JDC, and Pool.

## Expected behavior

Every owner should have one defined cleanup operation that removes all state reachable from that owner. Cleanup should be idempotent and invoked for disconnects, close messages, tip changes, failed opens, terminal errors, fallback, and shutdown.

## Acceptance criteria

- Inventory per-downstream, per-channel, per-job, and per-group maps in each application.
- Define a single cleanup boundary for each owner.
- Remove pending requests, jobs, templates, vardiff state, share counters, empty groups, senders, and allocator resources together.
- Prune stale per-channel job mappings when a new tip makes them unreachable.
- Preserve intentionally shared aggregate-channel state when an individual downstream disconnects.
- Ensure failed per-downstream sends during fan-out do not prevent delivery to healthy recipients, and route every failed recipient through one defined cleanup path.
- Make cleanup safe when called more than once.
- Prevent late responses from recreating state for a removed owner.
- Audit JDS for equivalent ownership gaps.
- Add tests for every terminal path, tip activation, last-member removal, repeated cleanup, and multi-recipient fan-out failure.

### JDC

- project-loupe/audit-sv2-apps#75
- project-loupe/audit-sv2-apps#175

### tProxy

- project-loupe/audit-sv2-apps#190
- project-loupe/audit-sv2-apps#191

[...]

PR #823, opened 2026-08-27
harden tProxy Sv1 session and channel lifecycle

### LINKED ISSUE #36: tProxy: Implement ` handle_extranonce_subscribe`
Currently the Sv1 `mining.extranonce_subscribe` is not implemented on our tProxy, we should do it.

Context: https://github.com/stratum-mining/stratum/pull/1791#discussion_r2327878046

### LINKED ISSUE #139: tProxy's`handle_set_extranonce_prefix` is wrong

### LINKED ISSUE #164: non-aggregated tProxy does not close Sv1 connection after receiving `CloseChannel`
currently, we simply remove the channel from the `HashMap`, which correctly stops submission of Sv2 shares for this channel

https://github.com/stratum-mining/sv2-apps/blob/f080061b288381a15c3283413e0eb882451cb192/miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs#L289-L306

on #156 we are simply expanding this naive approach to also account for Group Channels

while writing [`non_aggregated_translator_correctly_deals_with_close_channel_message` Integration Test](https://github.com/plebhash/sv2-apps/blob/134a4d15954d32e9dc847cc76249000b039b2f09/integration-tests/tests/translator_integration.rs#L959), I noticed that Sv1 shares keep coming to tProxy

this test has 3 extended channels (with ids 0, 1, and 2), under a group channel with id 100

```
2026-01-05T23:22:49.492121Z  INFO translator_sv2::sv2::channel_manager::mining_message_handler: Received: CloseChannel(channel_id: 0, reason_code: )
...
2026-01-05T23:22:54.501667Z  INFO translator_sv2::sv2::channel_manager::mining_message_handler: Received: CloseChannel(channel_id: 100, reason_code: )
[2026-01-05 20:22:54] thread 0: 581076 hashes, 1704 khash/s
2026-01-05T23:22:54.830849Z  INFO translator_sv2::sv1::downstream::message_handler: Received mining.submit from SV1 downstream for channel id: 2
[2026-01-05 20:22:54] accepted: 18/18 (100.00%), 1704 khash/s (yay!!!)
[2026-01-05 20:22:55] thread 0: 450876 hashes, 1617 khash/s
2026-01-05T23:22:55.109807Z  INFO translator_sv2::sv1::downstream::message_handler: Received mining.submit from SV1 downstream for channel id: 2
[2026-01-05 20:22:55] accepted: 19/19 (100.00%), 1617 khash/s (yay!!!)
[2026-01-05 20:22:55] thread 0: 443432 hashes, 1699 khash/s
2026-01-05T23:22:55.370753Z  INFO translator_sv2::sv1::downstream::message_handler: Received mining.submit from SV1 downstream for channel id: 2
[2026-01-05 20:22:55] accepted: 20/20 (100.00%), 1699 khash/s (yay!!!)
[2026-01-05 20:22:55] thread 0: 2317869 hashes, 1679 khash/s
2026-01-05T23:22:55.420974Z  INFO translator_sv2::sv1::downstream::message_handler: Received mining.submit from SV1 downstream for channel id: 1
[2026-01-05 20:22:55] accepted: 21/21 (100.00%), 1679 khash/s (yay!!!)
[2026-01-05 20:22:55] thread 0: 2640461 hashes, 1683 khash/s
2026-01-05T23:22:55.624309Z  INFO translator_sv2::sv1::downstream::message_handler: Received mining.submit from SV1 downstream for channel id: 0
```

as we can see, Sv1 shares continue arriving for all channels, even though we sent `CloseChannel` messages directed to channel 0 and to group channel 100

ideally, those Sv1 connections should be killed as a result of `CloseChannel`

### LINKED ISSUE #386: tProxy needs a state machine to track each sv1 message sent by downstream clients before upstream channel opens
on #382 we applied a hotpatch for sv1 handshake that @Shourya742 feels needs to be elaborated

@Shourya742 via https://github.com/stratum-mining/sv2-apps/issues/382#issuecomment-4147656939
> This commit f77095f, worked on my end. Give it a try, not a polished solution, but fixes the sv1 handshake issue.

also @Shourya742 via Discord DM:
> Via polishing I meant having a state machine to track each sv1 message sent by downstream clients before upstream channel opens...

---

ref: https://github.com/stratum-mining/sv2-apps/issues/382#issuecomment-4147791175

### LINKED ISSUE #650: tProxy silently discards 100% of shares against pools with version_rolling_allowed: false
Two distinct defects, one of which is only visible because of the other. Found with real ASIC hashrate against a third-party production SV2 pool that advertises `version_rolling_allowed: false`.

**Net effect: 100% of shares are discarded inside the tProxy, while the SV1 miner is told every share was accepted.**

## Defect 1 (root cause) — offered version-rolling mask ignores upstream capability

`miner-apps/translator/src/lib/sv1/sv1_server/downstream_message_handler.rs:39-41` answers `mining.configure` by granting the SV1 downstream:

```rust
data.version_rolling_mask = request
    .version_rolling_mask()
    .map(|mask| HexU32Be(mask & 0x1FFFE000));
```

This never consults whether the upstream pool permits version rolling. When the upstream sends `NewExtendedMiningJob { version_rolling_allowed: false }`, the miner has already been authorised to roll `0x1FFFE000`, and share validation rejects exactly those bits — `channels-sv2/src/client/extended.rs:590-598`:

```rust
if !job.0.version_rolling_allowed {
    if (share.version & 0x1fffe000) != 0 {
        return Err(ShareValidationError::VersionRollingNotAllowed(...));
    }
}
```

The granted mask **is** the rejected bit range, bit for bit, so the failure rate is 100% rather than probabilistic.

## Defect 2 — the drop site logs nothing

`miner-apps/translator/src/lib/sv2/channel_manager/mod.rs:604` (aggregated) and `:647` (non-aggregated) discard every `ShareValidationError` through a bare `else`:

```rust
let value = self.extended_channels.get_mut(&m.channel_id)
    .map(|mut extended_channel| extended_channel.validate_share(m.clone()));
if let Some(Ok(_result)) = value {
    info!("SubmitSharesExtended: valid share, forwarding it to upstream ...");
    ...
} else {
    return Ok(());   // <-- Err(_) and None both land here, unlogged at any level
}
```

`Some(Err(_))` (validation failed) and `None` (no such channel) are indistinguishable, and neither is reported. Meanwhile the SV1 path has already responded `{ result: true }` to the miner, so from the miner's perspective every share was accepted.

Notably, the other roles do handle this error explicitly — `pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs` matches `Err(ShareValidationError::VersionRollingNotAllowed(code))`, and jd-client does the same in two places. **The translator is the only role that discards share-validation errors silently.**

## Reproduction

SV1 ASIC → tProxy (non-aggregated) → SV2 pool advertising `version_rolling_allowed: false` on extended jobs.

Adding a diagnostic at the drop site (patch below) produces, for every submit:

```
ERROR translator_sv2::sv2::channel_manager: SubmitSharesExtended: share REJECTED by
local validation (non-aggregated) | channel_id: 1,
error: VersionRollingNotAllowed("version-rolling-not-allowed")
```

Measured, same binary and same config file, changing only `upstream_address`:

| | pool with `version_rolling_allowed: false` | pool with `true` |
|---|---|---|
| `mining.submit` received | 1028 | 480 |
| forwarded upstream | **0** | 138 |
| `VersionRollingNotAllowed` | **1028 (100%)** | 0 |
| `DoesNotMeetTarget` | 0 | 342 |

The negotiated mask logged identically in both runs (`Negotiated version_rolling_mask: Some(HexU32Be(536862720))` = `0x1FFFE000`), so the miner's behaviour is constant; the only variable is the upstream's `version_rolling_allowed`. The `DoesNotMeetTarget` rejections in the right-hand column are expected in that setup (a rate-shaping proxy sits downstream) and are shown only to demonstrate the diagnostic reports real rejections rather than manufacturing them.

## Versions

Both defects present on `main` as of this writing — sv2-apps `44af69e5`, stratum `c1a79913`.

## Possibly related

#402 includes a report of a very high share-rejection rate against Braiins Pool via non-aggregated tProxy. That may share this root cause, but it is not the same observation: those shares reached the pool and were rejected there, whereas these never leave the translator. Flagging in case the two turn out to be connected.

## Suggested direction

1. **Mask negotiation** — clamp the offered mask to `0` when the upstream forbids version rolling. This is awkward as written, because `mining.configure` must be answered before the first `NewExtendedMiningJob` reveals the upstream's capability. Options include deferring the `mining.configure` reply until the first job is known, or issuing `mining.set_version_mask` to revise the grant once capability is learned.
2. **Diagnostics** — log every `ShareValidationError` at both drop sites, distinguishing `Some(Err(_))` from `None`. Minimal patch that produced the evidence above:

```rust
} else {
    match value {
        Some(Err(ref e)) => error!(
            "SubmitSharesExtended: share REJECTED by local validation (non-aggregated) | channel_id: {}, error: {:?}",
            m.channel_id, e
        ),
        None => error!(
            "SubmitSharesExtended: channel not found for local validation (non-aggregated) | channel_id: {}",
            m.channel_id
        ),
        Some(Ok(_)) => {}
    }
    return Ok(());
}
```

3. **Worth considering separately:** whether the SV1 `{ result: true }` should be deferred until the share is actually accepted upstream. Answering `true` for a share that is about to be discarded locally makes this class of bug invisible from the miner side, and corrupts any hashrate accounting derived from SV1 accept counts.

Happy to open a PR for (2), and for (1) if there's a preferred direction on where the capability check should live.

### LINKED ISSUE #831: tProxy: validate the first SV1 request before opening an upstream channel
## Affected applications and crates

- tProxy SV1 downstream admission and channel-open path.

## Problem

For a new SV1 connection, tProxy requests an upstream mining channel before parsing and validating the peer's first JSON-RPC message.

An unauthenticated peer can send a malformed, unsupported, or invalid first request and still cause tProxy to allocate pending state and transmit an upstream channel-open request before the SV1 session is known to be usable.

## Impact

Cheap invalid downstream traffic can consume pending-open state, upstream channel capacity, extranonce space, and request-processing bandwidth, reducing availability for legitimate miners.

## Expected behavior

A downstream must make a valid, permitted initial SV1 state transition before tProxy allocates an upstream mining channel on its behalf.

## Acceptance criteria

- Parse the first frame and validate its JSON-RPC shape, method, and parameters before requesting a channel.
- Define which initial SV1 methods and orderings are accepted.
- Do not allocate pending channel-open state or send an upstream open for malformed, unsupported, or wrong-state messages.
- Ensure rejection and disconnect clean up all per-peer state.
- Preserve compatibility with valid configure/subscribe/authorize sequences.
- Add tests asserting that invalid first messages emit no upstream open request.
- Add tests for malformed JSON, unsupported methods, invalid parameters, repeated initial messages, and a valid session.

## Loupe findings

- [Loupe #227](https://github.com/project-loupe/audit-sv2-apps/issues/227)

## Related sv2-apps issues and PRs

- https://github.com/stratum-mining/sv2-apps/issues/386 tracks the broader SV1 session state machine.
- https://github.com/stratum-mining/sv2-apps/issues/743 bounds pre-channel message buffering.
- https://github.com/stratum-mining/sv2-apps/issues/756 bounds pending upstream channel opens.
- https://github.com/stratum-mining/sv2-apps/pull/823 already includes validation of the initial SV1 request before requesting an upstream channel, reusing `sv1_api` parsing and parameter validation. The PR is still open; this fix is not yet on `main`.

## Separate hardening trackers

This issue is scoped to validation before channel opening. The following resource limits and their regression tests are tracked separately and are not acceptance criteria here:

- https://github.com/stratum-mining/sv2-apps/issues/776: admission/handshake deadlines, idle connections, and connection limits.
- https://github.com/stratum-mining/sv2-apps/issues/817: malformed-SV1 processing budgets and bounded error logging.
- https://github.com/stratum-mining/sv2-apps/issues/789: bounded inbound/outbound SV1 connection queues.
- https://github.com/stratum-mining/sv2-apps/issues/768: internal application queue bounds.

## Intended PR grouping

This belongs in a tProxy-local PR and may be grouped with other compatible SV1 admission fixes using a separate commit.

### LINKED ISSUE #832: tProxy: invalidate downstream jobs when an advertised clean job supersedes them
## Affected applications and crates

- tProxy aggregated and non-aggregated SV1 job validation.

## Problem

A downstream-specific difficulty transition can cause tProxy to advertise a `mining.notify` with `clean_jobs = true`, while retained validation state continues to use the original upstream job's `clean_jobs = false` history.

After an easier target is applied, an attacker can submit work for a job the downstream was explicitly told to discard. The old work may now satisfy the easier local target and be acknowledged or forwarded even though the advertised clean-job transition invalidated it.

## Impact

Stale or invalidated shares can be accepted locally, forwarded upstream, or included in incorrect accounting. Honest miners and tProxy can disagree about which jobs are active.

## Expected behavior

The validation job set for each downstream must exactly match the job lifecycle and `clean_jobs` semantics that tProxy advertised to that downstream.

## Acceptance criteria

- Record the effective `clean_jobs` value actually sent to each downstream.
- When a downstream receives `clean_jobs = true`, invalidate every older job for that downstream before accepting further shares.
- Preserve job-specific target, extranonce, version-rolling, and routing state.
- Do not let a later target change make an invalidated job eligible again.
- Keep one downstream's clean transition from deleting valid jobs belonging to another downstream.
- Apply the invariant in aggregated and non-aggregated modes.
- Bound retained job history independently of invalidation correctness.
- Add tests for old-job replay before and after easier and harder difficulty transitions.
- Verify rejected stale shares receive the correct SV1 error and are never forwarded.

## Loupe findings

- [Loupe #228](https://github.com/project-loupe/audit-sv2-apps/issues/228)

## Related sv2-apps issues and PRs

- #823 contains an open commit that introduces bounded job-specific validation state and avoids treating ordinary difficulty updates as clean jobs. That is candidate remediation, not a fix on `main`; its final form must explicitly satisfy this issue's invalidation tests.
- #650 concerns version-rolling authorization and share-error reporting, not clean-job invalidation.

## Intended PR grouping

This belongs in a tProxy-local PR. Prefer integrating the scoped fix into #823 if that PR remains the active lifecycle stack, with a separate commit and explicit Loupe #228 regression coverage.

GitGab19:
This PR is built on top of #741 and:

- Tracks SV1 setup progress and enables job notifications after both `mining.subscribe` and `mining.authorize` complete.
- Requires upstream jobs to support version rolling.
- Disconnects SV1 miners when the upstream closes their channel.
- Applies extranonce prefix changes without recreating channels.
- Honors `mining.extranonce.subscribe`, notifying supported miners and disconnecting unsupported ones.
- Handles malformed mining notifications without panicking.
- Starts vardiff only after the downstream SV2 channel is open.
- Retains bounded, job-specific validation state so late SV1 shares use the target and extranonce assigned to their job without treating difficulty updates as clean jobs.
- Returns standard SV1 rejection errors for missing jobs, duplicate shares, and low-difficulty shares, with bounded per-downstream duplicate tracking.
- Logs unexpected upstream-channel validation failures while keeping expected target filtering at debug level.

Closes #36
Closes #139
Closes #164
Closes #386
Closes #650

Closes stratum-mining/sv2-apps#831
Closes project-loupe/audit-sv2-apps#227

Closes stratum-mining/sv2-apps#832
Closes project-loupe/audit-sv2-apps#228


miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -464,9 +498,205 @@ impl HandleMiningMessagesFromServerOwnedAsync for ChannelManager {
...
+                    continue;
+                };
+
+                let mut downstream_update = m.clone();
+                downstream_update.channel_id = channel_id;
+                downstream_update.extranonce_prefix =
+                    updated_prefix?.try_into().map_err(TproxyError::shutdown)?;
+                self.channel_manager_io
+                    .sv1_server_sender
+                    .send(MiningOwned::SetExtranoncePrefix(downstream_update))
```
plebhash: fable clanker finding:

> forwarding the per-channel `SetExtranoncePrefix` to the sv1 server immediately opens a window where a subscribed miner silently mines garbage:
>
> 1. upstream sends future job J (created under the old prefix, captured per-job by `channels_sv2`)
> 2. upstream sends `SetExtranoncePrefix` → this fan-out queues `mining.set_extranonce` to the miner
> 3. upstream sends `SetNewPrevHash` activating J → the miner mines J with the **new** extranonce1
> 4. `record_job_validation_context` mapped J to the new extranonce1 (the set_extranonce event precedes J's notify in the per-downstream FIFO), so sv1-side validation reconstructs what the miner hashed and answers `result: true`
> 5. channel-level validation reconstructs J's coinbase with its captured **old** prefix → merkle mismatch → the share dies in the unlogged `else` branch of the submit path
>
> so the miner sees 100% accepted while nothing reaches the upstream, until the first job actually created under the new prefix. the rustdoc on `aggregated_jobs_use_current_prefix` already names exactly this job class ("immediately for the active job, at the next block for a future one") but only guards late joiners.
>
> suggestion: defer this fan-out behind the same check — hold the per-channel updates pending while `!aggregated_jobs_use_current_prefix()`, flushed from the same three call sites that flush `pending_downstream_channels`, so `mining.set_extranonce` enters each miner's FIFO right before the first current-prefix notify (ideally with `clean_jobs: true` forced on it). old-job shares then stay valid end to end (old extranonce1 mapping + captured old prefix, on both tProxy layers and upstream), and it also shrinks the reject burst for firmwares that apply the new extranonce1 immediately instead of at the next job. the non-aggregated arm below has the same issue and can reuse the same pending/flush pattern per channel.

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
```diff
@@ -644,31 +645,19 @@ impl ChannelManager {
...
+                        // so this naturally leaves no local bytes to prepend.
+                        let local_prefix_and_index =
+                            self.extended_channels.with(&m.channel_id, |c| {
+                                c.get_extranonce_prefix()[c.upstream_prefix_len() as usize..]
+                                    .to_vec()
+                            });
+                        if let Some(local_prefix_and_index) =
+                            local_prefix_and_index.filter(|prefix| !prefix.is_empty())
+                        {
+                            let mut new_extranonce = local_prefix_and_index;
```
plebhash: fable clanker finding:

> just below this hunk, the validation-failure path is still
>
> ```rust
> } else {
>     return Ok(());
> }
> ```
>
> both here (L665) and in the aggregated arm (L630). `Some(Err(_))` (share rejected by local validation) and `None` (channel not found) are indistinguishable and neither is logged, while the sv1 side has already answered `result: true` to the miner.
>
> the version-rolling guard in this PR removes the `VersionRollingNotAllowed` case from #650, but every other `ShareValidationError` (`DoesNotMeetTarget`, stale job, duplicate) still vanishes here — and the failure modes around extranonce-prefix transitions land exactly in this branch, which is what makes them hard to observe in the field. defect 2 of #650 includes a ready-made logging patch for these two sites.
>
> since defect 2 (this) and defect 3 (`result: true` sent before validation) both survive this PR, I'd also drop `Closes #650` from the description — or split what remains into a follow-up issue — so it doesn't auto-close half-fixed.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: Fixed, now the `mining.set_extranonce` message is queued and immediately sent before the new `mining.notify`.

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
GitGab19: Added some logs in both arms (aggregated and not). 

In case of an error, now the Sv1Server answers back to the Sv1 miner with the right error code.

bit-aloo:
Clanker review:

# Correctness issues

## 1. `mining.set_extranonce` can be silently lost during the subscribe window

**Locations:** `downstream.rs:650–700`, `sv1_server/mod.rs:794`

The downstream task decides whether to forward the notification using `session_state.is_subscribed()` **at processing time**. However, that flag is flipped by the server task only after the subscribe response has been built and sent.

The race looks like this:

1. The server builds the subscribe response with the old `extranonce1`.
2. The server sends the response.
3. The downstream task processes the queued `SetExtranonce` event.

   * `is_subscribed == false`
   * It updates `extranonce1` silently.
   * It drops the cached notification.
4. The server calls `record_response(Subscribe)`.

At this point, the miner has the **old** `extranonce1`, while every new job context is recorded with the **new** one. All shares from that miner then fail hash validation until it reconnects.

The window is only microseconds, but it creates a correctness hole in exactly the path this PR is trying to harden.

**Suggested fix:** Make the decision in the server task at enqueue time in `handle_upstream_extranonce_change`. This is the same task that runs `record_response`, so there is no race.

For example:

```text
Sv1ServerEvent::SetExtranonce {
    message,
    notify_miner: bool,
}
```

When the miner is not yet subscribed, eagerly apply `extranonce1` / `extranonce2_len` under the lock so the upcoming subscribe response carries the new value. The event then only preserves FIFO ordering: drop `cached_notify` and decrement the pending counter.

When the miner is already subscribed, the subscribe response was sent before the event was enqueued, so FIFO guarantees the correct wire order.

---

## 2. Aggregated bootstrap job rewinds the shared keepalive history

**Locations:** `sv1_server/mod.rs:1150–1195`, `sv1_server/mod.rs:1880`

A late joiner's bootstrap job (child `channel_id`) is `activate()`d into the shared `AGGREGATED_CHANNEL_ID` store with `clean_jobs = false`.

Suppose the active job is:

```text
orig#3  (nTime = T0 + 3i)
```

The bootstrap `orig` replaces it as active, while `orig#3` moves to past.

The next `create_keepalive_job` then derives from `orig` and emits:

```text
orig#4  (nTime = T0 + i)
```

to **every miner**.

That's the exact slice miners already exhausted under `orig#1`. They re-search it, and their own `accepted_share_hashes` cache rejects any rediscovered shares as duplicates.

This was latent before—the old `Vec::last()` had the same shape—but the new code explicitly special-cases bootstrap jobs, so this seems like the right place to finish fixing it.

**Suggested fix:** When:

```text
m.channel_id != AGGREGATED_CHANNEL_ID
```

and the store already has an active job whose original ID equals `notify.job_id`, don't mutate the shared store. Instead, send the late joiner the **current stored active notify** with `clean_jobs = true`.

---

## 3. Late aggregated joiners can hang at `mining.subscribe` for up to a block

**Locations:** `channel_manager/mod.rs:1052–1070`, `channel_manager/mod.rs:1252`

The deferral itself is correct, and the doc comment is honest about the cost: **"next block plus one job."**

The problem is that SV1 miners typically time out a pending subscribe after ~30–60 seconds and reconnect.

During a prefix transition, this can produce a loop:

```text
subscribe
  → deferred
  → miner times out
  → reconnects
  → subscribe
  → deferred
  → ...
```

When the compatible job finally arrives, N child channels may be opened and immediately closed through:

```text
DownstreamNotPresent → CloseChannel
```

This isn't a leak, but it creates unnecessary churn, and the miner sees a connection failure rather than simply a delayed setup.

### Possible no-wait alternative

The required building blocks already exist:

1. Allocate the child prefix.
2. Call `set_upstream_prefix(old_bytes)` on it.
3. Create the `ExtendedChannel`.
4. Replay the aggregated channel's active/future jobs, each with its captured prefix, switching the child's upstream region before each replay.
5. Call `set_upstream_extranonce_prefix(new)`.

The subscribe response then advertises the old variant, and `forward_job_to_sv1_server` already emits `mining.set_extranonce` on the first new-prefix job.

**Tradeoff:** A miner that doesn't support `set_extranonce` gets one disconnect later instead of waiting now.

Either approach seems reasonable, but it's worth making the choice explicit. If the deferral stays, I'd at least add a bounded timeout that rejects the request with an error so the miner fails fast.

---

## 4. A supporting miner can be disconnected mid-handshake

**Location:** `downstream.rs:686`

The disconnect currently fires when:

```text
is_subscribed && !supports_set_extranonce
```

However, `mining.extranonce.subscribe` commonly arrives **after** `mining.subscribe`, and for some firmware it can even arrive after `authorize`.

A prefix change landing in that window therefore disconnects a miner that would have supported the change once setup completed.

**Suggested fix:** While `Starting`, cache the `set_extranonce` and re-evaluate it at `SetupComplete`. At that point, either send it after the setup responses or disconnect if the miner still doesn't support it.

# Ergonomics / simplification

## 5. `Sv1ServerEvent` should carry typed messages, not JSON

**Locations:** `downstream.rs:40`, `downstream.rs:579–650`

The server currently:

1. Builds a typed `Notify`.
2. Converts it to `json_rpc::Message`.
3. Sends it through `Sv1ServerEvent`.
4. Every downstream task re-parses it with `Notify::try_from(notification.clone())`.
5. The downstream also string-matches `notification.method`.

This happens once per miner per job.

An event enum like:

```text
Notify(Arc<Notify>)
SetDifficulty(Message)
SetExtranonce(SetExtranonce)
SetupComplete
```

would move JSON conversion to the socket-write boundary.

It would also remove:

* `InvalidMiningNotifyNotification`
* `InvalidSetExtranonceNotification`
* the `"cached mining.notify was not a notification"` branch
* two of the new unit tests that only cover unparseable internal messages

Also, `Sv1ServerEvent::notification(m)` is just a one-line alias for:

```text
Sv1ServerEvent::Notification(m)
```

so I'd drop it.

---

## 6. Duplicated extranonce-size arithmetic in `handle_set_extranonce_prefix`

**Locations:** `mining_message_handler.rs:529`, `mining_message_handler.rs:623`

Both branches manually compute:

```text
upstream + local_prefix_and_index + rollable <= MAX_EXTRANONCE_LEN
```

before calling:

```text
allocator.set_upstream_prefix(...)
channel.set_upstream_extranonce_prefix(...)
```

Those methods already perform the same validation transactionally—I verified this in the pinned `channels_sv2`.

Roughly 60 lines could therefore collapse into mapping `ExceedsMaxLength` into the appropriate fallback.

The only thing lost is the numeric:

```text
InvalidExtranonceSize {
    prefix_len,
    rollable_size,
}
```

payload, which can be reconstructed from the channel at the error site if it's still useful.

---

## 7. Collision-ID derivation is hard to read

**Location:** `mining_message_handler.rs:88–115`

The current code first checks a collision predicate, then uses another nested `if` to recover **which** ID actually collided.

A helper like:

```text
fn colliding_channel_id(&self, m) -> Option<ChannelId>
```

could return the offending ID directly.

That would read more clearly and avoid having the predicate and ID-recovery logic drift apart.

---

## 8. `handle_set_target_without_vardiff` only disconnects the first failed sender

**Location:** `sv1_server/mod.rs:1629`

Finishing the broadcast is the right fix, but if two miners are dead, the second one isn't noticed until the next `SetTarget`.

Since this already runs in the server task, calling:

```text
handle_downstream_disconnect(...)
```

inline for each failed sender seems simpler than threading a single ID through `Action::Disconnect`.

---

## 9. `Sv1AcceptedShareCache::insert_if_new` is O(4096) per share

**Location:** `downstream.rs:160`

`VecDeque::contains` scans up to 4096 32-byte hashes on every submit.

This is probably fine at current rates, but if this ever sits in front of many miners, a `HashSet` + `VecDeque` pair would give O(1) membership checks while preserving bounded eviction order.

# Nits / housekeeping

* **`utils.rs:36–62`** — The `# Arguments / # Returns` documentation for `validate_sv1_share` now attaches to `Sv1ShareValidationOutcome`, because the enum was inserted between the comment and the function. Move the enum above the doc block.

* **`bitcoin-core-sv2/Cargo.toml` / `stratum-apps/Cargo.toml`** — These point at a personal fork branch. They should be reverted once the Stratum PR lands, following the same pattern as the codec branch.

* **`bitcoin_core_ipc_jdp_io.rs`** — This is purely a doc-comment reflow and appears unrelated to this change.

* **`#[allow(clippy::result_large_err)]` on `disconnect_downstream_for_upstream_close`** — This suggests `TproxyError` has grown fairly large. Boxing the error kind is out of scope here, but probably worth a follow-up.

GitGab19:
@bit-aloo, about your review:

1. Fixed, the decision is now made in the server task which builds and records the subscribe response. Pre-subscribe changes are included in the response, and queued events cannot overwrite that prefix afterward.

2. Fixed in https://github.com/stratum-mining/sv2-apps/pull/741 and included here. Late joiners receive the latest shared job without resetting its history or rewinding nTime. Keepalives only go to miners whose own interval has elapsed.

3. Removed the deferral. New miners receive the active job’s prefix in their subscribe response, and inherited jobs are replayed with their captured prefixes. The local allocation is preserved, and later prefix changes follow the normal notification ordering.

4. We now cache the prefix change and check support when its matching job is actually delivered. This allows the miner to announce support during setup, without adding a timeout. If support is still missing when that job needs to be sent, we disconnect.

5. Done, jobs and prefix changes now use typed events. Removed the internal JSON parsing, the corresponding error variants, and the tests for malformed internal notifications.

6. Removed the duplicated non-aggregated checks. I kept the aggregate precheck because that path updates the shared allocator and multiple channels; the individual setters being transactional doesn’t make the whole operation transactional.

7. This comes from https://github.com/stratum-mining/sv2-apps/pull/741, where you also commented on it. I’d handle the readability change there and bring it here with the stack.

8. I’d keep the broader cleanup changes out of scope here. Each downstream also has its own disconnect cleanup path. The cross-application ownership question is tracked in https://github.com/stratum-mining/sv2-apps/issues/777.

9. I’d keep the bounded `VecDeque` for now, consistent with `channels_sv2`. We can revisit it if profiling shows that the lookup is becoming a bottleneck.

For the other points:

- Fixed the placement of the `validate_sv1_share` docs.
- The fork dependencies will be removed once https://github.com/stratum-mining/stratum/pull/2319 merges.
- The IPC formatting comes from https://github.com/stratum-mining/sv2-apps/pull/741 and was produced by `cargo +nightly fmt`.
- Agreed that boxing the error should be a separate change.

Added regression tests for the subscribe race, capability timing, and late-join prefix transitions.

miner-apps/translator/src/lib/sv1/job_store.rs
```diff
@@ -0,0 +1,158 @@
...
+use stratum_apps::stratum_core::channels_sv2::client::MAX_PAST_JOBS;
+
+/// Bounded active and past-job storage for the SV1 side of the translator.
+///
+/// This mirrors the client-channel lifecycle in `channels_sv2`: one active job and at most
+/// the configured cap of jobs retired under the current chain tip. SV1 future jobs need no
+/// equivalent because tProxy does not advertise them to miners before activation.
+#[derive(Debug)]
+pub(super) struct Sv1JobStore<T> {
+    active_job: Option<(String, T)>,
```
bit-aloo: Can we have named type for this tuple?

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
```diff
@@ -185,6 +196,120 @@ pub struct ChannelManager {
...
+                .sv1_advertised_extranonce_prefixes
+                .with(&channel_id, |advertised_prefix| {
+                    advertised_prefix != &job_prefix
+                })
+                .ok_or_else(|| {
+                    error!(
+                        channel_id,
+                        "Missing SV1 extranonce prefix state for active downstream channel"
+                    );
+                    TproxyError::shutdown(TproxyErrorKind::ChannelNotFound)
```
bit-aloo: Shoudlnt' this be a disconnet?

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
```diff
@@ -873,21 +1041,38 @@ impl ChannelManager {
...
+                    channel.get_extranonce_prefix().to_vec(),
+                    channel.get_active_job().cloned(),
+                    channel
+                        .get_future_jobs()
+                        .map(|(_, job)| job.clone())
+                        .collect::<Vec<_>>(),
+                    channel.get_chain_tip().cloned(),
+                )
+            })
+            .ok_or_else(|| TproxyError::shutdown(TproxyErrorKind::ChannelNotFound))?;
```
bit-aloo: Here, if the AGGREGATE_CHANNLE_ID is not present, then we are shutting down?

bit-aloo (commented):
For the ok_or_else thingys, can we have explicit `if let Some() = ` syntax, would be easier to reason about error in case of none.

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
GitGab19: Correct, given that the shared aggregate channel is installed before the aggregate state becomes `Connected`.

miner-apps/translator/src/lib/sv2/channel_manager/mod.rs
GitGab19: No, it's a shutdown because that condition is reached only if something is broken internally, as for [this case](https://github.com/stratum-mining/sv2-apps/pull/823#discussion_r4013726468)

GitGab19:
> For the ok_or_else thingys, can we have explicit `if let Some() = ` syntax, would be easier to reason about error in case of none.

Done for the reviewed points, by replacing the `ok_or_else` with explicit `let Some(...) = ... else`,  so the error behavior and invariants are easier to see.

miner-apps/translator/src/lib/sv1/job_store.rs
GitGab19: Done, I replaced the tuple with a private `ActiveJob<T>` type

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -79,6 +88,23 @@ const SV1_MIN_DIFFICULTY_FOR_INTEGER_POWER_OF_TWO_ROUNDING: f64 = 1.0;
 // Maximum keepalive nTime offset from the original upstream job.
 const MAX_FUTURE_BLOCK_TIME: u32 = 2 * 60 * 60;
 
+#[derive(Clone, Debug)]
+struct StoredSv1Job {
+    notify: server_to_client::Notify,
+    /// Original upstream job time used to cap locally generated keepalive jobs.
+    keepalive_base_time: u32,
```
bit-aloo, line 96: do we need this?considering its notify.time.0?

miner-apps/translator/src/lib/sv1/downstream.rs
```diff
@@ -24,12 +27,29 @@ use stratum_apps::{
...
+    Notify(Arc<server_to_client::Notify>),
+    SetDifficulty(json_rpc::Message),
+    SetExtranonce {
+        message: server_to_client::SetExtranonce,
+        /// Decided by the server in the same task that builds and records subscribe responses.
+        /// Otherwise the prefix is already included in the upcoming subscribe response.
+        notify_miner: bool,
+    },
+    SetupComplete,
+}
```
bit-aloo, line 47: I would expect something like this to be part of sv1 crate in stratum core, having typed messages instead of us creating a wrapper for it.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
```diff
@@ -87,9 +113,17 @@ const MAX_FUTURE_BLOCK_TIME: u32 = 2 * 60 * 60;
...
 const MAX_QUEUED_SV1_HANDSHAKE_MESSAGES: usize = 8;
 
+fn sv1_setup_request(request: &Client2Server) -> Option<Sv1SetupRequest> {
+    match request {
+        Client2Server::Subscribe(_) => Some(Sv1SetupRequest::Subscribe),
+        Client2Server::Authorize(_) => Some(Sv1SetupRequest::Authorize),
+        _ => None,
+    }
+}
+
```
bit-aloo: I don't think we need this helper?

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: We need to keep this separately. `notify.time` is advanced for every generated keepalive, while this value remains the original upstream job time. 

The future-time limit must stay relative to that original value, otherwise the limit would move forward at every keepalive. 

I’ll clarify the comment.

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: `sv1_api` already has the protocol-level `Server2Client` enum, but this one has a different role. It's the tProxy queue for one downstream and also carries local ordering state such as `SetupComplete` and `notify_miner`. 

`Arc<Notify>` also lets us share one job across aggregated miners.

I don’t think those application-level details belong in the Sv1 crate, but I’ll clarify the comment so the distinction is explicit.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Agree, I'll remove it soon

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Done in https://github.com/stratum-mining/sv2-apps/commit/ab69227930a830ce40d5dd15356a1c18d0cd2a42.

The comment now makes explicit that `notify.time` advances while `keepalive_base_time` remains fixed.

miner-apps/translator/src/lib/sv1/downstream.rs
GitGab19: Clarified in https://github.com/stratum-mining/sv2-apps/commit/7f559f5ff5f0537fec7716dfc4cdd089c075e73e.

The documentation now distinguishes this tProxy-local delivery queue from the protocol-level types in `sv1_api`.

miner-apps/translator/src/lib/sv1/sv1_server/mod.rs
GitGab19: Done in https://github.com/stratum-mining/sv2-apps/commit/21fc789b5b01149c52c833da03e73623924e28aa.

I removed the helper and inlined the typed match at its only call site.

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -745,14 +897,10 @@ impl HandleMiningMessagesFromServerOwnedAsync for ChannelManager {
...
-                .await
-                .map_err(|e| {
-                    error!("Failed to send immediate NewExtendedMiningJob: {:?}", e);
-                    TproxyError::shutdown(TproxyErrorKind::ChannelErrorSender)
-                })?;
+            self.forward_job_to_sv1_server(message).await?;
+        }
+        if self.mode.is_aggregated() {
+            self.open_pending_aggregated_downstream_channels().await?;
         }
```
bit-aloo: This can be removed now?, Requests are only buffered in the NoChannel and Pending states, so those calls could never find anything

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
```diff
@@ -957,14 +1105,11 @@ impl HandleMiningMessagesFromServerOwnedAsync for ChannelManager {
...
-                .map_err(|e| {
-                    error!("Failed to send NewExtendedMiningJob: {:?}", e);
-                    TproxyError::shutdown(TproxyErrorKind::ChannelErrorSender)
-                })?;
+            self.forward_job_to_sv1_server(message).await?;
+        }
+
+        if self.mode.is_aggregated() {
+            self.open_pending_aggregated_downstream_channels().await?;
         }
```
bit-aloo: Same for this callsite

bit-aloo (commented):
Some, removals and then will approve

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: Done in https://github.com/stratum-mining/sv2-apps/commit/a826d7f77a3055b848f98604ce87666934178d94

miner-apps/translator/src/lib/sv2/channel_manager/mining_message_handler.rs
GitGab19: Done in https://github.com/stratum-mining/sv2-apps/commit/a826d7f77a3055b848f98604ce87666934178d94

bit-aloo (approved):
Thanks Gitgab. ACK

PR #827, opened 2026-09-02
JDC: document `user_identity` in `README.md`

### LINKED ISSUE #417: JDC needs to authenticate itself via `AllocateMiningJobToken.user_identifier`
it should simply take `user_identity` that already exists on its `.toml` configuration

plebhash:
follow-up to #501 

close #417


bit-aloo (approved):
ACK

Issue #779, opened 2026-08-20
propagate and authenticate JDP client identity


GitGab19:
## Affected applications

- JDC
- JDS
- Pool

## Problem

JDC already sends its configured identity through `AllocateMiningJobToken.user_identifier` (#501), but JDS does not use that field.

As a result, JDS cannot associate token allocation and job declaration activity with a client identity, and Pool cannot verify that a `SetCustomMiningJob` arrives on a channel whose `user_identity` matches the identity under which the token was allocated.

## Expected behavior

JDS should bind the identity to each allocated token and carry it over to the active token. When Pool asks JDS to validate a `SetCustomMiningJob`, it should pass the channel's `user_identity`, and JDS should reject the job if it does not match the identity bound to the token.

The binding is per token, not per connection: a JDC may legitimately allocate tokens under different identities over the same JDP connection (e.g. one per Mining channel it opens with Pool).

The Noise handshake (NX) authenticates only the server, so `user_identifier` is an unauthenticated claim. JDS cannot authenticate it; it can only enforce its consistency with the channel identity that Pool accepted.

The check belongs in `JobDeclarator::handle_set_custom_mining_job`, where the active token is looked up. `JobValidationEngine` does not need to change.

## Acceptance criteria

- Add explicit JDC configuration for the JDP user identifier. (done in #501)
- Propagate it unchanged in token-allocation requests. (done in #501)
- Bind it to each allocated token in JDS (`AllocatedTokenData` and `ActiveTokenData`), not to the downstream connection.
- Pool passes the channel's `user_identity` to `JobDeclarator::handle_set_custom_mining_job`; JDS answers `SetCustomMiningJob.Error` with `invalid-mining-job-token` when it does not match the identity bound to the token.
- Reject an empty `user_identifier` before allocating a token, and reject an empty `user_identity` in JDC configuration.
- Avoid logging raw identity values where an opaque identifier is sufficient.
- Add tests: matching identity accepted; token allocated under one identity used on a channel with a different identity rejected; two identities on the same JDP connection each accepted on their matching channel; empty identity rejected.

## Related sv2-apps issues

- #416
- #417
- #610

nulllpc:
Can I take this issue? @GitGab19

GitGab19:
Sure @nulllpc!

plebhash:
refined acceptance criteria on the issue description above

PR #839, opened 2026-09-07
feat(jds): validate and bind user_identifier to downstream connections

nulllpc:
This implements the identity tracking requirements for JDS to ensure reliable token allocation and job declaration activity.

- Bind user_identifier to the Downstream state during AllocateMiningJobToken
- Reject empty strings or mid-session mismatched identities with a new InvalidUserIdentifier disconnect error to enforce connection consistency
- Remove raw user_identifier logging in the declaration handler to prevent unnecessary identity leakage
- Pass the bound user_identity to JobValidationEngine::handle_set_custom_mining_job so upstream pools can authenticate the custom job
- Add integration tests verifying JDS correctly disconnects downstreams that send empty or mismatched identities, while allowing replayed identities

Solves #779


integration-tests/tests/bitcoin_core_ipc_jdp_io.rs
```diff
@@ -6,8 +6,8 @@
 //! - `DeclareMiningJob` returns `Error(stale-chain-tip)` when the declared BIP34 height is
 //!   intentionally mismatched.
 //! - `DeclareMiningJob` does not retain client-supplied transactions when validation fails.
-//! - `DeclareMiningJob` rejects a coinbase that does not carry exactly one input, without
-//!   tearing down the IPC connection.
+//! - `DeclareMiningJob` rejects a coinbase that does not carry exactly one input, without tearing
+//!   down the IPC connection.
```
plebhash, line 10: was this intentional?

integration-tests/tests/jd_integration.rs
```diff
@@ -1562,3 +1562,170 @@ async fn jdc_per_upstream_identity_switches_on_fallback() {
...
+    assert!(
+        res.is_err(),
+        "JDS should have disconnected the client for empty identity"
+    );
+
+    shutdown_all!(pool);
+}
+
+// This test verifies that JDS requires all subsequent token allocations on a single
+// connection to use the exact same identity, preventing mid-session identity swapping.
```
plebhash, line 1620: hmm why isn't a JDC allowed to allocate tokens under different indentities?

maybe JDC opened multiple channels with Pool, each with a different `user_identity`

integration-tests/tests/jd_integration.rs
```diff
@@ -1562,3 +1562,170 @@ async fn jdc_per_upstream_identity_switches_on_fallback() {
...
+    assert!(
+        res.is_err(),
+        "JDS should have disconnected the client for inconsistent identity"
+    );
+
+    shutdown_all!(pool);
+}
+
+// This test verifies that JDS accepts subsequent token allocations on a single
+// connection as long as they use the exact same (replayed) identity.
```
plebhash, line 1682: hmm why isn't a JDC allowed to allocate tokens under different indentities?

maybe JDC opened multiple channels with Pool, each with a different `user_identity`

pool-apps/jd-server/src/lib/job_declarator/job_validation/mod.rs
```diff
@@ -49,6 +49,7 @@ pub trait JobValidationEngine: Send + Sync {
     async fn handle_set_custom_mining_job(
         &self,
         downstream_id: DownstreamId,
+        user_identity: Option<String>,
```
plebhash, line 52: why is this `Option`?

on Pool, every channel always has a `user_identity`

pool-apps/jd-server/src/lib/job_declarator/job_validation/bitcoin_core_ipc.rs
```diff
@@ -756,9 +756,17 @@ impl JobValidationEngine for BitcoinCoreIPCEngine {
     async fn handle_set_custom_mining_job(
         &self,
         downstream_id: DownstreamId,
+        user_identity: Option<String>,
```
plebhash, line 759: why is this `Option`?

on Pool, every channel always has a `user_identity`

pool-apps/jd-server/src/lib/job_declarator/downstream/mod.rs
```diff
@@ -68,6 +68,8 @@ pub struct Downstream {
     /// Extensions that JDS requires
     #[allow(unused)]
     pub required_extensions: Vec<u16>,
+    /// The authenticated user identity bound to this connection via AllocateMiningJobToken.
+    pub user_identity: SharedLock<Option<String>>,
```
plebhash, line 72: I don't think adding this parameter to `Downstream` is the right choice

this design seems heavily based around the assumption that every JDC will always have one single Mining Channel with Pool

while that's currently true with SRI JDC, that's not a restriction imposed by spec, and a JDC that opens multiple Channels is theoretically possible

we want to enforce cohesion between token allocation and `user_identity`... but enforcing 1 single `user_identity` per Connection seems like a misguided assumption

plebhash (changes requested):
we're not yet in the right direction here

we should be touching JDS `token_management/mod.rs` and adding a new `user_identity: String` field to `AllocatedTokenData` and `ActiveTokenData`

(tbh I think those types deserve a dedicated commit turning them from tuples into structs)

then during the execution of  `JobDeclarator::handle_set_custom_mining_job` we should make sure the channel's `user_identity` (passed from Pool) matches the one associated with the active token

---

https://github.com/stratum-mining/sv2-apps/issues/779#issuecomment-5590518378

nulllpc:
Thanks @plebhash for the review. I'll refine the PR. I'll turn it into draft for now

nulllpc:
Hi @plebhash sorry it took me a bit of time to really wrap my head around this, but I think I got it this time. Please review when you have the time, thanks!

pool-apps/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -488,3 +504,160 @@ impl JobDeclarator {
...
+            )
+            .unwrap(),
+            Arc::new(TaskManager::new()),
+        )
+        .await
+        .unwrap()
+    }
+
+    #[tokio::test]
+    async fn set_custom_mining_job_rejects_mismatched_identity() {
```
nulllpc, line 595: Note for the mismatched-identity test: I make it a unit test in jd-server rather than an integration test. Getting an active token over the wire needs either a real JDC (which can't produce a mismatch — one user_identifier per upstream connection) or hand-rolled job-declaration data against live bitcoin-core state. It's too much work just to test an internal comparison inside JDS. Let me know if you think differently

PR #847, opened 2026-09-10
tProxy+JDC: try every DNS address of an upstream, not just the first

### LINKED ISSUE #846: parsing `upstream` connections from config files only tries the first address DNS returns
## Affected applications

- tProxy
- JDC

## Problem

`stratum_apps::network_helpers::resolve_host` performs a DNS lookup and then keeps only the first address `tokio::net::lookup_host` yields, discarding the rest. Every caller therefore dials exactly one candidate, chosen by whatever order the resolver happened to return.

A host that publishes both an AAAA and an A record, where only one of the two accepts connections on the Sv2 port, is reported as unreachable. `stratum.ckpool.org` is a live example:

```
$ getent ahosts stratum.ckpool.org
2604:2dc0:100:240f::1  STREAM stratum.ckpool.org
51.81.56.15            STREAM

2604:2dc0:100:240f::1:3336  connection refused
51.81.56.15:3336            accepts
```

tProxy configured against it never reaches the working address:

```
INFO  translator_sv2::translator_runtime: Trying upstream 1 of 3: stratum.ckpool.org:3336
INFO  stratum_apps::network_helpers::resolve_hostname: Resolving hostname 'stratum.ckpool.org' via DNS...
ERROR translator_sv2::sv2::upstream: Failed to connect to [2604:2dc0:100:240f::1]:3336: Connection refused (os error 111).
...
WARN  translator_sv2::translator_runtime: Max retries reached for stratum.ckpool.org:3336, moving to next upstream
INFO  translator_sv2::translator_runtime: Trying upstream 2 of 3: <backup pool>
```

Two consequences beyond the failed connection:

- The retry budget is spent for nothing. Resolution is deterministic within a run, so attempts 2 and 3 re-resolve the same name and dial the same dead address.
- The failure escalates into upstream fallback. An operator whose primary pool is reachable silently ends up mining on a backup entry, which is a payout-relevant misroute rather than a connectivity annoyance. With a single-entry upstream list the application simply refuses to start.

Three connection paths are affected, all of them via `resolve_host`:

- tProxy → pool (`translator/src/lib/sv2/upstream/mod.rs`)
- JDC → pool (`jd-client/src/lib/upstream/mod.rs`)
- JDC → JDS (`jd-client/src/lib/job_declarator/mod.rs`)

The Pool and JDC template receivers are not affected: they hand the `host:port` string straight to `TcpStream::connect`, which already walks every resolved address. Their use of `resolve_host_port` only derives a `SocketAddr` to fill in the `SetupConnection` endpoint fields and never dials it.

Hostname support was introduced in #286, closing #89; the single-address truncation has been present since.

## Expected behavior

No upstream should be declared unreachable until every address its hostname resolves to has been attempted. `tokio::net::TcpStream::connect` already implements exactly this when given more than one address, so `resolve_host` should return the full candidate list and let the connect call walk it.

Once several candidates exist, the address recorded up front no longer identifies the connection. The peer address should be read back from the connected socket so that connection logs, and the `endpoint_host` / `endpoint_port` fields tProxy puts in `SetupConnection`, describe the address actually in use rather than the first candidate.

Fallback to the next upstream should remain reserved for a host where no address accepts a connection.

## Acceptance criteria

- `resolve_host` returns every address the name resolves to, in resolver order, and still errors when the lookup yields none.
- tProxy, JDC → pool, and JDC → JDS pass the full candidate list to `TcpStream::connect`.
- The connected peer address is read from the socket and used for connection logging and for tProxy's `SetupConnection` endpoint fields.
- A connection failure logs the full candidate list, so an operator can tell "no address accepted" apart from "one address refused".
- `resolve_host_port` keeps its single-address contract, with docs directing callers that mean to dial at the address-iterating path.
- Add a test asserting `resolve_host` preserves everything `lookup_host` returned, so a future `.next()` or `.take(1)` fails rather than silently stranding a dual-stack host.

## Related sv2-apps issues

- #89

plebhash:
close #846

### DIFF
--- a/miner-apps/jd-client/src/lib/upstream/mod.rs
+++ b/miner-apps/jd-client/src/lib/upstream/mod.rs
@@ -152,7 +152,7 @@ impl Upstream {
         task_manager: Arc<TaskManager>,
         required_extensions: Vec<u16>,
     ) -> JDCResult<Self, error::Upstream> {
-        let addr = resolve_host(&upstream_entry.pool_host, upstream_entry.pool_port)
+        let candidate_addrs = resolve_host(&upstream_entry.pool_host, upstream_entry.pool_port)
             .await
             .map_err(|e| {
                 error!(
@@ -162,10 +162,16 @@ impl Upstream {
                 JDCError::fallback(JDCErrorKind::NetworkHelpersError(e.into()))
             })?;
 
-        let stream = tokio::time::timeout(TCP_CONNECT_TIMEOUT, TcpStream::connect(addr))
-            .await
-            .map_err(JDCError::fallback)?
-            .map_err(JDCError::fallback)?;
+        // Each resolved address gets an attempt, so a pool that publishes both an IPv6 and an
+        // IPv4 address is still reachable when only one of the two accepts connections.
+        let stream = tokio::time::timeout(
+            TCP_CONNECT_TIMEOUT,
+            TcpStream::connect(&candidate_addrs[..]),
+        )
+        .await
+        .map_err(JDCError::fallback)?
+        .map_err(JDCError::fallback)?;
+        let addr = stream.peer_addr().map_err(JDCError::fallback)?;
         info!("Connected to upstream at {}", addr);
         debug!("Begin with noise setup in upstream connection");
 
--- a/miner-apps/jd-client/src/lib/job_declarator/mod.rs
+++ b/miner-apps/jd-client/src/lib/job_declarator/mod.rs
@@ -137,7 +137,7 @@ impl JobDeclarator {
         mode: JDMode,
         task_manager: Arc<TaskManager>,
     ) -> JDCResult<Self, error::JobDeclarator> {
-        let addr = resolve_host(&upstream_entry.jds_host, upstream_entry.jds_port)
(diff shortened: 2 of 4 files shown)


miner-apps/jd-client/src/lib/upstream/mod.rs
```diff
@@ -162,10 +162,16 @@ impl Upstream {
...
 
-        let stream = tokio::time::timeout(TCP_CONNECT_TIMEOUT, TcpStream::connect(addr))
-            .await
-            .map_err(JDCError::fallback)?
-            .map_err(JDCError::fallback)?;
+        // Each resolved address gets an attempt, so a pool that publishes both an IPv6 and an
+        // IPv4 address is still reachable when only one of the two accepts connections.
+        let stream = tokio::time::timeout(
+            TCP_CONNECT_TIMEOUT,
+            TcpStream::connect(&candidate_addrs[..]),
```
bit-aloo, line 169: Oh, I did knew connect can take a slice of address. Thats pretty cool.

bit-aloo (approved):
ACK

PR #857, opened 2026-09-13
pool payout policy isolation for solo mining

### LINKED ISSUE #623: pool: `payout_mode` in `Downstream` should be handled per-channel and not as a single value
## Problem

Pool derives a payout policy from the `user_identity` in each `OpenStandardMiningChannel` or `OpenExtendedMiningChannel`, but stores the result in one connection-wide `DownstreamData::payout_mode`.

Opening a later channel overwrites that value. On the next template, Pool uses the last value to construct jobs for all channels on the connection. On a multiplexed connection, one channel can therefore redirect another channel’s coinbase payout.

This state exists only in the Pool application; JDC and tProxy do not store a connection-wide `payout_mode` for multiple independently identified mining channels.

## Expected behavior

Payout authorization must be associated with the mining channel that supplied the relevant identity. A channel must never inherit a payout policy from a different channel opened on the same connection.

Group-channel jobs also need an explicit policy: channels with incompatible payout modes must either be separated into compatible groups or rejected from sharing a group.

## Acceptance criteria

- Store the resolved payout policy per standard or extended channel.
- Use that channel’s policy whenever regenerating its jobs.
- Define and enforce how group channels handle incompatible payout policies.
- Ensure opening or closing one channel cannot change another channel’s coinbase outputs.
- Add tests with two channels using different payout identities across a template refresh.

## References

- Loupe findings:
  - project-loupe/audit-sv2-apps#31
  - project-loupe/audit-sv2-apps#138

lucasbalieiro:
closes #623
closes https://github.com/project-loupe/audit-sv2-apps/issues/31
closes https://github.com/project-loupe/audit-sv2-apps/issues/138

now, scoping the payout based on a channel basis, not downstream;

the issue states:
> Group-channel jobs also need an explicit policy: channels with incompatible payout modes must either be separated into compatible groups or rejected from sharing a group.

I've chosen the rejection path;


plebhash:
we're trying to enforce the following:

https://github.com/stratum-mining/sv2-apps/blob/6d62f53acfd389287770b6943137a294859fe460/CONTRIBUTING.md?plain=1#L105

this helps the review process, both by humans and agents

plebhash:
> the issue states:
> 
> > Group-channel jobs also need an explicit policy: channels with incompatible payout modes must either be separated into compatible groups or rejected from sharing a group.
> 
> I've chosen the rejection path;

I think this is definitely a viable direction, but it feels more like a hotpatch than a careful decision on how we want to design this functionality, so I think  we should consider the UX implications:

on the current direction of this PR, we are simply prohibiting the functionality gaps that the Loupe issues mislabeled as "vulnerabilities"

more specifically, a connection is never allowed to have two different channels with conflicting payout modes (because we still only allow one group per channel)

this prohibition is enacted by denying the ability for any connection to have incompatible payout modes

the only exception is a `REQUIRES_STANDARD_JOBS` connection, which is usually opened directly by Mining Devices and therefore it's extremely unlikely to hold multiple channels anyways

---

theoretically, the "perfect" solution would be to implement a new grouping policy based on different payout modes

if a connection ever gets multiple channels, with clusters of channels following different payout modes, then different groups are established to each of those clusters

tbh that would actually be pretty interesting, since that's the first real use-case I could ever envision for different groups under the same connection

unfortunately though, that would mean a substantial refactor... and I'm not sure there's real demand for such elaborate functionality to really justify that kind of engineering effot today

so a KISS hotpatch does feels justified here, and I think @lucasbalieiro's instincts went on the right direction

---

with that said, I feel this PR is sitting in a weird limbo between these two opposite directions

we replaced `Downstream.payout_mode` with `Downstream.{channel_payout_modes,group_payout_mode}` (which brings on a whole bunch of extra complexity) but kept pretty much the exact same functionality while forbidding the undesired edge cases

so I'm wondering if we shouldn't drastically simplify things here, and simply forbid channels from using incompatible payout modes under the same connection?

that would make the scope of this PR much smaller, and we'd still be addressing the core problem at hand

p.s.: if we go in that direction, this should be very clearly documented as a deliberate trade-off, because in a way it kills elaborate use-cases that ideally should be permitted under a "perfect" implementation... this documentation will remain as a reminder in case we ever get signals of real demand for this kind of functionality

lucasbalieiro:
@plebhash, I've simplified things here;

plebhash:
@lucasbalieiro we're missing version bumping on the PR

see `CONTRIBUTING.md` for versioning rules

pool-apps/pool/README.md
plebhash, line 147: we're forgetting to update this

integration-tests/tests/pool_solo_mining.rs
```diff
@@ -207,6 +208,315 @@ async fn pool_solo_mining_invalid_payout_address() {
...
+            MiningOwned::OpenExtendedMiningChannel(OpenExtendedMiningChannelOwned {
+                request_id: 1,
+                user_identity: "sri/donate/worker.1".try_into().unwrap(),
+                nominal_hash_rate: -1.0,
+                max_target: [0xff; 32].into(),
+                min_extranonce_size: 8,
+            }),
+        ))
+        .await
+        .unwrap();
```
plebhash, line 345: clanker finding: 

> The second open is refused by the payout check, not the `-1.0` hashrate, so this duplicates
> `pool_rejects_incompatible_group_payout_policy` and never tests "set the policy only after a
> successful open". Invert it: first open `sri/donate/worker.1` with the bad hashrate and fail,
> then `sri/solo/<addr>` must succeed. 
>
> Also both test names still say "group".

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -188,16 +215,26 @@ impl HandleMiningMessagesFromClientOwnedAsync for ChannelManager {
...
 
+                if !payout_is_compatible(downstream, &payout_mode, &self.coinbase_reward_script)? {
+                    let open_standard_mining_channel_error = OpenMiningChannelErrorOwned {
+                        request_id,
+                        error_code: ERROR_CODE_OPEN_MINING_CHANNEL_INCOMPATIBLE_PAYOUT_MODE
+                            .to_string()
+                            .try_into()
+                            .expect("error code must be valid string"),
+                    };
+                    return Ok(vec![(
```
plebhash, line 232: add a log here, otherwise this rejection goes silent to the operator

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -407,6 +448,48 @@ impl HandleMiningMessagesFromClientOwnedAsync for ChannelManager {
...
+                    &self.coinbase_reward_script,
+                )? {
+                    let open_extended_mining_channel_error = OpenMiningChannelErrorOwned {
+                        request_id,
+                        error_code: ERROR_CODE_OPEN_MINING_CHANNEL_INCOMPATIBLE_PAYOUT_MODE
+                            .to_string()
+                            .try_into()
+                            .expect("error code must be valid string"),
+                    };
+                    return Ok(vec![(
```
plebhash, line 498: add a log here, otherwise the error goes silent to the operator

pool-apps/pool/README.md
lucasbalieiro: updated, tks

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
lucasbalieiro: added a `warn!` saying: `Rejecting standard channel with an incompatible payout policy`

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
lucasbalieiro: added a `warn!` saying: `Rejecting extended channel with an incompatible payout policy`

integration-tests/tests/pool_solo_mining.rs
lucasbalieiro: updated

lucasbalieiro (commented):
about:

> @lucasbalieiro we're missing version bumping on the PR
> see CONTRIBUTING.md for versioning rules


unsure if I got the logic there, in this new commit i've done a minor bump:
678097c33cadf120e7a095f32269b052e1c63da5

plebhash:
clanker finding:

> b1551260 has no behaviour change (two `let`s moved, three calls reflowed), and its message describes what main already does. Please drop it, and add the #623 URL to d109b1ff.

pool-apps/pool/src/lib/utils.rs
```diff
@@ -15,6 +15,19 @@ pub use stratum_apps::payout::{PayoutMode, PayoutModeError};
 
 pub(crate) type DownstreamMessage = (MiningOwned, Option<Vec<Tlv>>);
 
+pub(crate) const ERROR_CODE_OPEN_MINING_CHANNEL_INCOMPATIBLE_PAYOUT_MODE: &str =
+    "incompatible-payout-mode";
+
+/// Returns whether two payout modes produce the same group-channel coinbase policy.
```
plebhash: doc still says "group-channel coinbase policy"

since the last reshape this is the connection policy

lucasbalieiro: thanks

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -37,10 +38,30 @@ use jd_server_sv2::job_declarator::SetCustomMiningJobResponse;
...
+    downstream
+        .payout_mode
+        .with(|current| {
+            current.as_ref().is_none_or(|current| {
+                payout_modes_are_compatible(current, payout_mode, pool_reward_script)
+            })
+        })
+        .map_err(PoolError::shutdown)
+}
+
```
bit-aloo: Lets have this as a method in downstream

pool-apps/pool/src/lib/downstream/mod.rs
```diff
@@ -70,7 +70,7 @@ pub struct Downstream {
     pub channel_id_factory: Arc<AtomicU32>,
     /// Extensions that have been successfully negotiated with this client
     pub negotiated_extensions: SharedLock<Vec<u16>>,
-    /// Payout mode derived from user_identity (None until channel is opened)
+    /// Payout policy for this connection, derived from the first channel's user_identity.
     pub payout_mode: SharedLock<Option<PayoutMode>>,
```
bit-aloo: Can this be OnceLock?

bit-aloo (commented):
Few suggestions, rest looks ok

pool-apps/pool/src/lib/utils.rs
```diff
@@ -15,6 +15,19 @@ pub use stratum_apps::payout::{PayoutMode, PayoutModeError};
...
+/// Returns whether two payout modes produce the same connection payout policy.
+pub(crate) fn payout_modes_are_compatible(
+    first: &PayoutMode,
+    second: &PayoutMode,
+    pool_reward_script: &stratum_apps::config_helpers::CoinbaseRewardScript,
+) -> bool {
+    first.coinbase_outputs(100, pool_reward_script)
+        == second.coinbase_outputs(100, pool_reward_script)
+}
+
```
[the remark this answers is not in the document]
lucasbalieiro: unsure, because this feature is for OUR pool specifically;

I've addressed your point because in the first implementation of this feature I've made stratum-apps/src/payout.rs owns PayoutMode and coinbase_outputs; so this will help keep consistency;

* moved compatibility logic to PayoutMode::has_same_coinbase_outputs_as.
* bumped stratum-apps to 0.8.1 and updated lockfiles.

pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs
lucasbalieiro: sure, moved to  `Downstream::payout_is_compatible`;

pool-apps/pool/src/lib/downstream/mod.rs
lucasbalieiro: nice, I didn't know that this thing existed; TIL;

but upon reading the documentation, I do not think that this will be a good move given the current state;

my reason is that the payout policy is cleared when the downstream closes its last channel so it can establish a new policy. 

IIUC, `OnceLock` would make the first policy permanent.

see: `fn handle_close_channel` in `pool-apps/pool/src/lib/channel_manager/mining_message_handler.rs`

bit-aloo: Ah! makes sense

bit-aloo (approved):
ACK. Thanks Lucas.

plebhash (approved):

PR #616, opened 2026-07-15
binary inner split sv2-app fallout

bit-aloo:
companion to https://github.com/stratum-mining/stratum/pull/2225

PR #865, opened 2026-09-16
chore: bump `bitcoin_core_sv2` to `0.6.0`

plebhash:
missed on #616 and #677


GitGab19 (approved):

GitGab19:
Does it mean the `semver-check` check didn't spot this breaking change in the public APIs?

plebhash:
> Does it mean the `semver-check` check didn't spot this breaking change in the public APIs?

no public APIs were changed, so this is the blindspot that we always had on CI

the original PR description I posted wrongly stated that public APIs were broken, which is incorrect (my bad for blindly copypasting clanker slop)

`bitcoin_core_sv2` MAJOR is still at `0.x.y`, so we always bump MINOR for breaking API changes, which was the original direction of the PR

nevertheless, a bump is still necessary, because otherwise we cannot publish to `crates.io`

and a PATCH bump would not be appropriate, therefore the PR contents remain the same, and I'm only adapting the PR/commit description

Issue #873, opened 2026-09-17
replace `mining_device` with `cpu_miner_sv2`


plebhash:
in 2025 I started [`sv2-cpu-miner`](https://github.com/plebhash/sv2-cpu-miner) based on experimental (and discontinued) `tower-stratum`/`sv2-services` crate as a dependency

recently I started refactoring it to use `sv2-apps` and `stratum` crates as dependencies exclusively

this implementation has a few advantages over `mining_device` (which we've already considered deprecating, and has been temporarily been living as a module in `integration_tests_sv2` until we find a better alternative)

more specifically, it's architecture was designed from scratch with the following test-oriented UX in mind:
- multi-channel support: enables emulating proxies against mining servers
- nominal hashrate multiplier: enables forcing vardiff edge cases
- cpu throttling: enables fine-tuning cpu usage
- single submit mode: avoids undesired share spamming
- graceful shutdown: no hanging zombie processes

obviously I'm biased here, but there's many things I don't like about `mining_device`:
- poor modularization: the code evolved as a spaghetti inside a single module
- no integration with `stratum-core` crates like `handlers_sv2`: invites pattern fragmentation
- performance overengineering: why add complexity to maximize hashrate? does it really matter whether we're getting 10kH/s or 20kH/s ? and why would I want to fry my CPU for some tests?

maybe one could argue that now with AI we could simply refactor it to fulfil our needs, but since I have already spent time on `sv2-cpu-miner` and have a good understanding of the architecture and design principles, I prefer to start there

---

so I propose we add it as a new crate called `cpu_miner_sv2` under the `miner-apps` workspace... this crate name follows the `*_sv2` crate naming pattern on `sv2-apps`

with `cpu_miner_sv2`, we will be able to perform much more versatile tests, either by manually deploying it as a binary executable, or on integration tests

with the addition of this crate, we deprecate `mining_device` by purging it from the codebase, and replacing it inside integration tests

plebhash:
note: there's already some integration tests on the repository

they should be migrated to `integration-tests/tests`

plebhash:
note: this is currently blocked by https://github.com/stratum-mining/stratum/issues/2393

PR #874, opened 2026-09-18
add `Clanker-driven smoke tests` section to `AGENTS.md`

plebhash:
⚠️ WARNING ⚠️ this is an experiment ⚠️

It should not be taken as an encouragement to replace or abandon the multitude of testing methodologies that are already established as best-practices in software engineering.

In fact, we should always be allocating (human) engineering efforts into classic methodologies, instead of relying purely on clanker-driven test coverage.

Examples where we should be doing this kind of coverage on the long run:
- `integration_tests_sv2`
- `interoperability_tests_sv2`
- `stratamoto` (?)

With that said, I think some interesting insights might come from this exercise, and since we're experimenting with `AGENTS.md` I felt it could be interesting to see what kind of hybrid human/AI workflows we might refine out of this.

---

This PR adds a `## Clanker-driven smoke tests` section to `AGENTS.md`, documenting the pre-release smoke-test workflow against the hosted `pool_sv2` deployment so agents can drive it under explicit human instruction. A testing session is confined to one throwaway workspace directory plus user-space processes, with an explicit teardown step, so it leaves nothing behind on the host.

## Topologies

```mermaid
flowchart LR
  sv1["10x minerd (Sv1)"]
  sv2["10x sv2-cpu-miner (Sv2)"]
  tproxy[translator_sv2]
  jdc[jd_client_sv2]
  pool["hosted pool_sv2 :3333"]
  jds["hosted JDS :3334"]
  core["local Bitcoin Core (IPC)"]

  sv1 -- "T1 / T2 / J2" --> tproxy
  sv2 -- "J1" --> jdc
  tproxy -- "T1 / T2" --> pool
  tproxy -- "J2" --> jdc
  jdc --> pool
  jdc --> jds
  core --> jdc
```

## Core scenarios

| # | Fleet | Path | Key config | Expected |
| --- | --- | --- | --- | --- |
| T1 | 10x `minerd` | tproxy -> pool | `aggregate_channels = true` | 1 upstream channel |
| T2 | 10x `minerd` | tproxy -> pool | `aggregate_channels = false` | 10 upstream channels |
| J1 | 10x `sv2-cpu-miner` | JDC -> pool/JDS | defaults | 10 downstream channels, locally declared jobs |
| J2 | 10x `minerd` | tproxy -> JDC -> pool/JDS | tproxy `enable_vardiff = false` | JDC owns difficulty end to end |

## `sv2-cpu-miner` variations

Run against both JDC and the hosted pool. These reach server-side paths a plain mining device never touches:

| Knob | Server path exercised | Watch for |
| --- | --- | --- |
| `n_standard_channels`, `n_extended_channels` | per-channel bookkeeping and group-channel job routing on one connection (proxy emulation) | all channels under one client in `/api/v1/clients/{id}/channels` |
| `requires_standard_jobs = true` | per-channel `NewMiningJob` instead of group `NewExtendedMiningJob`; nothing else in the toolkit reaches this branch | jobs and shares still flow |
| standard channel against JDC | JDC bridges a standard downstream to an extended upstream channel | pool reports extended, JDC reports standard |
| `nominal_hashrate_multiplier` | initial target derivation and vardiff convergence from a skewed seed | `SetTarget` converges to `shares_per_minute` |
| `single_submit = true` | fast share-acceptance check | one accepted share per channel, then idle |
| `cpu_usage_percent` | throttling; advertised hashrate scales with it | ten processes without pegging the host |

## Edge cases

| Case | Action | Pass |
| --- | --- | --- |
| Upstream failover | sever primary via local `socat` relay | tproxy falls back to next `[[upstreams]]`; JDC walks its list, then solo-mines |
| Proxy restart under load | Ctrl+C with 10 miners attached | graceful shutdown; `minerd` re-attaches on its own |
| Miner churn | kill some, add late joiners | channels drop from monitoring; newcomers submit within seconds |
| Bad config | wrong port, wrong `authority_pubkey`, malformed TOML | fast failure naming the actual problem |
| Bitcoin Core outage (JDC) | node down at startup and mid-run | clear error, sane recovery |
| New-block boundary | run across a mainnet block | at most a few stale rejects at the prev-hash flip |
| Soak | leave running for hours | flat memory, responsive monitoring, zero rejects |

Verification goes through the monitoring HTTP endpoints (tproxy `:9092`, JDC `:9091`); healthy run = channels open, shares accepted, zero rejected. Also documents the `min_individual_miner_hashrate` gotcha (ASIC-seeded difficulty stalls CPU miners for ~10 min).

---

Generated with Fable 5


AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
 ## Ponytail
 
 If the plugin is not already installed into the coding agent harness, make sure to follow [Ponytail](https://ponytail.dev/) rules. But avoid installing it as a plugin, unless explicitly instructed to do so. This is only a repository-wide convention. Also avoid writing comments that reference "ponytail" in a compressed and implicit way, prefer explaining the actual rationale instead.
+
+## Clanker-driven smoke tests
```
GitGab19: Would you consider moving the runbook itself to something like `docs/smoke-tests.md` and keeping a short pointer here with the "only when explicitly instructed" guard?

This file is loaded into every agent session (it is symlinked as `CLAUDE.md`), and this section triples its size with content that only matters when a human asks for a smoke test. It also restates values that live elsewhere (hosted IP, ports, `monitoring_cache_refresh_secs`, route names), so it will drift from the config examples and the monitoring crate over time.

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+
+Before a release we smoke-test the miner apps against the hosted `pool_sv2` deployment (`75.119.150.111`, pool `:3333`, JDS `:3334`). Sv1 miners use [`minerd`](https://github.com/stratum-mining/cpuminer) (`minerd -a sha256d -t 1`, one process per miner); Sv2 miners use [`sv2-cpu-miner`](https://github.com/plebhash/sv2-cpu-miner). Build `translator_sv2` and `jd_client_sv2` from the release commit; JDC scenarios need a synced local Bitcoin Core node with IPC enabled.
```
GitGab19: Two pointers would keep this from going stale:

- `RELEASE.md`'s release process has no smoke-test step yet, so either add one linking here or soften "Before a release we smoke-test".
- Once #873 lands, the build instructions for `sv2-cpu-miner` change (it becomes `cpu_miner_sv2` under `miner-apps/`). Worth a note.

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+
+Only do this when explicitly instructed to — it is not a regular development workflow. A session touches the system in exactly two ways: one throwaway workspace directory holding every config, log, and helper file, and the user-space test processes themselves. Both are gone after Teardown.
+
+### Core scenarios
+
+10 miners each, all using `miner-apps/*/config-examples/mainnet/*-hosted-*` configs with a real `user_identity`:
```
GitGab19: `*-hosted-*` matches two JDC files, and `jdc-config-hosted-infra-example.toml` uses the hosted Sv2 TP with no local Bitcoin Core at all.

Suggest naming `jdc-config-bitcoin-core-ipc-hosted-infra-example.toml` explicitly, and noting that the hosted-TP variant lets the JDC scenarios run without a node (only the Bitcoin Core outage case needs one).

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+
+- Translator + 10 Sv1 miners, run once with `aggregate_channels = true` (expect one upstream channel) and once `false` (expect one upstream channel per miner).
+- JDC + 10 Sv2 miners pointed at JDC's `listening_address`.
+- JDC + a translator whose single `[[upstreams]]` is that JDC (`address`/`port` = JDC `listening_address`, `authority_pubkey` = JDC `authority_public_key`), fed by 10 Sv1 miners. Set the translator's `enable_vardiff = false` here so JDC owns difficulty.
```
GitGab19: `tproxy-config-local-jdc-example.toml` already does most of this (JDC address/port, JDC pubkey, `aggregate_channels = false`), so this could point there and only call out the `enable_vardiff` change.

Two follow-ups:

- Should this translator-behind-JDC scenario also state `aggregate_channels = false`? With aggregation on, JDC sees a single channel and sets one target for all 10 miners, which isn't really "JDC owns difficulty" per miner.
- That example keeps `enable_vardiff = true`, while the hosted-pool example's comment says to set it false with JDC. The two examples disagree; probably worth its own issue.

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+
+### Running a fleet of `sv2-cpu-miner`
+
+It takes one process per miner and reads a single TOML file, but every field can be overridden per process with a `CPU_MINER__<FIELD>` environment variable. A fleet is therefore one shared config plus per-process overrides of whatever must differ:
+
+```sh
+for i in $(seq 1 10); do
+  CPU_MINER__USER_IDENTITY="release-test-$i" CPU_MINER__DEVICE_ID="cpu-miner-$i" \
+    cpu_miner_sv2 -c config.toml -f "miner-$i.log" &
+done
```
GitGab19: Since this is meant to be agent-driven: each Bash call from a harness runs in a fresh shell, so the job table is gone by the time teardown runs. Suggest recording PIDs as they are launched (`echo $! >> "$WS/pids"`), or launching via `setsid` / `systemd-run --user`.

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+
+### Sv2 miner variations
+
+`sv2-cpu-miner` is a generic Sv2 Mining Protocol client, not only a mining device, so its knobs reach server-side paths that ten plain miners never touch. Both JDC and the hosted pool accept standard and extended channels, so these are worth running against each:
+
+- **Many channels on one connection** (`n_extended_channels`, `n_standard_channels`; at least one must be greater than 0): emulates a proxy. Exercises per-channel bookkeeping and group-channel job routing on the server from a single connection. In `/api/v1/clients/{id}/channels` a single client should show every channel.
+- **`requires_standard_jobs = true`** (which forces `n_extended_channels = 0`): sets the `REQUIRES_STANDARD_JOBS` flag on `SetupConnection`, so the server must send a per-channel `NewMiningJob` instead of a group-channel `NewExtendedMiningJob`. Nothing else in our testing toolkit reaches that branch.
```
GitGab19: The miner doesn't force this, it refuses to start: `validate()` errors with "n_extended_channels must be 0 when requires_standard_jobs is true", and the shipped `config.toml` defaults to two extended channels.

Suggest: "(the miner rejects the config unless `n_extended_channels = 0`)".

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+- **`cpu_usage_percent`**: throttle so ten processes don't peg the machine. The startup hashrate measurement runs at this usage, so the advertised hashrate scales with it too.
+
+The miner also handles `SetGroupChannel` regrouping, but no sv2-apps server ever sends that message, so that path cannot be covered here.
+
+### Edge cases
+
+Beyond the happy path, walk the failure modes a real user hits. Roughly in order of how likely they are to burn a release:
+
+- **Upstream failover**: sever the primary pool connection mid-run by interposing a local relay: `socat TCP-LISTEN:13333,fork,reuseaddr TCP:75.119.150.111:3333`, point the primary `[[upstreams]]` at `127.0.0.1:13333`, and kill the `socat` process when it's time to cut the cord (or, for the startup-time variant, just configure the primary at a dead port). The translator should fall back to the next `[[upstreams]]` entry without dropping its Sv1 miners; JDC should walk its own upstream list and, once exhausted, keep mining solo on `coinbase_reward_script`. Only when every option is gone may the process exit, with a clear error.
```
GitGab19: As of today the Sv1 miners *are* dropped on fallback: the Sv1 server task exits on the fallback token (`sv1_server/mod.rs`, the `fallback_token.cancelled()` arm) and `cleanup_for_fallback` in `translator_runtime.rs` builds a brand-new `Sv1Server`. All 10 `minerd` sockets close and re-attach on their own. An agent following this line as written would flag expected behaviour as a regression.

Suggest: "The translator should reconnect to the next `[[upstreams]]` entry; expect the Sv1 miners to disconnect and re-attach by themselves within a few seconds, and `/api/v1/sv1/clients` to show all 10 again."

Keeping Sv1 sessions alive across fallback would be a feature request, not a pass criterion.

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+- **Proxy restart under load**: Ctrl+C the translator/JDC while all 10 miners are attached — expect a graceful shutdown, not a hang or panic. On restart, Sv1 miners re-attach by themselves (`minerd` retries forever); `sv2-cpu-miner` exits when the server closes the connection, so the Sv2 fleet must be relaunched — that is expected, not a bug.
```
GitGab19: An agent can't press Ctrl+C; `kill -INT <pid>` is the equivalent.

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+- **Bitcoin Core outage (JDC)**: stop the local node mid-run and try starting JDC with the node down. Look for a comprehensible error and sane recovery once the node is back, not a crash loop.
+- **New-block boundary**: keep a run alive across at least one real mainnet block. A couple of stale rejects right at the prev-hash change are acceptable; shares that stop flowing after it are not.
+- **Soak**: leave the full setup running for a few hours. Monitoring stays responsive, memory stays flat, `shares_rejected` stays at zero.
+
+### `min_individual_miner_hashrate` gotcha
+
+The example configs assume ASIC hashrate (e.g. `10 TH/s`), which seeds an initial difficulty near ~8000. CPU miners (~1 MH/s) then submit nothing for ~10 min while vardiff ramps down, so shares look stuck. For CPU testing, lower `min_individual_miner_hashrate` to ~`1_000_000.0` so shares flow within seconds. Sv2 miners self-measure and advertise their hashrate, so they need no such change.
```
GitGab19: I get a different number for the example values: 10 TH/s at 6 shares/min gives a raw difficulty of ~23k, ~16k after the Sv1 power-of-two rounding.

Suggest dropping the figure: "seeds an initial difficulty in the tens of thousands".

AGENTS.md
```diff
@@ -48,3 +48,67 @@ You always draft github issues under human supervision. Your role here is to hel
...
+
+### Verification
+
+Both apps serve the same monitoring routes (translator `:9092`, JDC `:9091`): `/api/v1/server/channels` for upstream channel count and `shares_submitted`/`shares_rejected`, and `/api/v1/clients/{id}/channels` for per-miner shares, plus `/api/v1/sv1/clients` on the translator, which returns 404 on JDC since it has no Sv1 downstreams. Snapshots are cached for `monitoring_cache_refresh_secs` (15 in the examples), so allow that long for a change to show up. Healthy run = channels open, shares accepted, zero rejected.
+
+### Teardown
+
+Stop the fleets with `kill $(jobs -p)` in the shells that launched them, stop any helper processes (`socat`), confirm nothing lingers — `pgrep -f 'minerd|cpu_miner_sv2|socat'` comes back empty (background jobs survive a closed shell) — and delete the workspace with one `rm -r`.
```
GitGab19: `kill $(jobs -p)` only works in the shell that launched the fleet, which an agent harness doesn't keep between calls; `kill $(cat "$WS/pids")` from the launch loop would.

Also `pgrep -f 'minerd|cpu_miner_sv2|socat'` matches unrelated host processes (any other `socat`, for instance), so matching on the workspace path is a safer lingering-process check.

GitGab19 (commented):
Some points found by Fable

